PR #414, opened 2019-06-21
Improvements around security groups in OpenStack

ereslibre:
## Allow Cilium healthchecks on the OpenStack security groups

By opening 4240/tcp, we allow cilium to perform health checks among
other members.

## Remove unneeded ports from OpenStack security groups

* 8285/udp was specific for Flannel UDP backend
* 80/tcp and 443/tcp
  * In the past, for Velum: remove, old
  * Aiming for ingress in the future?: remove, can be added later
    when we have ingress
* 8080/tcp unsure, remove
* 8081/tcp unsure, remove

## Allow Kubelet port also on control plane nodes on OpenStack

If we are opening the Kubelet API port on worker nodes, do the same
on master nodes so we are consistent.

If they both need to be removed, remove them altogether.

## Document all open ports in security groups for Openstack

Add a comment to every open port, so we keep track of what's the
reason for them to be open.

### DIFF
--- a/ci/infra/openstack/master-instance.tf
+++ b/ci/infra/openstack/master-instance.tf
@@ -65,8 +65,10 @@ resource "openstack_compute_instance_v2" "master" {
   }
 
   security_groups = [
-    "${openstack_compute_secgroup_v2.secgroup_base.name}",
-    "${openstack_compute_secgroup_v2.secgroup_master.name}",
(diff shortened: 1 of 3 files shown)


ereslibre:
Closing, superseded by https://github.com/SUSE/skuba/pull/587

PR #429, opened 2019-07-02
Add interface between kubeadm ClientSetFromFile and Skuba.

bergmannf:
Adding an interface between GetAdminClientSet & Skuba allows substituting a mock implementation that can return a fake ClientSet.

This interface is also passed as an explicit parameter to all functions that need access to the `GetAdminClientSet` function.

This can be used to write unit tests for our code.

## Why is this PR needed?

This PR makes our code more testable, by untangling the explicit dependency on `ClientSetFromFile` implemented in `kubeadm`.


bergmannf:
Superseded by #588 - it seems after the rename or making the repository public it is no longer correctly connecting to old forks.

So the force-push of changes is not detected for this PR.

PR #492, opened 2019-07-18
[WIP] init of openstack provider integration

Klaven:
## Why is this PR needed?
This is the initial shot at implementing an openstack cloud provider integration. It is based off the initial implementation provided by @ereslibre. This PR is on hold for right now in favor of the in-tree provider.

Fixes #https://github.com/SUSE/avant-garde/issues/577

## What does this PR do?
Provides skuba with a integration to openstack

### DIFF
--- a/internal/app/skuba/cluster/init.go
+++ b/internal/app/skuba/cluster/init.go
@@ -37,8 +37,10 @@ import (
 type initOptions struct {
 	ControlPlane      string
 	KubernetesVersion string
+	CloudProvider     string
 }
 
+// NewInitCmd creates a new `skuba cluster init` cobra command
 func NewInitCmd() *cobra.Command {
 	initOptions := initOptions{}
 
@@ -58,6 +60,7 @@ func NewInitCmd() *cobra.Command {
 
 			initConfig := cluster.InitConfiguration{
 				ClusterName:         args[0],
+				CloudProvider:       initOptions.CloudProvider,
 				ControlPlane:        initOptions.ControlPlane,
 				CiliumImage:         cilium.GetCiliumImage(),
 				CiliumInitImage:     cilium.GetCiliumInitImage(),
@@ -83,6 +86,7 @@ func NewInitCmd() *cobra.Command {
 	if skuba.BuildType == "development" {
 		cmd.Flags().StringVar(&initOptions.KubernetesVersion, "kubernetes-version", "", "The kubernetes version to bootstrap with (only in development build)")
 	}
+	cmd.Flags().StringVar(&initOptions.CloudProvider, "cloud-provider", "", "Enable cloud provider integration with the chosen cloud. Valid values: openstack")
 	cmd.MarkFlagRequired("control-plane")
 
 	return cmd
--- a/internal/pkg/skuba/deployments/ssh/cloud.go
+++ b/internal/pkg/skuba/deployments/ssh/cloud.go
@@ -1,0 +1,85 @@
+/*
+ * Copyright (c) 2019 SUSE LLC. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package ssh
+
+import (
+	"bytes"
+	"io/ioutil"
+	"text/template"
+
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
+	"github.com/SUSE/skuba/pkg/skuba"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/cmd/kubeadm/app/images"
+)
+
+func init() {
+	stateMap["cloud.deploy-controller-manager"] = cloudDeployControllerManager
+}
+
+func cloudDeployControllerManager(t *Target, data interface{}) error {
+	template, err := template.New("").Parse(kubernetes.OpenstackCloudControllerManagerTemplate)
+	if err != nil {
+		return err
+	}
+	var rendered bytes.Buffer
+	err = template.Execute(&rendered,
+		struct {
+			Image      string
+			SecretName string
+		}{
+			Image:      images.GetGenericImage(skuba.ImageRepository, "hyperkube", kubernetes.LatestVersion().String()),
+			SecretName: kubernetes.OpenstackCloudControllerManagerSecretName,
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	defer t.ssh("rm -rf /tmp/cloud.d")
+	if err := t.target.UploadFileContents("/tmp/cloud.d/openstack-cloud-controller-manager.yaml", rendered.String()); err != nil {
+		return err
+	}
+
+	cloudConfContents, err := ioutil.ReadFile(skuba.OpenstackCloudConfFile())
+	if err != nil {
+		return err
+	}
+
+	client, err := kubernetes.GetAdminClientSet()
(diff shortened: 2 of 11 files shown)


internal/pkg/skuba/kubernetes/cloud.go
```diff
@@ -0,0 +1,133 @@
...
+		   readOnly: true
+		 - mountPath: /etc/config
+		   name: openstack-cloud-config-volume
+		   readOnly: true
+	   tolerations:
+	   - key: node.cloudprovider.kubernetes.io/uninitialized
+		 value: "true"
+		 effect: NoSchedule
+	   - key: node-role.kubernetes.io/master
+		 effect: NoSchedule
```
inercia: Maybe it is an effect of GitHub, but bormatting seems to be wrong here and on the `hostPaths`

internal/pkg/skuba/kubernetes/cloud.go
```diff
@@ -0,0 +1,133 @@
...
+	   nodeSelector:
+		 node-role.kubernetes.io/master: ""
+	   volumes:
+	   - hostPath:
+		   path: /etc/kubernetes/pki
+		   type: DirectoryOrCreate
+		 name: k8s-certs
+	   - hostPath:
```
inercia: Do we need to mount the certificates?

bergmannf: I would imagine the mounted certificates are needed in our development environment, where we need certificates signed from the SUSE CA for the keystone endpoint.

I expect some customers will have a similar certificate setup.

pkg/skuba/actions/cluster/init/assets.go
```diff
@@ -0,0 +1,73 @@
...
+ 
+ Create a file inside this directory named ~cloud.conf~, with the [supported contents](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/provider-configuration.md):
+ 
+ ~~~
+ [Global]
+ username=user
+ password=pass
+ auth-url=https://<keystone_ip>/identity/v3
+ tenant-id=c869168a828847f39f7f06edd7305637
+ domain-id=2a73b8f597c04551a0fdc8e95544be8a
```
bergmannf: `domain-id` is optional according to https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#openstack - so we could make the example configuration even slimmer.

pkg/skuba/actions/cluster/init/assets.go
```diff
@@ -0,0 +1,73 @@
...
+ this name will be used by the ~Openstack~ cloud controller manager to reconcile
+ node metadata.
+ `, "~", "`")
+
+	openstackCloudConfTemplate = `[Global]
+ username=user
+ password=pass
+ auth-url=https://<keystone_ip>/identity/v3
+ tenant-id=c869168a828847f39f7f06edd7305637
+ domain-id=2a73b8f597c04551a0fdc8e95544be8a
```
bergmannf: Same as above `domain-id` is optional, so we could remove it.

mjura:
Today I have tested `in-tree` openstack cloud provider, it works properly and it is easier to setup. All details I have included in https://github.com/mjura/caasp-cpi/blob/master/README.md (load balancer and storage provisioning works out-of-the-box)

My proposal would be to implement this integration in two steps:
1. Based on `in-tree` cloud provider configuration, then `openstack-cloud-controller-manager` is not needed
2. Migrate to `out-tree` cloud provider configuration with all features like storage provisioning.

Please comment on this

internal/app/skuba/cluster/init.go
```diff
@@ -83,6 +86,7 @@ func NewInitCmd() *cobra.Command {
 	if skuba.BuildType == "development" {
 		cmd.Flags().StringVar(&initOptions.KubernetesVersion, "kubernetes-version", "", "The kubernetes version to bootstrap with (only in development build)")
 	}
+	cmd.Flags().StringVar(&initOptions.CloudProvider, "cloud-provider", "", "Enable cloud provider integration with the chosen cloud. Valid values: openstack")
```
mjura: Probably we would like to add `cloud-config` option with path to cloud provider configuration file, when it is provided by user or generated by terraform look on https://github.com/SUSE/skuba/pull/499

internal/pkg/skuba/deployments/ssh/cloud.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
+	"github.com/SUSE/skuba/pkg/skuba"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/cmd/kubeadm/app/images"
+)
+
+func init() {
+	stateMap["cloud.deploy-controller-manager"] = cloudDeployControllerManager
```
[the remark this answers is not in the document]
mjura: I would replace it with `cloud.deploy-cloud-provider`

internal/pkg/skuba/deployments/ssh/cloud.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func cloudDeployControllerManager(t *Target, data interface{}) error {
+	template, err := template.New("").Parse(kubernetes.OpenstackCloudControllerManagerTemplate)
```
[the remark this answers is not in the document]
mjura: Yes, next step we would like to add other cloud providers

internal/pkg/skuba/deployments/ssh/cloud.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return err
+	}
+	var rendered bytes.Buffer
+	err = template.Execute(&rendered,
+		struct {
+			Image      string
+			SecretName string
+		}{
+			Image:      images.GetGenericImage(skuba.ImageRepository, "hyperkube", kubernetes.LatestVersion().String()),
```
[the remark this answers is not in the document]
mjura: No it is different image `cloud-provider-openstack:1.15.0`

internal/pkg/skuba/kubernetes/cloud.go
```diff
@@ -0,0 +1,133 @@
...
+		 image: {{.Image}}
+		 command:
+		 - /usr/local/bin/cloud-controller-manager
+		 - --cloud-config=$(CLOUD_CONFIG)
+		 - --cloud-provider=openstack
+		 - --leader-elect=true
+		 - --use-service-account-credentials
+		 - --allocate-node-cidrs=true
+		 - --configure-cloud-routes=true
+		 - --cluster-cidr=172.17.0.0/16
```
mjura:
```
    - openstack-cloud-controller-manager
    - --allocate-node-cidrs=true
    - --authentication-kubeconfig=/etc/kubernetes/controller-manager.conf
    - --authorization-kubeconfig=/etc/kubernetes/controller-manager.conf
    - --bind-address=127.0.0.1
    - --client-ca-file=/etc/kubernetes/pki/ca.crt
    - --cluster-cidr=10.244.0.0/16
    - --kubeconfig=/etc/kubernetes/controller-manager.conf
    - --leader-elect=true
    - --requestheader-client-ca-file=/etc/kubernetes/pki/front-proxy-ca.crt
    - --use-service-account-credentials=true
    - --cloud-config=$(CLOUD_CONFIG)
    - --cloud-provider=openstack
```

internal/pkg/skuba/kubernetes/cloud.go
mjura:
```
  - key: node-role.kubernetes.io/master
    operator: "Exists"
    effect: NoSchedule
```

internal/app/skuba/cluster/init.go
ereslibre: I would try to not add the `--cloud-config` here until we have more information about how other cloud providers are integrated. `cloud.conf` is an OpenStack specific detail and I think we should try to avoid exposing specific workflow flags on the tool in something as generic as `skuba cluster init`.

The workflow would be that they run: `skuba cluster init --cloud-provider openstack`, they get generated the cluster definition folder with some [default `cloud.conf` as described in the RFC](https://github.com/SUSE/caasp-rfc/blob/master/accepted/0024-openstack-cloud-provider-integration.md), they change it as they want (they can copy some `cloud.conf` they have around and overwrite the sample one etc...), and then they bootstrap and join the nodes.

Klaven:
this PR is no longer productive. when we move to out-of-tree providers it will be of little use.

PR #494, opened 2019-07-18
Remove CaaSP pattern installation from autoyast

ereslibre:
## Why is this PR needed?

Remove remaining CaaSP pattern installation from the autoyast
profile.


ereslibre:
Small leftover from https://github.com/SUSE/skuba/pull/485

PR #497, opened 2019-07-18
Fix Jenkins cluster naming

tdaines42:
## Why is this PR needed?

The VMware load balancer was failing to setup because of how the clusters were being named by Jenkins. Also the cloud init wait wasn't waiting for the lb to be done.

Fixes #

## What does this PR do?

Updates how we name the clusters with jenkins replacing `/` with `-`.
Makes it so the LB should be ready before the cloud init wait is started.

### DIFF
--- a/ci/infra/vmware/lb-instance.tf
+++ b/ci/infra/vmware/lb-instance.tf
@@ -145,7 +145,8 @@ resource "vsphere_virtual_machine" "lb" {
 }
 
 resource "null_resource" "lb_wait_cloudinit" {
-  count = "${var.lbs}"
+  depends_on = ["vsphere_virtual_machine.lb"]
+  count      = "${var.lbs}"
 
   connection {
     host  = "${element(vsphere_virtual_machine.lb.*.guest_ip_addresses.0, count.index)}"
--- a/ci/jenkins/pipelines/prs/skuba-test-vmware.Jenkinsfile
+++ b/ci/jenkins/pipelines/prs/skuba-test-vmware.Jenkinsfile
@@ -11,7 +11,7 @@ pipeline {
         VMWARE_ENV_FILE = credentials('vmware-env')
         GITHUB_TOKEN = credentials('github-token')
         PLATFORM = 'vmware'
-        TERRAFORM_STACK_NAME = "${JOB_NAME}-${BUILD_NUMBER}"
+        TERRAFORM_STACK_NAME = "${JOB_NAME.replaceAll("/","-")}-${BUILD_NUMBER}"
         PR_CONTEXT = 'jenkins/skuba-test-vmware'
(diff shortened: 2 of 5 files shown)


ci/infra/vmware/lb-instance.tf
```diff
@@ -160,3 +161,23 @@ resource "null_resource" "lb_wait_cloudinit" {
...
+    environment = {
+      host = "${element(vsphere_virtual_machine.lb.*.guest_ip_addresses.0, count.index)}"
+    }
+
+    command = <<EOT
+apistatus=$(curl -s "$host:9000/stats;csv" | awk -F ',' '/apiserver,FRONTEND/{print$18}')
+while [ "$apistatus" != "OPEN" ]
+do
+  sleep 5
+  apistatus=$(curl -s "$host:9000/stats;csv" | awk -F ',' '/apiserver,FRONTEND/{print$18}')
```
danielorf: You probably want a timeout (either in loop cycles or time) so this doesn't get stuck.

ci/infra/vmware/lb-instance.tf
```diff
@@ -160,3 +161,23 @@ resource "null_resource" "lb_wait_cloudinit" {
...
+  count      = "${var.lbs}"
+
+  provisioner "local-exec" {
+    environment = {
+      host = "${element(vsphere_virtual_machine.lb.*.guest_ip_addresses.0, count.index)}"
+    }
+
+    command = <<EOT
+apistatus=$(curl -s "$host:9000/stats;csv" | awk -F ',' '/apiserver,FRONTEND/{print$18}')
+while [ "$apistatus" != "OPEN" ]
```
stefsuse: probably a good idea to add a max_retry value here

tdaines42: Removed this since it isn't needed

ci/infra/vmware/lb-instance.tf
tdaines42: Removed this since it isn't needed

dannysauer:
Is it sufficient to replace just `/`, or are there other chars which might sneak in?  Would a "replace everything but `[a-z0-9]`" or something like that be doable?

tdaines42:
> Is it sufficient to replace just `/`, or are there other chars which might sneak in? Would a "replace everything but `[a-z0-9]`" or something like that be doable?

The names come from Jenkins based on the job name so we have some control over that. I think if we want to add any other replacement maybe it should go into testrunner.

tdaines42:
> approved in the interest of moving forward
> 
> But I'm thinking that if one character was incorrectly encoded it is likely that others are, or will be. Is this a more widespread issue with how we are generating the names, perhaps?

Names are usually chosen by who or what is running testrunner so in this case Jenkins was using the job name. Maybe an issue could be opened to look into this more.

PR #499, opened 2019-07-19
ci: infra: Generate cloud provider config file on nodes

mjura:
## Why is this PR needed?

It will generate cloud provider integration config file for OpenStack

Issue # https://github.com/SUSE/avant-garde/issues/568

## What does this PR do?

1. Use OpenStack and Terraform variables to generate `openstack.conf` file needed by skuba for enabling cloud provider integration
2. Will suggest about next steps how to bootstrap `skuba` cluster, example output below
```
null_resource.generate_cloud_provider_conf (local-exec): Executing: ["/bin/sh" "-c" "bash generate-cloud-provider-conf"]

null_resource.generate_cloud_provider_conf (local-exec): [ INFO ] ### Run following commands to bootstrap skuba cluster:

null_resource.generate_cloud_provider_conf (local-exec):  skuba cluster init --control-plane 10.86.2.28 --cloud-provider openstack mjura-cluster
null_resource.generate_cloud_provider_conf (local-exec):  mv openstack.conf mjura-cluster/cloud/openstack.conf
null_resource.generate_cloud_provider_conf (local-exec):  cd mjura-cluster
null_resource.generate_cloud_provider_conf (local-exec):  skuba node bootstrap --target 10.86.1.214 --sudo --user sles caasp-master-mjura-0
null_resource.generate_cloud_provider_conf (local-exec):  skuba node join --role worker --target 10.86.0.195 --sudo --user sles caasp-worker-mjura-0
null_resource.generate_cloud_provider_conf (local-exec):  skuba node join --role worker --target 10.86.2.19 --sudo --user sles caasp-worker-mjura-1
```

### DIFF
--- a/ci/infra/openstack/generate-cloud-provider-conf
+++ b/ci/infra/openstack/generate-cloud-provider-conf
@@ -1,0 +1,79 @@
+#!/bin/bash
+
+log()   { (>&1 echo -e "$@") ; }
+cmd()   { log "[ CMD ] $@" ; }
+info()  { log "[ INFO ] $@" ; }
+error() { (>&2 echo -e "[ ERROR ] $@") ;}
+
+if [ -z "${OS_AUTH_URL}" ] || [ -z "${OS_USERNAME}" ] || \
+   [ -z "${OS_PASSWORD}" ] || [ -z "${OS_PROJECT_ID}" ] || \
+   [ -z "${OS_PRIVATE_SUBNET_ID}" ] || [ -z "${OS_PUBLIC_NET_ID}" ]; then
+  error '$OS_AUTH_URL $OS_USERNAME $OS_PASSWORD $OS_PROJECT_ID'
+  error '$OS_PRIVATE_SUBNET_ID $OS_PUBLIC_NET_ID must be specified'
+  error 'Please download and source your OpenStack RC file'
+  exit 1
+fi
+
+OPENSTACK_CONF="openstack.conf"
+
+umask 077
+
+cat << EOF > "${OPENSTACK_CONF}"
+[Global]
+auth-url="${OS_AUTH_URL}"
+username="${OS_USERNAME}"
+password="${OS_PASSWORD}"
+tenant-id="${OS_PROJECT_ID}"
+tenant-name="${OS_PROJECT_NAME}"
+domain-id="${OS_USER_DOMAIN_ID}"
+domain-name="${OS_USER_DOMAIN_NAME}"
+region="${OS_REGION_NAME}"
+ca-file="${CA_FILE}"
+[LoadBalancer]
+lb-version=v2
+subnet-id="${OS_PRIVATE_SUBNET_ID}"
+floating-network-id="${OS_PUBLIC_NET_ID}"
+create-monitor=yes
+monitor-delay=1m
+monitor-timeout=30s
+monitor-max-retries=3
+[BlockStorage]
+trust-device-path=false
+bs-version=v2
+ignore-volume-az=true
+EOF
+
+umask 022
+
+[ -z "$OS_PROJECT_NAME" ] && sed -i '/^tenant-name=/d' "${OPENSTACK_CONF}"
+[ -z "$OS_USER_DOMAIN_ID" ] &&  sed -i '/^domain-id=/d' "${OPENSTACK_CONF}"
+[ -z "$OS_USER_DOMAIN_NAME" ] && sed -i '/^domain-name=/d' "${OPENSTACK_CONF}"
+[ -z "$CA_FILE" ] && sed -i '/^ca-file=/d' "${OPENSTACK_CONF}"
+
+if [ -z "${TR_STACK}" ] || [ -z "${TR_LB_IP}" ] || \
+   [ -z "$TR_MASTER_IPS" ] || [ -z "$TR_WORKER_IPS" ] || \
+   [ -z "${TR_USERNAME}" ]; then
+  error '$TR_STACK $TR_LB_IP $TR_MASTER_IPS $TR_WORKER_IPS must be specified'
+  exit 1
+fi
+
+info "### Run following commands to bootstrap skuba cluster:\n"
+cmd " skuba cluster init --control-plane ${TR_LB_IP} --cloud-provider openstack ${TR_STACK}-cluster"
+cmd " mv openstack.conf ${TR_STACK}-cluster/cloud/openstack/openstack.conf"
+cmd " cd ${TR_STACK}-cluster"
+
+i=0
+for MASTER in $TR_MASTER_IPS; do
+  if [ $i -eq "0" ]; then
+    cmd " skuba node bootstrap --target ${MASTER} --sudo --user ${TR_USERNAME} caasp-master-${TR_STACK}-0"
+  else
+    cmd " skuba node join --role master --target ${MASTER} --sudo --user ${TR_USERNAME} caasp-master-${TR_STACK}-${i}"
+  fi
+  ((++i))
+done
+
+i=0
+for WORKER in $TR_WORKER_IPS; do
+  cmd " skuba node join --role worker --target ${WORKER} --sudo --user ${TR_USERNAME} caasp-worker-${TR_STACK}-${i}"
+  ((++i))
+done
--- a/ci/infra/openstack/variables.tf
+++ b/ci/infra/openstack/variables.tf
@@ -116,3 +116,13 @@ variable "rmt_server_name" {
   default     = ""
   description = "SUSE Repository Mirroring Server Name"
 }
+
+variable "cpi_enable" {
+  default     = false
+  description = "Enable CPI integration with OpenStack"
+}
+
+variable "ca_file" {
+  default     = ""
+  description = "Used to specify the path to your custom CA file"
+}
--- a/ci/infra/openstack/cloud-provider.tf
+++ b/ci/infra/openstack/cloud-provider.tf
@@ -1,0 +1,19 @@
+resource "null_resource" "generate_cloud_provider_conf" {
+  depends_on = ["null_resource.master_reboot", "null_resource.worker_reboot"]
+  count      = "${var.cpi_enable ? 1 : 0}"
+
+  provisioner "local-exec" {
+    environment = {
+      CA_FILE              = "${var.ca_file}"
+      TR_STACK             = "${var.stack_name}"
+      TR_USERNAME          = "${var.username}"
+      TR_LB_IP             = "${openstack_networking_floatingip_v2.lb_ext.address}"
+      TR_MASTER_IPS        = "${join(" ", openstack_networking_floatingip_v2.master_ext.*.address)}"
+      TR_WORKER_IPS        = "${join(" ", openstack_networking_floatingip_v2.worker_ext.*.address)}"
+      OS_PRIVATE_SUBNET_ID = "${openstack_networking_subnet_v2.subnet.id}"
+      OS_PUBLIC_NET_ID     = "${data.openstack_networking_network_v2.external_network.id}"
+    }
+
+    command = "bash generate-cloud-provider-conf"
+  }
+}


mjura:
I have tested it and works fine

mjura:
@ereslibre Ok, let's get change approach then. I would like to generate then locally cloud provider config file, which will be used by `skuba`. This is because `terraform` knows the best about private network subnet id, floating network id about masters and workers. With this change we will get fully automated cloud provider integration process.

ereslibre:
@mjura, I think generating it it's fine, but we should document to the user how to copy the terraform generated `cloud.conf` on their cluster definition folder. The idea is to not tie both processes together, if they use terraform it's handier, and they are a `cp` away from configuring the `cloud.conf`.

mjura:
@ereslibre I have got your point, this change will only generate `openstack.conf` when `enable_cpi` variable will be switched on. At the end it will produce also suggestion to the user how to spin `skuba` cluster. This can help with automating bootstraping process.

`terraform` and `skuba` are still separate in the process.

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,99 @@
...
+monitor-max-retries=3
+[BlockStorage]
+trust-device-path=false
+bs-version=v2
+ignore-volume-az=true
+EOF
+
+umask 0022
+
+sed -i '/^auth-url=/ s|=.*|="'${OS_AUTH_URL}'"|
```
pablochacin: why not use variable substitution in the cat command above instead of using sed once the file is created?

mjura: Thank you, you are right. I have changed that

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,91 @@
...
+monitor-max-retries=3
+[BlockStorage]
+trust-device-path=false
+bs-version=v2
+ignore-volume-az=true
+EOF
+
+umask 0022
+
+if [ -n "$OS_PROJECT_ID" ]; then
```
pablochacin: I think you can avoid this complex logic and using `sed` if you follow this pattern:
```
TENANT_ID=${OS_PROJECT_ID:-${OS_PROJECT_NAME}}
if [[ -z TENANT_ID ]]; then
  echo "[ERROR] Either OS_PROJECT_ID or OS_PROJECT_NAME must be specified"
fi
```

You can then use TENANT_ID in the cat command above

mjura: Yeah, I was thinking about that, but `tenant-id` or `tenant-name` option have to be used in `openstack.conf` file. In this case `sed` is used for setting and cleaning unnecessary option. Using `sed` is good solution, so I would rather to keep it.

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,91 @@
...
+  fi
+else
+  echo; echo "[ WARN ] \$TR_MASTER_IPS wasn't provided"
+fi
+
+if [ -n "$TR_WORKER_IPS" ]; then
+  i=0
+  for WORKER in $TR_WORKER_IPS; do
+    echo " skuba node join --role worker --target $WORKER --sudo --user sles caasp-worker-${TR_STACK}-${i}"
+    ((++i))
```
dannysauer: As above, the `${i}` in the `echo` could be replaced with `$((i++))` and remove the need for the separate arithmatic expansion.  And initialize `i` to `-1` to start at `0`, since `++i` and `i++` both expand to the result after incrementing. :/

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+
+umask 0077
```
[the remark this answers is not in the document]
mjura: This is because default Linux file mask is coded on 4 digits, first digit is defining `special mode` for
```
4 - setuid
2 - setgid
1 - sticky-bit
0 - no special mode for file and it is ignored
```

ci/infra/openstack/generate-cloud-provider-conf
mjura: I have optimized logic for this script, which is still human readable.

ci/infra/openstack/generate-cloud-provider-conf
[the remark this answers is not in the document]
mjura: I have changed it

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,91 @@
...
+if [ -n "$OS_PROJECT_ID" ]; then
+  sed -i '/^tenant-name=/d
+          /^tenant-id=/ s|=.*|="'${OS_PROJECT_ID}'"|
+' openstack.conf
+elif [ -n "$OS_PROJECT_NAME" ]; then
+  sed -i '/^tenant-id=/d
+          /^tenant-name=/ s|=.*|="'${OS_PROJECT_NAME}'"|
+' openstack.conf
+else
+  echo "[ WARN ] \$OS_PROJECT_NAME and \$OS_PROJECT_ID weren't provided"
```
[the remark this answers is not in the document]
mjura: Ok, done

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,91 @@
...
+if [ -n "$TR_MASTER_IPS" ]; then
+  MASTERS=($TR_MASTER_IPS)
+  echo " skuba node bootstrap --target ${MASTERS[0]} --sudo --user sles caasp-master-${TR_STACK}-0"
+  if [ "${#MASTERS[@]}" -gt "1" ]; then
+    i=0
+    for MASTER in $TR_MASTER_IPS; do
+      if [ $i -eq "0" ]; then ((++i)); continue; fi
+      echo " skuba node join --role master --target $MASTER --sudo --user sles caasp-master-${TR_STACK}-${i}"
+    done
+  fi
```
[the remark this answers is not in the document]
mjura: I would like to keep this script still human readable

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,91 @@
...
+fi
+
+if [ -n "$CA_FILE" ]; then
+  sed -i '/^ca-file=/ s|=.*|="'${CA_FILE}'"|
+' openstack.conf
+else
+  sed -i '/^ca-file=/d' openstack.conf
+fi
+
+echo; echo "[ INFO ] ### Run following commands to bootstrap skuba cluster:"; echo
```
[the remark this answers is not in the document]
mjura: Done

ci/infra/openstack/generate-cloud-provider-conf
mjura: I would like to keep it as it is, since it is easy to read and understand

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,72 @@
...
+echo " skuba cluster init --control-plane ${TR_LB_IP} --cloud-provider openstack ${TR_STACK}-cluster"
+echo " mv openstack.conf ${TR_STACK}-cluster/cloud/openstack.conf"
+echo " cd ${TR_STACK}-cluster"
+
+MASTERS=($TR_MASTER_IPS)
+echo " skuba node bootstrap --target ${MASTERS[0]} --sudo --user sles caasp-master-${TR_STACK}-0"
+if [ "${#MASTERS[@]}" -gt "1" ]; then
+  i=0
+  for MASTER in $TR_MASTER_IPS; do
+    if [ $i -eq "0" ]; then ((++i)); continue; fi
```
dannysauer: Why not echo the bootstrap command here in place of the `continue`, and make this an if/else? Then the array will also be unnecessary. Some readers get confused by shell arrays. :) 

Either way, the increment of `i` also needs to happen outside the if, which would be easier with an `else` rather than `continue`. :)

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+if [ -z "${OS_AUTH_URL}" ] || [ -z "${OS_USERNAME}" ] || \
+   [ -z "${OS_PASSWORD}" ] || [ -z "${OS_PROJECT_ID}" ] || \
+   [ -z "${OS_PRIVATE_SUBNET_ID}" ] || [ -z "${OS_PUBLIC_NET_ID}" ]; then
+  echo '[ ERROR ] $OS_AUTH_URL $OS_USERNAME $OS_PASSWORD $OS_PROJECT_ID'
+  echo '[ ERROR ] $OS_PRIVATE_SUBNET_ID $OS_PUBLIC_NET_ID must be specified'
+  echo '[ ERROR ] Please download and source your OpenStack RC file'
```
dannysauer: Errors should probably go to stderr, either `>&2` or, more readably, `>/dev/stderr`

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+if [ -z "${OS_AUTH_URL}" ] || [ -z "${OS_USERNAME}" ] || \
+   [ -z "${OS_PASSWORD}" ] || [ -z "${OS_PROJECT_ID}" ] || \
+   [ -z "${OS_PRIVATE_SUBNET_ID}" ] || [ -z "${OS_PUBLIC_NET_ID}" ]; then
```
dannysauer: Maybe consider `[[` for at least this test? With the double bracket keyword, not only do variables not need quoted, but the "or" can also go in the test. Getting rid of a lot of extra punctuation would make this quite a bit easier on the eyes, IMO. The double bracket is supported on any bash that this will ever run under.

But this is legit olde tyme shell style that still works on posix /bin/sh, so that's fine if you'd prefer to keep it this way for backwards compatibility. It's way less efficient, since single bracket is a built-in vs a keyword, but it's not like performance is the main goal here.

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,72 @@
...
+fi
+
+echo -e "\n[ INFO ] ### Run following commands to bootstrap skuba cluster:\n"
+echo " skuba cluster init --control-plane ${TR_LB_IP} --cloud-provider openstack ${TR_STACK}-cluster"
+echo " mv openstack.conf ${TR_STACK}-cluster/cloud/openstack.conf"
+echo " cd ${TR_STACK}-cluster"
+
+MASTERS=($TR_MASTER_IPS)
+echo " skuba node bootstrap --target ${MASTERS[0]} --sudo --user sles caasp-master-${TR_STACK}-0"
+if [ "${#MASTERS[@]}" -gt "1" ]; then
```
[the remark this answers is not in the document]
mjura: I agree, I have included your change

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,72 @@
...
+bs-version=v2
+ignore-volume-az=true
+EOF
+
+umask 022
+
+[ -z "$OS_PROJECT_NAME" ] && sed -i '/^tenant-name=/d' $OPENSTACK_CONF
+[ -z "$OS_USER_DOMAIN_ID" ] &&  sed -i '/^domain-id=/d' $OPENSTACK_CONF
+[ -z "$OS_USER_DOMAIN_NAME" ] && sed -i '/^domain-name=/d' $OPENSTACK_CONF
+[ -z "$CA_FILE" ] && sed -i '/^ca-file=/d' $OPENSTACK_CONF
```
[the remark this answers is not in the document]
mjura: Ok

ci/infra/openstack/generate-cloud-provider-conf
mjura: Good idea :+1:

ci/infra/openstack/generate-cloud-provider-conf
mjura: Sure, let's define some logging functions

ci/infra/openstack/generate-cloud-provider-conf
mjura: let's keep it

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,72 @@
...
+
+if [ -z "${TR_STACK}" ] || [ -z "${TR_LB_IP}" ] || [ -z "$TR_MASTER_IPS" ] || [ -z "$TR_WORKER_IPS" ]; then
+  echo '[ ERROR ] $TR_STACK $TR_LB_IP $TR_MASTER_IPS $TR_WORKER_IPS must be specified'
```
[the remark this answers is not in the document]
mjura: sure, new error function was definde

mjura:
@dannysauer Thank you for reviewing and all suggestions. I hope you will like it

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,77 @@
...
+
+info "### Run following commands to bootstrap skuba cluster:\n"
+cmd " skuba cluster init --control-plane ${TR_LB_IP} --cloud-provider openstack ${TR_STACK}-cluster"
+cmd " mv openstack.conf ${TR_STACK}-cluster/cloud/openstack/openstack.conf"
+cmd " cd ${TR_STACK}-cluster"
+
+i=0
+for MASTER in $TR_MASTER_IPS; do
+  if [ $i -eq "0" ]; then
+    cmd " skuba node bootstrap --target ${MASTER} --sudo --user sles caasp-master-${TR_STACK}-0"
```
bergmannf: Given that the `sles` username is also a variable, maybe it should be passed into the script as well and be variable substituted?

On the other hand if the customer use our provided images, this should always be sles anyhow, so I guess we could also keep it.

ci/infra/openstack/generate-cloud-provider-conf
```diff
@@ -0,0 +1,77 @@
...
+  else
+    cmd " skuba node join --role master --target ${MASTER} --sudo --user sles caasp-master-${TR_STACK}-${i}"
```
bergmannf: Same as above with the `username`

ci/infra/openstack/generate-cloud-provider-conf
mjura: You are right, thank you

ci/infra/openstack/generate-cloud-provider-conf
mjura: You are right, thank you

ci/infra/openstack/variables.tf
```diff
@@ -116,3 +116,13 @@ variable "rmt_server_name" {
   default     = ""
   description = "SUSE Repository Mirroring Server Name"
 }
+
+variable "cpi_enable" {
+  default     = false
```
drpaneas: Do you think we need to modify the examples file as well or not needed?

drpaneas:
If documentation is needed for this PR, please add the `documentation` label.

ci/infra/openstack/variables.tf
mjura: I think it is not needed, CPI will be enabled only in specific use cases

PR #495, opened 2019-07-18
[WIP] ci/infra/testrunner: added e2e for skuba upgrade plan

mssola:
Signed-off-by: Miquel Sabaté Solà <msabate@suse.com>

PR #501, opened 2019-07-19
e2e tests for skuba upgrade plan

MaximilianMeister:
### Part of

https://github.com/SUSE/avant-garde/issues/525

### Fixes

https://github.com/SUSE/avant-garde/issues/590
https://github.com/SUSE/avant-garde/issues/591

### Follow up for

https://github.com/SUSE/skuba/pull/495


ci/infra/testrunner/skuba/skuba.py
```diff
@@ -81,8 +91,8 @@ def node_bootstrap(self):
         self._verify_bootstrap_dependency()
 
         master0_ip = self.platform.get_nodes_ipaddrs("master")[0]
-        cmd = "node bootstrap --user {username} --sudo --target \
-                 {ip} my-master-0".format(ip=master0_ip, username=self.conf.nodeuser)
+        cmd = "{} node bootstrap --user {username} --sudo --target \
+                 {ip} my-master-0".format(self.SKUBA_VERBOSE, ip=master0_ip, username=self.conf.nodeuser)
```
mssola: Instead of doing this, pass it on the `env` dictionary (I don't know if it's exposed for this funcion). Just saying because doing this failed miserably in `skuba-update`.

ci/jenkins/pipelines/skuba-cluster-upgrade-plan-1.14-nightly.Jenkinsfile
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+pipeline {
+   agent { node { label 'caasp-team-private' } }
+
+   environment {
+        OPENSTACK_OPENRC = credentials('ecp-openrc')
+        TERRAFORM_STACK_NAME = "${JOB_NAME}-${BUILD_NUMBER}"
+        GITHUB_TOKEN = credentials('github-token')
+        VMWARE_ENV_FILE = credentials('vmware-env')
```
mssola: You will also need the `WORKSPACE` one, no?

MaximilianMeister:
> Adding a pipeline for each test is too much, I think. We would end up with tons of pipelines if we keep adding tests. So, for now I'd go for just one pipeline for these four tests. In the future we might split things up a bit, but for now I see it as complicating things a bit too much thinking

I chose one pipeline for each test, because we are setting up and tearing down a whole cluster on each test. the destruction of a cluster for my understanding is the latter end of `e2e`. I'd keep it separated, otherwise we'll start and destroy x amount of clusters in one pipeline, which I find much more confusing, especially if one fails, and the following ones would not run at all.

I am not really worried about too many pipelines, as we can even group them together if we want.

ci/jenkins/pipelines/skuba-cluster-upgrade-plan-1.14-nightly.Jenkinsfile
MaximilianMeister: not sure... `WORKSPACE` should always be set/available by default i think

ci/infra/testrunner/skuba/skuba.py
pablochacin: Same comment here with the use of `self.conf.verbose`

pablochacin:
> I am not really worried about too many pipelines, as we can even group them together if we want.

I would like to hear @tdaines42 opinion here. I think that the number of pipelines would eventually make maintenance harder.

mssola:
@pablochacin hopefully addressed your comments

tdaines42:
> > I am not really worried about too many pipelines, as we can even group them together if we want.
> 
> I would like to hear @tdaines42 opinion here. I think that the number of pipelines would eventually make maintenance harder.

I would prefer to have one or two pipelines and job templates for this. You could then define the jobs in `ci/jenkins/jobs.yaml` like

```
- project:
    name: caasp-jobs/caasp-v4-upgrade-plan
    repo-name: skuba
    repo-owner: SUSE
    repo-credentials: github-token
    version:
        - latest
        - 1.14
    jobs:
        - '{name}-cluster-{version}-nightly'
        - '{name}-node-{version}-nightly'
```

ci/jenkins/pipelines/skuba-upgrade-nightly.Jenkinsfile
```diff
@@ -0,0 +1,42 @@
...
+        GITHUB_TOKEN = credentials('github-token')
+        VMWARE_ENV_FILE = credentials('vmware-env')
+   }
+
+   stages {
+        stage('Getting Ready For Cluster Deployment') { steps {
+            sh(script: 'make -f skuba/ci/Makefile pre_deployment', label: 'Pre Deployment')
+        } }
+
+        parallel {
```
MaximilianMeister: would it create 2 stacks with the same name ("`${JOB_NAME}-${BUILD_NUMBER}`") at the same time? maybe you'd have to pass this as an `env` to each test and remove the global `TERRAFORM_STACK_NAME` env?

mssola: Done!

pablochacin:
This PR has significant conflicts with https://github.com/SUSE/skuba/pull/525 due to stylig changes in classes `Skuba`and `Utils`, which are not essential to the purpose of the PR.

To minimize the effort required to get both merged, it was discussed with @MaximilianMeister the following strategy:
1. Drop the commits that generate the conflicts
1. Rebase to PR 525
1. Reintroduce required changes dropped in first step
1. Address styling issues in a subsequent PR

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -152,6 +156,16 @@ def node_reset(self, role="worker", nr=0):
         except Exception as ex:
             raise Exception("Error executing cmd {}".format(cmd)) from ex
 
+    @step
+    def node_upgrade(self, role, nr):
+        self._verify_bootstrap_dependency()
+        return self._run_skuba("node upgrade plan {}-{}".format(role, nr))
+
+    @step
+    def cluster_upgrade(self):
```
MaximilianMeister: this should be called `cluster_upgrade_plan`

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -152,6 +156,16 @@ def node_reset(self, role="worker", nr=0):
         except Exception as ex:
             raise Exception("Error executing cmd {}".format(cmd)) from ex
 
+    @step
+    def node_upgrade(self, role, nr):
```
MaximilianMeister: this should be called `node_upgrade_plan`

ci/infra/testrunner/testrunner.py
```diff
@@ -39,15 +39,18 @@ def build_skuba(options):
...
-    skuba.cluster_init()
+    skuba.cluster_init(kubernetes_version=options.kubernetes_version)
     skuba.node_bootstrap()
 
 
 def cluster_status(options):
     print(Skuba(options.conf, options.platform).cluster_status())
 
 
+def cluster_upgrade(options):
```
MaximilianMeister: `cluster_upgrade_plan`

ci/infra/testrunner/testrunner.py
```diff
@@ -67,6 +70,10 @@ def reset_node(options):
     Skuba(options.conf, options.platform).node_reset(role=options.role, nr=options.node)
 
 
+def upgrade_node(options):
```
MaximilianMeister: `node_upgrade_plan`

MaximilianMeister:
> we have to change the method names accordingly

adapted

ci/infra/testrunner/testrunner.py
```diff
@@ -39,15 +39,18 @@ def build_skuba(options):
...
+    skuba.cluster_init(kubernetes_version=options.kubernetes_version)
     skuba.node_bootstrap()
 
 
 def cluster_status(options):
     print(Skuba(options.conf, options.platform).cluster_status())
 
 
+def cluster_upgrade_plan(options):
+    print(Skuba(options.conf, options.platform).cluster_upgrade_plan())
```
chentex: is this print going to show during tests?

MaximilianMeister: hmm good question, i also don't know for what we have the `print` wrapper on some method calls. maybe @pablochacin can shed some light here

pablochacin: It shouldn't be there. Print is only for the commands that are expected to return an output to the user, like `status` and and `info`

ci/infra/testrunner/testrunner.py
```diff
@@ -133,8 +134,14 @@ def main():
     cmd_status = commands.add_parser("status", help="check K8s cluster status")
     cmd_status.set_defaults(func=cluster_status)
 
-    cmd_cluster_upgrade_plan = commands.add_parser("cluster-upgrade-plan", help="plan cluster upgrade")
-    cmd_cluster_upgrade_plan.set_defaults(func=cluster_upgrade_plan)
+    # common parameters for cluster and node upgrade commands
```
MaximilianMeister: this comment should be just for cluster upgrade commands

ci/infra/testrunner/testrunner.py
```diff
@@ -133,8 +134,14 @@ def main():
...
+    upgrade_args = ArgumentParser(add_help=False)
+    upgrade_args.add_argument("-a", "--action", dest="upgrade_action",
+                              help="action: plan or apply upgrade", choices=["plan", "apply"])
```
MaximilianMeister: there's no `cluster upgrade apply` only `node upgrade apply`

pablochacin: Ok, I see let me clarify there's `cluster upgrade plan`   and  `node upgrade plan/apply`?

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -157,14 +157,14 @@ def node_reset(self, role="worker", nr=0):
...
     @step
-    def node_upgrade_plan(self, role, nr):
+    def node_upgrade(self, action, role, nr):
         self._verify_bootstrap_dependency()
-        return self._run_skuba("node upgrade plan {}-{}".format(role, nr))
+        return self._run_skuba("node upgrade {} {}-{}".format(action, role, nr))
 
     @step
-    def cluster_upgrade_plan(self):
+    def cluster_upgraden(self, action):
```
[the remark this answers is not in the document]
pablochacin: fixed. thanks

ci/infra/testrunner/testrunner.py
MaximilianMeister: i'd still call this `cluster_upgrade_plan` as it doesn't really do a cluster_upgrade

PR #502, opened 2019-07-19
[WIP] Add timeout option

chentex:
## Why is this PR needed?

Fixes #

## What does this PR do?


pablochacin:
I'm concerned this change totally the way we implement the functions and this approach won't work for the skuba or terraform commands.

chentex:
not, going to be a feasible way, we have to find another way.

PR #504, opened 2019-07-22
Init of openstack provider integration

Klaven:
## Why is this PR needed?
This is the initial shot at implementing an openstack cloud provider integration. It is based off the initial implementation provided by @ereslibre. This differes from the original https://github.com/SUSE/skuba/pull/492 in that it uses an in-tree provider.
Fixes #SUSE/avant-garde#577

## What does this PR do?
Provides an integration for the openstack cloud provider.

### DIFF
--- a/pkg/skuba/actions/node/bootstrap/bootstrap.go
+++ b/pkg/skuba/actions/node/bootstrap/bootstrap.go
@@ -33,6 +33,7 @@ import (
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
 	"github.com/SUSE/skuba/internal/pkg/skuba/node"
 	"github.com/SUSE/skuba/pkg/skuba"
+	"github.com/SUSE/skuba/pkg/skuba/cloud"
 )
 
 // Bootstrap initializes the first master node of the cluster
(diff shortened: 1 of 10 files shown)


Klaven:
will squash when approved.

Klaven:
pushed updates to man pages

pkg/skuba/actions/node/bootstrap/bootstrap.go
```diff
@@ -144,3 +150,38 @@ func setApiserverAdmissionPlugins(initConfiguration *kubeadmapi.InitConfiguratio
...
+		ReadOnly:  true,
+		PathType:  "FileOrCreate",
+	}
+	cloudVolume = append(cloudVolume, cloudConfig)
+	initConfiguration.APIServer.ControlPlaneComponent.ExtraVolumes = cloudVolume
+	initConfiguration.ControllerManager.ExtraVolumes = cloudVolume
+}
+
+func setCloudConfiguration(initConfiguration *kubeadmapi.InitConfiguration) {
+
```
mjura: Please remove this empty line

Klaven:
> This PR creates some templates that must be filled by user, but we are not checking that users are really doing that. I think that gives users the impression that doing an `init` with `--cloud-provider` is enough. Maybe we could have some text in the template that should be removed by users, and check they are doing that before running the `bootstrap`...

@inercia if the user does not copy the template at minimum it will fail. checking for valid values is a bit harder? aka, if they do `init --cloud-provider openstack` and then try to run bootstrap this will fail and tell them they need the config file.

PR #505, opened 2019-07-23
Enable upgrade via intermediate pattern

MaximilianMeister:
To achieve kubeadm being upgraded without upgrading kubelet
and the container runtime, we decided to go with an inter-
mediate pattern "patterns-caasp-Node-[OldVersion]-[NewVersion]"

Fixes https://github.com/SUSE/avant-garde/issues/607

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/kubernetes.go
+++ b/internal/pkg/skuba/deployments/ssh/kubernetes.go
@@ -22,8 +22,10 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/version"
 
 	"github.com/SUSE/skuba/internal/pkg/skuba/deployments"
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
 )
 
 type KubernetesUploadSecretsErrorBehavior uint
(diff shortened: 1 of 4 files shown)


MaximilianMeister:
It still depends on the availability of some pattern rpm's in the build service

MaximilianMeister:
> It still depends on the availability of some pattern rpm's in the build service

the patterns are available now

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -76,8 +79,23 @@ func kubernetesInstallNodePattern(t *Target, data interface{}) error {
 	if !ok {
 		return errors.New("couldn't access kubernetes base OS configuration")
 	}
+
+	_, _, err := t.ssh("zypper", "--non-interactive", "install", "--force",
+		fmt.Sprintf("patterns-caasp-Node-%s", kubernetes.MajorMinorVersion(version.MustParseSemantic(kubernetesBaseOSConfiguration.KubernetesVersion))),
+		fmt.Sprintf("kubernetes-kubeadm=%s", kubernetesBaseOSConfiguration.KubernetesVersion),
```
flavio: I've nothing against that, but do we actually need to specify the kubeadm version to be installed? The pattern should already lead to the right version to be selected.

MaximilianMeister: if you install the `1.14` pattern, and kubernetes is relaxed to allow kubeadm `1.15` you will always get kubeadm `1.15` here. so we need this otherwise it's impossible to bootstrap a `1.14` cluster

MaximilianMeister: updated with a comment

PR #508, opened 2019-07-23
Openstack config file gets made read-only on bootstrap and join

Klaven:
## Why is this PR needed?
Currently once deployed your openstack config file becomes read only on the machine you run init from. This file should maintain its edit-ability from the deployment machine for the owner of the file.

### DIFF
--- a/pkg/skuba/actions/cluster/init/init.go
+++ b/pkg/skuba/actions/cluster/init/init.go
@@ -87,6 +87,7 @@ func Init(initConfiguration InitConfiguration) error {
 			return errors.Wrap(err, "unable to render template")
 		}
 		f.WriteString(str)
+		f.Chmod(0600)
 		f.Close()
 	}
 
--- a/pkg/skuba/cloud/cloud.go
+++ b/pkg/skuba/cloud/cloud.go
@@ -25,13 +25,26 @@ import (
 	"github.com/SUSE/skuba/pkg/skuba"
 )
 
+// HasCloudIntegration checks if the cluster has a cloud integration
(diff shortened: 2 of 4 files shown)


pkg/skuba/cloud/cloud.go
```diff
@@ -25,9 +25,9 @@ import (
 	"github.com/SUSE/skuba/pkg/skuba"
 )
 
+// HasCloudIntegration checks if the cluster has a cloud integration
 func HasCloudIntegration() bool {
 	if _, err := os.Stat(skuba.OpenstackCloudConfFile()); err == nil {
-		os.Chmod(skuba.OpenstackCloudConfFile(), 0400)
```
mjura: If `read-only` is issue then please change it to `read-write` `0600` for `user`. Restricted privileges for files are always good in case when you store credentials in it.

Klaven: Ok, after further discussions with @mjura we decided we will update skuba to write the `.template` file with 0600 permissions and then add a failing check on the `.conf` file if it's permissions are too permissive like `ssh` does.

pkg/skuba/actions/cluster/init/init.go
```diff
@@ -87,6 +87,7 @@ func Init(initConfiguration InitConfiguration) error {
 			return errors.Wrap(err, "unable to render template")
 		}
 		f.WriteString(str)
+		f.Chmod(0600)
```
Klaven: @mjura this would make all scaffold files have the 600 permissions. I currently can not think of an issue with this. and seemed to work just fine.

PR #516, opened 2019-07-25
Disable kured and skuba-update.timer during upgrade

MaximilianMeister:
## Why is this PR needed?

as a follow up to SUSE/skuba#505 we have to make sure to stop skuba-update.timer during the upgrade to prevent package updates during a node upgrade

and to prevent kured rebooting a node we have to flag the node temporarily during a node upgrade

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/kubernetes.go
+++ b/internal/pkg/skuba/deployments/ssh/kubernetes.go
@@ -20,9 +20,12 @@ package ssh
 import (
 	"fmt"
 	"path/filepath"
+	"time"
 
 	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/version"
+	"k8s.io/klog"
 
 	"github.com/SUSE/skuba/internal/pkg/skuba/deployments"
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
@@ -33,6 +36,7 @@ type KubernetesUploadSecretsErrorBehavior uint
 const (
 	KubernetesUploadSecretsFailOnError     KubernetesUploadSecretsErrorBehavior = iota
 	KubernetesUploadSecretsContinueOnError KubernetesUploadSecretsErrorBehavior = iota
+	kubeletTimeOutWait                                                          = 300
 )
 
 func init() {
@@ -42,6 +46,7 @@ func init() {
 	stateMap["kubernetes.install-node-pattern"] = kubernetesInstallNodePattern
 	stateMap["kubernetes.install-intermediate-node-pattern"] = kubernetesInstallIntermediateNodePattern
 	stateMap["kubernetes.restart-services"] = kubernetesRestartServices
+	stateMap["kubernetes.wait-for-kubelet"] = kubernetesWaitForKubelet
 }
 
 func kubernetesUploadSecrets(errorHandling KubernetesUploadSecretsErrorBehavior) Runner {
@@ -104,4 +109,21 @@ func kubernetesInstallIntermediateNodePattern(t *Target, data interface{}) error
 func kubernetesRestartServices(t *Target, data interface{}) error {
 	_, _, err := t.ssh("systemctl", "restart", "crio", "kubelet")
 	return err
+}
+
+func kubernetesWaitForKubelet(t *Target, data interface{}) error {
+	clientSet, err := kubernetes.GetAdminClientSet()
+	if err != nil {
+		return errors.Wrap(err, "Error getting client set")
+	}
+	for i := 0; i < kubeletTimeOutWait; i++ {
+		_, err := clientSet.CoreV1().Nodes().Get(t.target.Nodename, metav1.GetOptions{})
+		if err != nil {
+			klog.V(1).Infof("Still waiting for %s node to be registered, continuing...", t.target.Nodename)
+		} else {
+			return nil
+		}
+		time.Sleep(1 * time.Second)
+	}
+	return errors.Wrap(err, fmt.Sprintf("Timed out waiting for %s", t.target.Nodename))
 }
--- a/internal/pkg/skuba/deployments/ssh/kured.go
+++ b/internal/pkg/skuba/deployments/ssh/kured.go
@@ -22,13 +22,24 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog"
 
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
 	"github.com/SUSE/skuba/pkg/skuba"
 )
 
 func init() {
 	stateMap["kured.deploy"] = kuredDeploy
+	stateMap["kured.lock"] = kuredLock
+	stateMap["kured.unlock"] = kuredUnlock
 }
+
+const (
+	kuredDSName             = "kured"
+	kuredLockAnnotationJson = `{"metadata":{"annotations":{"weave.works/kured-node-lock":"'{\"nodeID\":\"manual\"}'"}}}`
+)
 
 func kuredDeploy(t *Target, data interface{}) error {
 	kuredFiles, err := ioutil.ReadDir(skuba.KuredDir())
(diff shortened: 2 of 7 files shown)


MaximilianMeister:
waiting for kubelet works fine:

```
I0725 13:49:26.215286    7130 states.go:35] === applying state kubernetes.restart-services ===
I0725 13:49:26.376607    7130 ssh.go:167] running command: "sudo sh -c 'systemctl restart crio kubelet'"
I0725 13:49:57.761664    7130 states.go:40] === state kubernetes.restart-services applied successfully ===
I0725 13:49:57.761719    7130 states.go:35] === applying state kubernetes.wait-for-kubelet ===
I0725 13:49:58.248354    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:49:59.733059    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:01.218698    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:02.702622    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:04.187240    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:05.670931    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:07.154887    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:08.639162    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:10.122821    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:11.606767    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:13.091153    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:14.576023    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:16.059411    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:17.544084    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:19.030748    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:20.515000    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:21.998981    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:23.483410    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:24.967028    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:26.450791    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:27.935007    7130 kubernetes.go:121] failed to get node my-master-0, continuing...
I0725 13:50:29.803416    7130 states.go:40] === state kubernetes.wait-for-kubelet applied successfully ===
```

internal/pkg/skuba/deployments/ssh/kured.go
```diff
@@ -47,3 +50,13 @@ func kuredDeploy(t *Target, data interface{}) error {
 	_, _, err = t.ssh("kubectl --kubeconfig=/etc/kubernetes/admin.conf apply -f /tmp/kured.d")
 	return err
 }
+
+func kuredLock(t *Target, data interface{}) error {
+	_, _, err := t.ssh("kubectl", "--kubeconfig=/etc/kubernetes/admin.conf", "-n", metav1.NamespaceSystem, "annotate", "ds", "kured", "weave.works/kured-node-lock='{\"nodeID\":\"manual\"}'")
```
lcavajani: isn't this going to fail if the target is a worker since there is no `/etc/kubernetes/admin.conf` ?

MaximilianMeister: good point, this should only be run on a master

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -105,3 +109,20 @@ func kubernetesRestartServices(t *Target, data interface{}) error {
 	_, _, err := t.ssh("systemctl", "restart", "crio", "kubelet")
 	return err
 }
+
+func kubernetesWaitForKubelet(t *Target, data interface{}) error {
+	clientSet, err := kubernetes.GetAdminClientSet()
+	if err != nil {
+		return errors.Wrap(err, "Error getting client set")
+	}
+	for i := 0; i < 300; i++ {
```
mssola: Could we move this integer into a proper constant?

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -105,3 +109,20 @@ func kubernetesRestartServices(t *Target, data interface{}) error {
...
+		_, err := clientSet.CoreV1().Nodes().Get(t.target.Nodename, metav1.GetOptions{})
+		if err != nil {
+			klog.V(1).Infof("failed to get node %s, continuing...", t.target.Nodename)
+		} else {
+			return nil
+		}
+		time.Sleep(1 * time.Second)
+	}
```
mssola: What happens if after all this time we are still getting errors? (let's consider the case in which the Kubelet is utterly broken).  We should be able to return a proper error (i.e. a timeout error) instead of just `nil`

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -107,6 +115,17 @@ func Apply(target *deployments.Target) error {
 			if err := target.Apply(nil, "kubernetes.restart-services"); err != nil {
 				return err
 			}
+			if err := target.Apply(nil, "kubernetes.wait-for-kubelet"); err != nil {
+				return err
```
mssola: What happens with the lock if the Kubelet failed to start? I wonder if we should then free the lock and annotate the node somehow :thinking:

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -140,6 +167,17 @@ func Apply(target *deployments.Target) error {
...
```
mssola: Same here

pkg/skuba/actions/node/upgrade/apply.go
MaximilianMeister: the lock will still be present, as the `apply` command will exit here. i'm not sure how to best continue in such a case, but if kubelet is not able to register the node, there's no chance you can perform anything at that stage. for best effort we can enhance the logging and print out a message to be aware of that i'd say

internal/pkg/skuba/deployments/ssh/kured.go
MaximilianMeister: i've used kubernetes client libraries to change the annotations now. so we're independent of the kubeconfig

internal/pkg/skuba/deployments/ssh/kubernetes.go
MaximilianMeister: done

pkg/skuba/actions/node/upgrade/apply.go
MaximilianMeister: i've added some logging now

MaximilianMeister:
> do you think you could fill the PR comment out. it is empty for me. Just a little "this is why we are doing this" would be great! thank you.

updated :+1:

MaximilianMeister:
> We have to ensure a node that has been previously locked by the user doesn't end unlocked as a result of our actions.
> 
> The current code doesn't take that into account. We had similar issues in the past with v3 that was uncordoning nodes the user had manually cordoned.

it's addressed now, thanks for the hint

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -120,13 +130,16 @@ func Apply(target *deployments.Target) error {
...
 			}
+			// FIXME: check if skuba-update was already disabled
 			err = target.Apply(nil, "skuba-update.start")
 			if err != nil {
 				return err
 			}
-			err = target.Apply(nil, "kured.unlock")
-			if err != nil {
-				return err
+			if !kuredLocked {
```
nirmoy: This should be if `if kuredLocked {`

internal/pkg/skuba/deployments/ssh/systemd.go
```diff
@@ -0,0 +1,32 @@
...
+package ssh
+
+import (
+	"k8s.io/klog"
+)
+
+// IsServiceEnabled returns if a service is enabled
+func (t *Target) IsServiceEnabled(serviceName string) (bool, error) {
+	klog.V(1).Info("checking if skuba-update.timer is enabled")
+	if stdout, _, err := t.silentSsh("systemctl", "is-enabled", serviceName); stdout == "enabled" {
```
MaximilianMeister: we still have an issue here... `silentSsh` will populate `err` if the service is disabled, because of the exit-code (1).

we either have to disregard the exit code in this case, or we find another command that tells us the same thing but with a (0) exit code

MaximilianMeister: we could do sth like:

```
t.silentSsh("systemctl", "is-enabled", serviceName, "||", ":")
```

but then we will never catch or return an error

MaximilianMeister: addressed in a separate commit. feel free to comment on this or suggest an alternative

internal/pkg/skuba/deployments/ssh/systemd.go
```diff
@@ -0,0 +1,32 @@
...
+package ssh
+
+import (
+	"k8s.io/klog"
+)
+
+// IsServiceEnabled returns if a service is enabled
+func (t *Target) IsServiceEnabled(serviceName string) (bool, error) {
+	klog.V(1).Info("checking if skuba-update.timer is enabled")
+	if stdout, _, err := t.silentSsh("systemctl", "is-enabled", serviceName, "||", ":"); stdout == "enabled" {
```
flavio: I don't like to check against a string. The string might change in the future or could be something completely different if the user has the system running in a language that is not English.

`t.silentSsh` is just a wrapper around [Session.Wait](https://godoc.org/golang.org/x/crypto/ssh#Session.Wait). According to the documentation:

> If the remote server does not send an exit status, an error of type *ExitMissingError is returned. If the command completes unsuccessfully or is interrupted by a signal, the error is of type *ExitError. Other error types may be returned for I/O problems.

I would instead look at the returned `error` and check if this is an instance of [*ExitError](https://godoc.org/golang.org/x/crypto/ssh#ExitError). If that's the case then look at the exit code of the remote command using the [ExitStatus](https://godoc.org/golang.org/x/crypto/ssh#Waitmsg.ExitStatus) method.

If the exit status == 1 -> the service was running, otherwise return the service is not running.

If the returned error is not of type `ExitStatus` then just propagate the error

MaximilianMeister: @flavio updated, thanks for the hint. please have another look

PR #519, opened 2019-07-25
Provide the ability to disable port security


evrardjp:
For some fringe cases, the openstack network where the
instanciated caasp lies might have port security disabled,
or the deployer want to disable port security when creating
the network all together.

Without this patch, it is impossible to create a network
with port security disabled, and attaching security groups
to a port security disabled network leads to failures in
the OpenStack API.

This fixes the port security case by adding a boolean,
defaulting to true, to determine if port security is enabled,
and use its value when creating network or attaching security
groups.

### DIFF
--- a/ci/infra/openstack/load-balancer.tf
+++ b/ci/infra/openstack/load-balancer.tf
@@ -2,9 +2,9 @@ resource "openstack_lb_loadbalancer_v2" "lb" {
   name          = "${var.stack_name}-lb"
   vip_subnet_id = "${openstack_networking_subnet_v2.subnet.id}"
 
-  security_group_ids = [
+  security_group_ids = "${var.enable_openstack_port_security == true ? [
     "${openstack_compute_secgroup_v2.secgroup_master_lb.id}",
-  ]
+  ] : [] }"
 }
 
 resource "openstack_lb_listener_v2" "kube_api_listener" {
--- a/ci/infra/openstack/load-balancer.tf
+++ b/ci/infra/openstack/load-balancer.tf
@@ -2,9 +2,7 @@ resource "openstack_lb_loadbalancer_v2" "lb" {
   name          = "${var.stack_name}-lb"
   vip_subnet_id = "${openstack_networking_subnet_v2.subnet.id}"
 
-  security_group_ids = "${var.enable_openstack_port_security == true ? [
-    "${openstack_compute_secgroup_v2.secgroup_master_lb.id}",
-  ] : [] }"
+  security_group_ids = "${var.enable_openstack_port_security == true ? ["${openstack_compute_secgroup_v2.secgroup_master_lb.id}", ] : []}"
 }
 
 resource "openstack_lb_listener_v2" "kube_api_listener" {
--- a/ci/infra/openstack/master-instance.tf
+++ b/ci/infra/openstack/master-instance.tf
@@ -44,7 +44,7 @@ data "template_file" "master-cloud-init" {
     register_rmt    = "${join("\n", data.template_file.master_register_rmt.*.rendered)}"
     commands        = "${join("\n", data.template_file.master_commands.*.rendered)}"
     username        = "${var.username}"
-    ntp_servers     = "${join("\n", formatlist ("    - %s", var.ntp_servers))}"
+    ntp_servers     = "${join("\n", formatlist("    - %s", var.ntp_servers))}"
   }
 }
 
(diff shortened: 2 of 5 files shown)


evrardjp:
Reversed the order of type and default to make lint happy.

evrardjp:
Changed email address, and  configured boolean as string to pass linting. Type is removed as it is inferred from default value.

ci/infra/openstack/worker-instance.tf
```diff
@@ -76,10 +76,10 @@ resource "openstack_compute_instance_v2" "worker" {
     name = "${var.internal_net}"
   }
 
-  security_groups = [
+  security_groups = "${var.enable_openstack_port_security == true ? [
```
bergmannf: I think this is missing the closing `}`.
```suggestion
  security_groups = "${var.enable_openstack_port_security} == true ? [
```

ci/infra/openstack/master-instance.tf
```diff
@@ -64,10 +64,10 @@ resource "openstack_compute_instance_v2" "master" {
     name = "${var.internal_net}"
   }
 
-  security_groups = [
+  security_groups = "${var.enable_openstack_port_security == true ? [
```
bergmannf: I think this is missing the closing `{`
```suggestion
  security_groups = "${var.enable_openstack_port_security} == true ? [
```

evrardjp:
Pushed again, because whitespaces. Thanks make lint.

evrardjp:
That won't work with the current packaging, would you be happy to bump the version required to at least 1.17 for the terraform-provider-openstack?
That would still keep terraform to .11 vs .12.

ci/infra/openstack/load-balancer.tf
```diff
@@ -2,9 +2,7 @@ resource "openstack_lb_loadbalancer_v2" "lb" {
   name          = "${var.stack_name}-lb"
   vip_subnet_id = "${openstack_networking_subnet_v2.subnet.id}"
 
-  security_group_ids = [
-    "${openstack_compute_secgroup_v2.secgroup_master_lb.id}",
-  ]
+  security_group_ids = "${var.enable_openstack_port_security == true ? ["${openstack_compute_secgroup_v2.secgroup_master_lb.id}", ] : []}"
```
evrardjp: This is wrong and I will fix it when I know we can bump terraform-provider-openstack

qnetter:
I'm a little confused - does adopting this change require updating terraform and/or the terraform openstack provider? Or can it be applied to the versions we are already shipping?

qnetter:
Also, @evrardjp - these questions via @mdarnell-suse...

Is the use case for this defect a deployment of CaaSP on top of OpenStack? 

Does this defect only show up if OpenStack port security groups are disabled? 

And is this defect particular to an ironic deployment of CaaSP or an instance-based CaaSP deployment?

jsuchome:
I cannot answer the questions for @evrardjp , but at least something from my POV: This change is not doing anything to default skuba deployments, right? It's only adding an option for disabling port security, but user has to specifically choose it as terraform variable...

evrardjp:
Sorry I am on holidays, so I missed this. Here are my answers

> Does adopting this change require updating terraform and/or the terraform openstack provider?

Yes. Terraform openstack provider needs to be bumped. There is still one version that's available for terraform 0.11 doing this kind of feature (I can double check which version when I am back from holidays). Of course, a bump of terraform to 0.12 would definitely help, as conditional code behaviour is greatly improved in that version.

> Or can it be applied to the versions we are already shipping?
It cannot, as the openstack provider is ... ancient.

> Also, @evrardjp - these questions via @mdarnell-suse...
> 
> Is the use case for this defect a deployment of CaaSP on top of OpenStack?

It's to cover a specific use case, when someone wants to create virtual ips, virtual nics in openstack.
Port security prevents this, and therefore needs to be disabled. Instead of doing so manually on each port after the fact, we should IMO support it from the start by adding a few conditionals.

> 
> Does this defect only show up if OpenStack port security groups are disabled?

Well, if port security groups are enabled, you can't have your own vip around, having the ability to disable port security on the network would basically allow any user workload.

> 
> And is this defect particular to an ironic deployment of CaaSP or an instance-based CaaSP deployment?

That could match both, assuming the ironic deployed one is using neutron to create the network (which is not mandatory, afaik).

toabctl:
@evrardjp @Itxaka needs rebase...

evrardjp:
> @evrardjp @Itxaka needs rebase...

Needs work in general, as this won't work with the current state -- We just need to clarify supported versions of terraform and its openstack provider.

ereslibre:
This PR has some conflicts that need resolving.

evrardjp:
I will refresh this based on 0.12, when 0.12 support lands.

evrardjp:
We should not have this until terraform is moved to 0.12.

jordimassaguerpla:
Hi!
Happy hacking

evrardjp:
Oh wow that's ancient. I suppose not many people want to deploy things that rely on specific network stack on top of kubernetes on top of openstack...

PR #522, opened 2019-07-25
Refactor platform

tdaines42:
## Why is this PR needed?

We had some generic functions in the Terraform class that should of been in the Platform class

Fixes https://github.com/SUSE/avant-garde/issues/575

## What does this PR do?

Makes the platform class the superclass of Terraform and move certain funtions there.
Also some refactoring

### DIFF
--- a/ci/infra/testrunner/platforms/platform.py
+++ b/ci/infra/testrunner/platforms/platform.py
@@ -1,24 +1,145 @@
-import sys
+import logging
+import os
+import shutil
 
-from platforms.openstack import Openstack
-from platforms.vmware import VMware
-from utils import Format
+from timeout_decorator import timeout
+
+from utils import (step, Utils)
+
+logger = logging.getLogger('testrunner')
 
 
 class Platform:
+    def __init__(self, conf):
+        self.conf = conf
+        self.utils = Utils(conf)
 
-    @staticmethod    
-    def get_platform(conf, platform):
-        if platform.lower() == "openstack":
-            platform = Openstack(conf)
-        elif platform.lower() == "vmware":
-            platform = VMware(conf)
-        elif platform.lower() == "bare-metal":
-            raise Exception("bare-metal is not available")
-        elif platform.lower() == "libvirt":
-            raise Exception("libvirt is not available")
-        else:
-            raise Exception("Platform Error: {} is not recognized".format(platform))
+        # Which logs will be collected from the nodes
+        self.logs = {"files": [],
+                     "dirs": ["/var/log/pods"],
+                     "services": ["kubelet"]}
 
-        return platform
+        # Files that will be deleted during the cleanup stage
+        self.tmp_files = []
 
+    @step
+    def cleanup(self):
+        """Clean up"""
+        try:
+            self._cleanup_platform()
+        except Exception as ex:
+            logger.warning("Received the following error '{}'\nAttempting to finish cleanup".format(ex))
+            raise Exception("Failure(s) during cleanup")
+        finally:
+            self._cleanup_files(self.tmp_files)
+
+    @timeout(600)
+    @step
+    def gather_logs(self):
+        logging_errors = False
+
+        node_ips = {"master": self.get_nodes_ipaddrs("master"),
+                    "worker": self.get_nodes_ipaddrs("worker")}
+
+        if not os.path.isdir(self.conf.log_dir):
+            os.mkdir(self.conf.log_dir)
+            logger.info(f"Created log dir {self.conf.log_dir}")
+
+        for node_type in node_ips:
+            for ip_address in node_ips[node_type]:
+                node_log_dir = self._create_node_log_dir(ip_address, node_type, self.conf.log_dir)
+                logging_error = self.utils.collect_remote_logs(ip_address, self.logs, node_log_dir)
+
+                if logging_error:
+                    logging_errors = logging_error
+
+        platform_log_error = self._get_platform_logs()
+
+        if platform_log_error:
+            logging_errors = platform_log_error
+
+        return logging_errors
+
+    def get_lb_ipaddr(self):
+        """
+        Get the IP of the Load Balancer
+        :return:
+        """
+        pass
+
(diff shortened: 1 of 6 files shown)


ci/infra/testrunner/platforms/platform.py
```diff
@@ -1,28 +1,100 @@
...
-            print("Todo: libvirt is not available")
-            sys.exit(0)
-        else:
-            raise Exception(Format.alert("Platform Error: {} is not applicable".format(platform)))
-
-        return platform
+    def __init__(self, conf, provisioner):
+        self.conf = conf
+        self.utils = Utils(conf)
+        self.provisioner = provisioner
```
pablochacin: Given the current separation of concerns, I think the provisioner should be used by Terraform, not by the Platform class.

ci/infra/testrunner/platforms/platform.py
```diff
@@ -1,28 +1,100 @@
...
+        """ Clean up """
+        cleanup_failure = False
+        try:
+            self._cleanup_platform()
+        except Exception as ex:
+            cleanup_failure = True
+            print(Format.alert("Received the following error {}".format(ex)))
+            print("Attempting to finish cleanup")
+
+        self.provisioner.cleanup()
```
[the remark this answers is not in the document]
pablochacin: Alternatively, if `provider` is moved to the `Terraform` class, then this call would be done by Terraform as a way for handling provide specific (openstack, vmware) cleanup, making the separation of concern clearer.

ci/infra/testrunner/platforms/terraform.py
```diff
@@ -3,88 +3,26 @@
...
-from timeout_decorator import timeout
-
 from utils import (Format, step, Utils)
 
 
 class Terraform:
-    def __init__(self, conf, platform):
+    def __init__(self, conf, platform, env_setup):
         self.conf = conf
+        self.env_setup = env_setup if env_setup else ""
```
pablochacin: I don't see the advantage of using a variable instead of a function for setting the environment. A function is more flexible because it allows other setups.

ci/infra/testrunner/platforms/platform.py
```diff
@@ -1,24 +1,145 @@
...
+    @staticmethod
+    def _cleanup_files(files):
+        for file in files:
+            if os.path.exists(file):
+                logger.info(f"Cleaning up {file}")
+
+                if os.path.isfile(file):
+                    os.remove(file)
+                else:
+                    shutil.rmtree(file)
```
dannysauer: Is there any concern over what happens if a subsequent file in the `files` iterable was located beneath a directory removed by `rmtree`?  It might be a good idea to sort `files` to enforce processing leaf-first, but that's probably deferrable for a separate PR (if it's a potential problem to begin with).

Catching the relevant exception (and perhaps just emitting it as a logged warning) on failure to remove for whatever reason may also be good enough.  File not found - ignoreable.  Permission denied - maybe a problem. :)

ci/infra/testrunner/platforms/terraform.py
```diff
@@ -2,143 +2,65 @@
...
-                "dirs": ["/var/log/pods"],
-                "services": ["kubelet"]}
+    def destroy(self, variables):
+        cmd = "destroy -auto-approve"
 
-        if not os.path.isdir(self.conf.log_dir):
-            os.mkdir(self.conf.log_dir)
-            logger.info(f"Created log dir {self.conf.log_dir}")
+        for var in variables:
+            cmd += f" -var {var}"
```
dannysauer: Does `{var}` need any quoting / escaping here?

ci/infra/testrunner/platforms/terraform.py
```diff
@@ -2,143 +2,65 @@
...
-            if num_master > -1:
-                self.conf.master.count = num_master
-                logger.warning("   Masters:{} ".format(num_master))
-
-            if num_worker > -1:
-                self.conf.worker.count = num_worker
-                logger.warning("   Workers:{} ".format(num_worker))
-
+        exception = None
+        print("Init terraform")
```
dannysauer: Should this be a logger call, or does it need printed?

ci/infra/testrunner/platforms/terraform.py
```diff
@@ -2,143 +2,65 @@
...
+        self._run_terraform_command(plan_cmd)
+
+        try:
+            self._run_terraform_command(apply_cmd)
+        except Exception as ex:
+            exception = ex
+        finally:
+            self._fetch_terraform_output()
+            if exception:
+                raise exception
```
dannysauer: The `finally` will re-raise an exception generated inside the `except` block, so you could put a bare`raise` in the except and save a couple lines.  But it will also re-raise any unhandled exception.  So, since all exceptions are ultimately being raised, this could be simplified as `try/finally`, leaving the except and both raises out.  Either way, call that optional cleanup for later. :)

ci/infra/testrunner/platforms/platform.py
tdaines42: I think files that might of been deleted already should be caught by `if os.path.exists(file):`

ci/infra/testrunner/platforms/terraform.py
tdaines42: I would assume that anyone using this would do that where it's being used. Not always good to assume though I guess.

ci/infra/testrunner/platforms/terraform.py
tdaines42: yep must of missed that one during the rebase

tdaines42: I think cause it was removed I'll just remove it my next pr

ci/infra/testrunner/platforms/terraform.py
tdaines42: Oh haha yeah 😂

ci/infra/testrunner/platforms/platform.py
dannysauer: Oh yeah. Doh. :D

PR #525, opened 2019-07-26
Redirect Testrunner command outputs to log

pablochacin:
## Why is this PR needed?
Presently testrunner prints all output, including the output from commands it invokes (skuba, terraform) producing a large amount of indeferentiate printouts (errors, informative messages, detailed command execution output).

To improve the output handling, it should be redirected to a log whose verbosity level should be controlled by means of the configuration file.

Fixes # https://github.com/SUSE/avant-garde/issues/622

## What does this PR do?

- Consolidate all logic for executing shell commands in `Utils.runshellcommand` to ensure consistent treatment of outputs and consistent tracing of command execution for debugging purposes
- Replaces `print` statements in all libraries for log statements. Remove superfluous outputs. 
- Implement a helper class for setting up logger. Direct output to console and to a file. File always has DEBUG level. Console level is managed by configuration file and by cli option
- Capture command outputs and return to caller. In case of exceptions, include `strerr`

### DIFF
--- a/ci/infra/testrunner/README.md
+++ b/ci/infra/testrunner/README.md
@@ -91,8 +91,9 @@ usage:
     This script is meant to be run manually on test servers, developer desktops, or Jenkins.
     This script supposed to run on python virtualenv from testrunner. Requires root privileges.
     Warning: it removes docker containers, VMs, images, and network configuration.
-    
+
        [-h] [-v YAML_PATH] [-p {openstack,vmware,bare-metal,libvirt}]
+       [-l {DEBUG,INFO,WARNING,ERROR}]
        {info,get_logs,cleanup,provision,build-skuba,bootstrap,status,join-node,remove-node,reset-node,ssh,test}
        ...
 
@@ -122,6 +123,9 @@ optional arguments:
                         -v myconfig.yaml
   -p {openstack,vmware,bare-metal,libvirt}, --platform {openstack,vmware,bare-metal,libvirt}
                         The platform you're targeting. Defaults to openstack
+  -l {DEBUG,INFO,WARNING,ERROR}, --log-level {DEBUG,INFO,WARNING,ERROR}
+                        log level
+
 ```
 
 
--- a/ci/infra/testrunner/kubectl/kubectl.py
+++ b/ci/infra/testrunner/kubectl/kubectl.py
@@ -15,7 +15,6 @@ class Kubectl:
 
 
     def create_deployment(self, name, image):
-        print("create a new deployment {}".format(name))
         try:
             self._run_kubectl("create deployment {name}  --image={image}"
                                 .format(name=name, image=image))
@@ -24,7 +23,6 @@ class Kubectl:
 
 
     def scale_deployment(self, name, replicas):
-        print("scale deployment {}".format(name))
         try:
             self._run_kubectl("scale deployment {name} --replicas={replicas}"
                              .format(name=name, replicas=replicas))
@@ -33,7 +31,6 @@ class Kubectl:
 
 
     def expose_deployment(self, name, port, nodeType="NodePort"):
-        print("expose deployment {}".format(name))
         try:
             self._run_kubectl("expose deployment {name} --port={port} --type={nodeType}"
                              .format(name=name, port=port, nodeType=nodeType))
@@ -42,7 +39,6 @@ class Kubectl:
 
 
     def wait_deployment(self, name, timeout):
-        print("wait deployment {}".format(name))
         try:
             self._run_kubectl("wait --for=condition=available deploy/{name} --timeout={timeout}m"
                              .format(name=name, timeout=timeout))
@@ -51,7 +47,6 @@ class Kubectl:
 
 
     def count_available_replicas(self, name):
-        print("count available replicas of deployment {}".format(name))
         try:
             result = self._run_kubectl("get deployment/{name} | jq '.status.availableReplicas'"
                                        .format(name=name))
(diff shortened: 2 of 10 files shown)


ci/infra/testrunner/README.md
```diff
@@ -122,6 +123,9 @@ optional arguments:
                         -v myconfig.yaml
   -p {openstack,vmware,bare-metal,libvirt}, --platform {openstack,vmware,bare-metal,libvirt}
                         The platform you're targeting. Defaults to openstack
+  -l {DEBUG,INFO,WARNING,ERROR}, --log-level {DEBUG,INFO,WARNING,ERROR}
```
dannysauer: I get that this is consistent with `-p` above, but this is a weird way to document options.  Usually short and long options are next to each other, avoiding duplication of the option list.  Things like GNU utilities and Maria DB do it like this:
```suggestion
  -l, --log-level {DEBUG,INFO,WARNING,ERROR}
```
Probably goes in another PR, but I figured it might be worth mentioning...

ci/infra/testrunner/utils/utils.py
```diff
@@ -139,14 +119,36 @@ def rsync(self, ip_address, remote_dir_path, local_dir_path):
...
+
+        if not os.path.isabs(cwd):
+            cwd = os.path.join(self.conf.workspace, cwd)
+
+        if not os.path.exists(cwd):
+            raise Exception(Format.alert("Directory {} does not exists".format(cwd)))
+
+        logger.info("Executing command: {} > {}".format(cwd, cmd))
+        p = subprocess.Popen(cmd, env=env, cwd=cwd, shell=True,
+                stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
dannysauer: Making stdout and stderr PIPEs is the behavior caused by `capture_output=True`, but specifying capture_output might be a little more readable.  That would mean using subprocess.run instead of directly calling subprocess.Popen, which is "only" Python 3.5 and newer - but SLES 15.1 is on 3.6.5, so that should be safe.  The other arguments in use are the same, so it'd be a drop-in - and the docs suggest that using `.run` is preferable to `.Popen` when possible.  It's up to you, though; just something to consider. :)

https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module

ci/infra/testrunner/utils/utils.py
```diff
@@ -155,35 +157,34 @@ def ssh_sock_fn(self):
...
         """Node info"""
-        print("Env vars: {}".format(sorted(os.environ)))
-
-        self.runshellcommand('ip a')
-        self.runshellcommand('ip r')
-        self.runshellcommand('cat /etc/resolv.conf')
+        info_lines  = "Env vars: {}\n".format(sorted(os.environ))
+        info_lines += self.runshellcommand('ip a')
+        info_lines += self.runshellcommand('ip r')
+        info_lines += self.runshellcommand('cat /etc/resolv.conf')
```
dannysauer: Running cat as a shell command is a little ugly.  Why not
```suggestion
        with open("/etc/resolv.conf", "r") as resolv
            info_lines += resolv.read()
```

dannysauer: I guess that'd likely also need a try block, and should probably be a separate PR which cleans up some other system commands which have native python equivalents (like `rm`).  Maybe nevermind. :)

ci/infra/testrunner/README.md
dannysauer: Ugh - this is Python's argparse formatting, isn't it?  So I guess that means messing with the help formatter...  So, definitely goes in a separate PR. :D
https://docs.python.org/3.6/library/argparse.html#formatter-class

pablochacin: I agree the help formatting should be improved. I'll open an issue for that.

ci/infra/testrunner/utils/utils.py
pablochacin: Opening an issue for this.

ci/infra/testrunner/utils/utils.py
```diff
@@ -139,14 +119,36 @@ def rsync(self, ip_address, remote_dir_path, local_dir_path):
...
+        if not cwd:
+            cwd = self.conf.workspace
+
+        if not os.path.isabs(cwd):
+            cwd = os.path.join(self.conf.workspace, cwd)
+
+        if not os.path.exists(cwd):
+            raise Exception(Format.alert("Directory {} does not exists".format(cwd)))
+
+        logger.info("Executing command: {} > {}".format(cwd, cmd))
```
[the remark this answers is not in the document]
pablochacin: I separated the info about executing the command from the detailed debug trace indicating the cdw and also environment variables

ci/infra/testrunner/kubectl/kubectl.py
```diff
@@ -61,30 +56,31 @@ def count_available_replicas(self, name):
...
+        shell_cmd = "curl {ip}:{port}{path}".format(ip=ip_address[worker],port=port,path=path)
         try:
-            return self.utils.runshellcommand_withoutput("curl {ip}:{port}{path}"
-                                                        .format(ip=ip_address[worker],port=port,path=path), False)
+            return self.utils.runshellcommand(sell_cmd, output=True)
         except Exception as ex:
-            raise Exception("Error executing cmd {}") from ex
-
-
+            raise Exception("Error testing service {] with path {} at node {}"
```
chentex: I see a `{ ]` instead of `{}`

PR #529, opened 2019-07-29
Add --strict-cap-defaults

drpaneas:
Changes the default CRI-O capabilities by adding the missing ones
from the docker container engine. In this way the customers will
be able to run the same containers that were working on v3
without any need for modification. At the same time though,
we are adding a new option at skuba '--strict-cap-defaults'.
This option will _not_ copy the docker defaults into CRI-O. Use
this option if you want a vanilla CRI-O experience.

## Why is this PR needed?

It fixes https://bugzilla.suse.com/show_bug.cgi?id=1142552
Fixes https://github.com/SUSE/avant-garde/issues/638

## What does this PR do?

By default it creates the dockerfied crio-configuration file during `skuba init` and it replaces the node's crio.conf with this file during `bootstrap` or `join`. The file can be found locally inside the `addons/cri/crio.conf`.

e.g.
```go
statesToApply[0]: kernel.load-modules
statesToApply[1]: kernel.configure-parameters
statesToApply[2]: apparmor.start
statesToApply[3]: cri.start
statesToApply[4]: kubelet.configure
statesToApply[5]: kubelet.enable
statesToApply[6]: kubeadm.join
statesToApply[7]: cni.cilium-update-configmap
statesToApply[8]: skuba-update.start


statesToApply[0]: kernel.load-modules
statesToApply[1]: kernel.configure-parameters
statesToApply[2]: apparmor.start
statesToApply[3]: cri.configure // this is now by default
statesToApply[4]: cri.start
statesToApply[5]: kubelet.configure
statesToApply[6]: kubelet.enable
statesToApply[7]: kubeadm.join
statesToApply[8]: cni.cilium-update-configmap
statesToApply[9]: skuba-update.start
```

However, if someone wants pure CRI-O experience, they can use `--strict-cap-defaults` during `skuba init`. This will do use the CRI-O default configuration.

Long story short, now the v3 containers will work to v4 as well without any need for modification in terms of capabilities.

Here's my testcase:

```yaml
---
apiVersion: v1
kind: Pod
metadata:
  name: leap
spec:
  containers:
  - name: app
    image: opensuse/leap:latest
    command: ['/bin/sh', '-c', 'sleep 3600']
---
apiVersion: v1
kind: Pod
metadata:
  name: sle12sp4
spec:
  containers:
  - name: app
    image: registry.suse.com/suse/sles12sp4:latest
    command: ['/bin/sh', '-c', 'sleep 3600']
---
apiVersion: v1
kind: Pod
metadata:
  name: sle15
spec:
  containers:
  - name: app
    image: registry.suse.com/suse/sle15:latest
    command: ['/bin/sh', '-c', 'sleep 3600']
---
apiVersion: v1
kind: Pod
metadata:
  name: sle15sp1
spec:
  containers:
  - name: app
    image: registry.suse.de/suse/containers/sle-server/15/containers/suse/sle15:15.1
    command: ['/bin/sh', '-c', 'sleep 3600']
```

```bash
kubectl apply -f test.yaml
declare -a arr=("sle12sp4" "leap" "sle15" "sle15sp1")
for pod in "${arr[@]}"; do echo -n "$pod: "; kubectl exec -it $pod -- su root -c id; done
for pod in "${arr[@]}"; do echo -n "$pod: "; kubectl exec -it $pod -- useradd panos; done
```

There must be no error from the above commands. Without this PR, the sles12sp4 containers fails due to PAM.

### DIFF
--- a/internal/app/skuba/cluster/init.go
+++ b/internal/app/skuba/cluster/init.go
@@ -38,6 +38,7 @@ type initOptions struct {
 	ControlPlane      string
 	KubernetesVersion string
 	CloudProvider     string
+	StrictCapDefaults bool
 }
 
 // NewInitCmd creates a new `skuba cluster init` cobra command
@@ -73,6 +74,7 @@ func NewInitCmd() *cobra.Command {
 				ImageRepository:     skuba.ImageRepository,
 				EtcdImageTag:        kubernetes.ComponentVersionForClusterVersion(kubernetes.Etcd, kubernetesVersion),
 				CoreDNSImageTag:     kubernetes.ComponentVersionForClusterVersion(kubernetes.CoreDNS, kubernetesVersion),
+				StrictCapDefaults:   initOptions.StrictCapDefaults,
 			}
 
 			err := cluster.Init(initConfig)
@@ -88,6 +90,8 @@ func NewInitCmd() *cobra.Command {
 	}
 	cmd.Flags().StringVar(&initOptions.CloudProvider, "cloud-provider", "", "Enable cloud provider integration with the chosen cloud. Valid values: openstack")
 	cmd.MarkFlagRequired("control-plane")
+
+	cmd.Flags().BoolVar(&initOptions.StrictCapDefaults, "strict-capability-defaults", false, "All the containers will start with CRI-O default capabilities")
 
 	return cmd
 }
--- a/internal/pkg/skuba/deployments/deployments.go
+++ b/internal/pkg/skuba/deployments/deployments.go
@@ -44,7 +44,15 @@ type Target struct {
 }
 
 func (t *Target) Apply(data interface{}, states ...string) error {
-	return t.Actionable.Apply(data, states...)
+	filteredStates := []string{}
+	for _, s := range states {
+		if s == "" {
+			continue
+		}
+		filteredStates = append(filteredStates, s)
+	}
+
+	return t.Actionable.Apply(data, filteredStates...)
 }
 
 func (t *Target) UploadFile(sourcePath, targetPath string) error {
--- a/internal/pkg/skuba/deployments/ssh/cri.go
+++ b/internal/pkg/skuba/deployments/ssh/cri.go
@@ -17,11 +17,49 @@
 
 package ssh
 
+import (
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/SUSE/skuba/pkg/skuba"
+	"github.com/pkg/errors"
+	"k8s.io/klog"
+)
+
 func init() {
+	stateMap["cri.configure"] = criConfigure
 	stateMap["cri.start"] = criStart
+	stateMap["cri.reset"] = crioReset
+}
+
+func criConfigure(t *Target, data interface{}) error {
+	criFiles, err := ioutil.ReadDir(skuba.CriDir())
+	if err != nil {
+		return errors.Wrap(err, "Could not read local cri directory: "+skuba.CriDir())
+	}
+	defer t.ssh("rm -rf /tmp/cri.d")
+
+	for _, f := range criFiles {
+		if err := t.target.UploadFile(filepath.Join(skuba.CriDir(), f.Name()), filepath.Join("/tmp/cri.d", f.Name())); err != nil {
+			return err
+		}
+	}
+
+	_, _, err = t.ssh("mv -f /etc/sysconfig/crio /etc/sysconfig/crio.backup")
+	_, _, err = t.ssh("mv -f /tmp/cri.d/default_flags /etc/sysconfig/crio")
+	return err
 }
 
 func criStart(t *Target, data interface{}) error {
 	_, _, err := t.ssh("systemctl", "enable", "--now", "crio")
 	return err
+}
+
+func crioReset(t *Target, data interface{}) error {
+	_, _, err := t.ssh("test -f /etc/sysconfig/crio.backup")
+	if err == nil {
+		_, _, _ = t.ssh("mv -f /etc/sysconfig/crio.backup /etc/sysconfig/crio")
+	}
+	klog.V(1).Info("CRI-O is running with default configuration already. Skipping reset...")
+	return nil
 }
--- a/pkg/skuba/actions/node/bootstrap/bootstrap.go
+++ b/pkg/skuba/actions/node/bootstrap/bootstrap.go
@@ -91,6 +91,11 @@ func Bootstrap(bootstrapConfiguration deployments.BootstrapConfiguration, target
 		return errors.Wrap(err, "error writing init configuration")
 	}
 
+	var criConfigure string
+	if _, err := os.Stat(skuba.CriDockerDefaultsConfFile()); err == nil {
+		criConfigure = "cri.configure"
+	}
+
 	fmt.Println("[bootstrap] applying init configuration to node")
 	err = target.Apply(
 		bootstrapConfiguration,
@@ -98,6 +103,7 @@ func Bootstrap(bootstrapConfiguration deployments.BootstrapConfiguration, target
 		"kernel.load-modules",
 		"kernel.configure-parameters",
 		"apparmor.start",
+		criConfigure,
 		"cri.start",
 		"kubelet.configure",
 		"kubelet.enable",
--- a/pkg/skuba/actions/node/join/join.go
+++ b/pkg/skuba/actions/node/join/join.go
@@ -56,10 +56,16 @@ func Join(joinConfiguration deployments.JoinConfiguration, target *deployments.T
 		return err
 	}
 
+	var criConfigure string
+	if _, err := os.Stat(skuba.CriDockerDefaultsConfFile()); err == nil {
+		criConfigure = "cri.configure"
+	}
+
 	statesToApply := []string{
 		"kernel.load-modules",
 		"kernel.configure-parameters",
 		"apparmor.start",
+		criConfigure,
 		"cri.start",
 		"kubelet.configure",
 		"kubelet.enable",
--- a/pkg/skuba/actions/cluster/init/constants.go
+++ b/pkg/skuba/actions/cluster/init/constants.go
@@ -65,6 +65,10 @@ var (
 			Location: skuba.GangwayManifestfile(),
 			Content:  gangwayManifest,
 		},
+		{
+			Location: skuba.CriDockerDefaultsConfFile(),
+			Content:  criDockerDefaultsConf,
+		},
 	}
 
 	cloudScaffoldFiles = map[string][]ScaffoldFile{
--- a/pkg/skuba/actions/cluster/init/init.go
+++ b/pkg/skuba/actions/cluster/init/init.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"text/template"
 
+	"github.com/SUSE/skuba/pkg/skuba"
 	"github.com/pkg/errors"
 	"k8s.io/klog"
 )
@@ -43,6 +44,7 @@ type InitConfiguration struct {
 	EtcdImageTag        string
 	CoreDNSImageTag     string
 	CloudProvider       string
+	StrictCapDefaults   bool
 }
 
 // Init creates a cluster definition scaffold in the local machine, in the current
@@ -72,6 +74,10 @@ func Init(initConfiguration InitConfiguration) error {
 		return errors.Wrapf(err, "could not change to cluster directory %q", initConfiguration.ClusterName)
 	}
 	for _, file := range scaffoldFilesToWrite {
+		// If '--strict-capability-defaults' is used, then don't modify the /etc/sysconfig/crio
+		if initConfiguration.StrictCapDefaults && file.Location == skuba.CriDockerDefaultsConfFile() {
+			continue
+		}
 		filePath, _ := filepath.Split(file.Location)
 		if filePath != "" {
 			if err := os.MkdirAll(filePath, 0700); err != nil {
--- a/pkg/skuba/actions/cluster/init/manifests.go
+++ b/pkg/skuba/actions/cluster/init/manifests.go
@@ -52,6 +52,14 @@ networking:
   serviceSubnet: 10.96.0.0/12
 useHyperKubeImage: true
 `
+	criDockerDefaultsConf = `## Path           : System/Management
+## Description    : Extra cli switches for crio daemon
+## Type           : string
+## Default        : ""
+## ServiceRestart : crio
+#
+CRIO_OPTIONS=--pause-image=registry.suse.de/devel/caasp/4.0/containers/containers/caasp/v4/pause:3.1 --default-capabilities="CHOWN,DAC_OVERRIDE,FSETID,FOWNER,NET_RAW,SETGID,SETUID,SETPCAP,NET_BIND_SERVICE,SYS_CHROOT,KILL,MKNOD,AUDIT_WRITE,SETFCAP"
+  `
 
 	masterConfTemplate = `apiVersion: kubeadm.k8s.io/v1beta1
 kind: JoinConfiguration
--- a/pkg/skuba/actions/node/reset/reset.go
+++ b/pkg/skuba/actions/node/reset/reset.go
@@ -26,5 +26,5 @@ import (
 // Reset the target node
 func Reset(resetConfiguration deployments.ResetConfiguration, target *deployments.Target) error {
 	fmt.Println("[reset] resetting the node")
-	return target.Apply(resetConfiguration, "kubeadm.reset")
+	return target.Apply(resetConfiguration, "cri.reset", "kubeadm.reset")
 }
--- a/pkg/skuba/constants.go
+++ b/pkg/skuba/constants.go
@@ -80,6 +80,14 @@ func PspDir() string {
 	return filepath.Join(AddonsDir(), "psp")
 }
 
+func CriDir() string {
+	return filepath.Join(AddonsDir(), "cri")
+}
+
+func CriDockerDefaultsConfFile() string {
+	return filepath.Join(CriDir(), "default_flags")
+}
+
 func PspUnprivManifestFile() string {
 	return filepath.Join(PspDir(), "podsecuritypolicy-unprivileged.yaml")
 }


internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -17,10 +17,37 @@
...
+
+	defer t.ssh("rm -rf /tmp/cri.d")
+
+	for _, f := range criFiles {
+		if err := t.target.UploadFile(filepath.Join(skuba.CriDir(), f.Name()), filepath.Join("/tmp/cri.d", f.Name())); err != nil {
+			return err
+		}
+	}
+
+	_, _, err = t.ssh("cp /tmp/cri.d/crio.conf /etc/crio/crio.conf")
```
nirmoy: This will replace existing cri conf. What happens if user custom tune crio before to run skuba. IIUC you want `default_capabilities` to have all the caps. Why not parse the fetch crio config file and modify?

drpaneas: > What happens if user custom tune crio before to run skuba
The whole point of this PR is to enhance the **out of the box** experience, so I guess if the user is _experienced enough_ to touch the `crio.conf` they should know what they are doing. _If I would have bought a system to bootstrap k8s I would let the system do its magic_. Modifying configuration files that are essential to the core functionality of the product is not something we can do much about it -- we can document them of course,

However, I understand and value your feedback, because I also agree that skuba is doing _far more_ things than just a plain _bootstrap_ & _join_. In the ML I have suggested to Rafa to introduce an _intermediate_ command called `skuba node prepare`. This will actually take care of all the configuration _shenanigans_. I think `crio.conf` would go there as well. If you are still not persuaded, think of the same reasoning for using Terraform in the first place as it applies here as well: we are providing _one of many_ ways to do it.

> Why not parse the fetch crio config file and modify?
Ideally yes, I would prefer to parse conf and modify it. I haven't figured out how to do it yet. My idea was to modify the `/etc/sysconfig/crio` with `sed` and `regex`, but I am still not having any successful results. Although this way (parse and modify) is more risky in comparison with a plain copy-paste of a file -- I would be in favor of it because of maintenance reason. Merging the current approach would be we would have to maintain the configuration (`crio.conf`) in 2 separate places (the crio RPM package + skuba package). It goes without question if you have figured out a better way of doing that, it's more than welcome.

nirmoy: > > What happens if user custom tune crio before to run skuba
> 
> The whole point of this PR is to enhance the **out of the box** experience, so I guess if the user is _experienced enough_ to touch the `crio.conf` they should know what they are doing. _If I would have bought a system to bootstrap k8s I would let the system do its magic_. Modifying configuration files that are essential to the core functionality of the product is not something we can do much about it -- we can document them of course,

We would need more than document. Think about a case when a node switching between two cluster one with and without `--strict-cap-defaults` option. That would require `skuba reset/remove` to restore `crio.conf` to its previous state. May be I am talking about some corner case here.

> It goes without question if you have figured out a better way of doing that, it's more than welcome.
Another option may be to parse the file and check if all required capabilities exist or not , when not and `--strict-cap-defaults` is set then throw error saying what skuba needs

internal/app/skuba/cluster/init.go
```diff
@@ -89,5 +91,7 @@ func NewInitCmd() *cobra.Command {
 	cmd.Flags().StringVar(&initOptions.CloudProvider, "cloud-provider", "", "Enable cloud provider integration with the chosen cloud. Valid values: openstack")
 	cmd.MarkFlagRequired("control-plane")
 
+	cmd.Flags().BoolVar(&initOptions.StrictCapDefaults, "strict-cap-defaults", false, "All the containers will start with CRI-O default capabilities")
```
Klaven: remove `cap` from this. while their product should run on ours, I don't think this setting should be named after it.

Klaven: that being said. I would do "strict by default" and say you turn this into `custom-ciro-config` I would then change the rest of this update to support that.

aka, by default it continues to work the same way it always has. If you want a less restrictive run you can run with `--custom-crio-config` (or something like that.)

drpaneas: `cap` in that case stands for capabilities (not SUSE CAP) :P 

I would also prefer by default to do the default cri-o stuff, but the PM and PO said otherwise.

internal/pkg/skuba/deployments/ssh/cri.go
drpaneas: I've changed the PR a bit. Now instead of copy-pasting the `/etc/crio/crio.conf`, I am changing only the suse specific file, which is `/etc/sysconfig/crio`. As a result, the capabilities listed in `/etc/crio/conf` are overwritten by:

```bash
CRIO_OPTIONS=--pause-image=registry.suse.de/devel/caasp/4.0/containers/containers/caasp/v4/pause:3.1 --default-capabilities="CHOWN,DAC_OVERRIDE,FSETID,FOWNER,NET_RAW,SETGID,SETUID,SETPCAP,NET_BIND_SERVICE,SYS_CHROOT,KILL,MKNOD,AUDIT_WRITE,SETFCAP"
```

In that case, even the user modifies the `/etc/crio/crio.conf` prior to running `skuba`, their configuration will remain intact. The only thing that will be overriden will be these two things:

1. The pause container
2. the default capabilities

I will document both of those in the docs.

As for the `skuba reset` I would prefer to deal with that in a different PR, as I think there's a bug there in general. What I could do in this PR in terms of reset is:

* either remove `/etc/sysconfig/crio` and restore it by running `sudo zypper in -y -f cri-o`.
* either create a backup of the file before replacing it.

> We would need more than document. Think about a case when a node switching between two cluster one with and without --strict-cap-defaults option. That would require skuba reset/remove to restore crio.conf to its previous state. May be I am talking about some corner case here.

No. You will never end up with different container engines into nodes, simply because the configuration is created at `skuba init`. Both `node bootstrap` and `node join` are using this configuration created by the `skuba init` earlier.

internal/app/skuba/cluster/init.go
MaximilianMeister: > `cap` in that case stands for capabilities (not SUSE CAP) :P

:man_facepalming: i thought it stands for `SUSE CAP`... this is quite misleading, do you think we could call it `--strict-capability-defaults` ?

pkg/skuba/actions/node/bootstrap/bootstrap.go
```diff
@@ -92,20 +92,39 @@ func Bootstrap(bootstrapConfiguration deployments.BootstrapConfiguration, target
...
-		"skuba-update.start",
-	)
+	if _, err := os.Stat(skuba.CriDockerDefaultsConfFile()); err == nil {
+		err = target.Apply(
+			bootstrapConfiguration,
+			"kubernetes.bootstrap.upload-secrets",
+			"kernel.load-modules",
+			"kernel.configure-parameters",
+			"apparmor.start",
+			"cri.configure",
```
inercia: This seems a bit redundant. Maybe you could have a variable `cri_states` that could be

```Go
criStates=[]string{"cri.start"}
if _, err := os.Stat(skuba.CriDockerDefaultsConfFile()); err == nil {
  criStates=append("cri.configure", criStates...)
}
```

and then something like:

```Go
err = target.Apply(
		bootstrapConfiguration,
		"kubernetes.bootstrap.upload-secrets",
		"kernel.load-modules",
		"kernel.configure-parameters",
		"apparmor.start",
		criStates,
		"kubelet.configure",
		"kubelet.enable",
		"kubeadm.init",
		"psp.deploy",
		"kured.deploy",
		"skuba-update.start",
	)
```

drpaneas: Let me fix that. Thanks Alvaro :D

pkg/skuba/actions/node/join/join.go
```diff
@@ -68,6 +68,12 @@ func Join(joinConfiguration deployments.JoinConfiguration, target *deployments.T
 		"skuba-update.start",
 	}
 
+	if _, err := os.Stat(skuba.CriDockerDefaultsConfFile()); err == nil {
+		statesToApply = append(statesToApply, "")
+		copy(statesToApply[3:], statesToApply[2:])
```
drpaneas: I don't like this hardcoded values. In case someone changes the order of states it might be a situation (by accident) where `crio.configure` will be called after the `crio.start`

internal/app/skuba/cluster/init.go
drpaneas: Sure I can do that.

internal/pkg/skuba/deployments/ssh/cri.go
drpaneas: Also, I've added a `reset` functionality where it restores the `crio.conf.backup` if it exists.

pkg/skuba/actions/node/bootstrap/bootstrap.go
drpaneas: I couldn't do exactly what you described because `criStatea` is an array and the `Apply` didn't like that. I did it, a bit differently though.

pkg/skuba/actions/node/join/join.go
drpaneas: Fixed now.

pkg/skuba/actions/node/bootstrap/bootstrap.go
inercia: Yeah, I understand, I was not completely sure you could do it that way, but anyway it looks much better in my view...

internal/pkg/skuba/deployments/ssh/cri.go
nirmoy: @drpaneas Looks way better now, Thanks!

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -17,11 +17,52 @@
...
+	}
+	defer t.ssh("rm -rf /tmp/cri.d")
+
+	for _, f := range criFiles {
+		if err := t.target.UploadFile(filepath.Join(skuba.CriDir(), f.Name()), filepath.Join("/tmp/cri.d", f.Name())); err != nil {
+			return err
+		}
+	}
+
+	_, _, err = t.ssh("cp /etc/sysconfig/crio /etc/sysconfig/crio.backup")
```
stefsuse: s/cp/mv

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -17,11 +17,52 @@
...
+	_, _, err = t.ssh("cp /tmp/cri.d/default_flags /etc/sysconfig/crio")
```
stefsuse: mv

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -17,11 +17,52 @@
...
 func init() {
+	stateMap["cri.configure"] = criConfigure
 	stateMap["cri.start"] = criStart
+	stateMap["cri.reset"] = crioReset
+}
+
+func criConfigure(t *Target, data interface{}) error {
+	criFiles, err := ioutil.ReadDir(skuba.CriDir())
+	if err != nil {
+		return errors.Wrap(err, "could not read local cri directory")
```
stefsuse: nice-to-have: print the bad directory

drpaneas: done, thx

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -17,11 +17,52 @@
...
+
+func criConfigure(t *Target, data interface{}) error {
+	criFiles, err := ioutil.ReadDir(skuba.CriDir())
+	if err != nil {
+		return errors.Wrap(err, "Could not read local cri directory: "+skuba.CriDir())
+	}
+	defer t.ssh("rm -rf /tmp/cri.d")
+
+	for _, f := range criFiles {
+		if err := t.target.UploadFile(filepath.Join(skuba.CriDir(), f.Name()), filepath.Join("/tmp/cri.d", f.Name())); err != nil {
```
inercia: why are you uploading file to a temporary directory? what is that `/tmp` dir used for? I only see you are copying `/tmp/cri.d/default_flags` from that directory...

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -17,11 +17,52 @@
...
+			return err
+		}
+	}
+
+	_, _, err = t.ssh("mv /etc/sysconfig/crio /etc/sysconfig/crio.backup")
```
inercia: Maybe it would be better to use a `mv -f`, just in case the backup file is already there (and we don't really care about it)

pkg/skuba/actions/node/join/join.go
```diff
@@ -68,6 +68,20 @@ func Join(joinConfiguration deployments.JoinConfiguration, target *deployments.T
...
+		for index, value := range statesToApply {
+			if value == "cri.start" {
+				beforeCriStart = index
+				break
+			}
+		}
+		statesToApply = append(statesToApply, "")
+		copy(statesToApply[beforeCriStart:], statesToApply[beforeCriStart-1:])
+		statesToApply[beforeCriStart] = "cri.configure"
+	}
```
inercia: In my view, this is a bit cryptic. I would use something like:
```Go
       criConfigure := ""
	if _, err := os.Stat(skuba.CriDockerDefaultsConfFile()); err == nil {
           criConfigure = "cri.configue"
       }

	statesToApply := []string{
		"kernel.load-modules",
		"kernel.configure-parameters",
		"apparmor.start",
                criConfigure,
		"cri.start",
		"kubelet.configure",
		"kubelet.enable",
		"kubeadm.join",
		"cni.cilium-update-configmap",
		"skuba-update.start",
	}
```

and make sure that we ignore empty states when applying them.

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -17,11 +17,52 @@
...
 func criStart(t *Target, data interface{}) error {
 	_, _, err := t.ssh("systemctl", "enable", "--now", "crio")
 	return err
 }
+
+func crioReset(t *Target, data interface{}) error {
+	_, _, err := t.ssh("test -f /etc/sysconfig/crio.backup")
+	if err == nil {
+		_, _, err = t.ssh("mv /etc/sysconfig/crio.backup /etc/sysconfig/crio")
+		if err != nil {
```
inercia: I think this will fail if 1) `/etc/sysconfig/crio` exists and 2) you do not use `mv -f`

Besides that, maybe you could skip the `test` and just go ahead and ` t.ssh("mv -f /etc/sysconfig/crio.backup /etc/sysconfig/crio")` ignoring any errors...

internal/pkg/skuba/deployments/ssh/cri.go
drpaneas: The same technique is used for other configurations within skuba:

- cni.go `	if err := t.target.UploadFile(filepath.Join(skuba.CniDir(), f.Name()), filepath.Join("/tmp/cni.d", f.Name())); err != nil {`

- dex.go `if err := t.target.UploadFile(filepath.Join(skuba.DexDir(), f.Name()), filepath.Join("/tmp/dex.d", f.Name())); err != nil {`

- gangway.go `	if err := t.target.UploadFile(filepath.Join(skuba.GangwayDir(), f.Name()), filepath.Join("/tmp/gangway.d", f.Name())); err != nil {`

- kubeadm.go `remoteKubeadmUpgradeConfFile := filepath.Join("/tmp/", skuba.KubeadmUpgradeConfFile())`

- kured.go `if err := t.target.UploadFile(filepath.Join(skuba.KuredDir(), f.Name()), filepath.Join("/tmp/kured.d", f.Name())); err != nil {`

- psp.go `	if err := t.target.UploadFile(filepath.Join(skuba.PspDir(), f.Name()), filepath.Join("/tmp/psp.d", f.Name())); err != nil {`

So for the sake of consistency, I followed the same pattern. This temporary directory gets deleted upon the completion of the state.

internal/pkg/skuba/deployments/ssh/cri.go
drpaneas: I added the `-f` but I didn't remove the error check. There's always something might go wrong  (e.g. `/etc/sysconfig/crio.backup` doesn't exist for some reason), so it doesn't harm.

pkg/skuba/actions/node/join/join.go
drpaneas: Indeed, this way better approach. I modified the `Apply()` function inside the `deployments.go` to exclude empty states. I tested it 2 times and it works fine :)

internal/pkg/skuba/deployments/ssh/cri.go
drpaneas: Done, thx :)

internal/pkg/skuba/deployments/ssh/cri.go
drpaneas: We would refactor this, but in another PR.

internal/pkg/skuba/deployments/deployments.go
```diff
@@ -44,6 +44,28 @@ type Target struct {
...
+			emptyStateFound = true
+		}
+	}
+
+	// Remove empty states so they don't get applied
+	if emptyStateFound {
+		for _, emptyState := range emptyStates {
+			states = remove(states, emptyState)
+		}
+	}
```
inercia: You could just filter the empty states like this:

```suggestion
    filteredStates := []string{}
    for _, s := range states {
        if s == "" {
            continue
        }   
        filteredStates = append(filteredStates, s)
    }
    return t.Actionable.Apply(data, newStates...)
```

or maybe it would make more sense to do this filtering inside of the `t.Actionale.Apply` function...

internal/pkg/skuba/deployments/ssh/cri.go
inercia: Ok, I see...

internal/pkg/skuba/deployments/ssh/cri.go
inercia: Well, it is a minor thing, but if you don't ignore the error, for exmaple a `reset` with no previous `node bootstrap/join` would lead to an error that could be ignorable... :confused:

internal/pkg/skuba/deployments/deployments.go
drpaneas: I love this. Less is more :D

internal/pkg/skuba/deployments/ssh/cri.go
dannysauer: The `-f` would only be necessary if `mv` is an alias to something like `mv -i` or `mv -n`, which should never be the case in a non-interactive shell.

Either way, why not join those two commands using the shell, reducing the time window for a race condition and cutting an extra `ssh` out of the process?  `test -f $file && mv $file $dst`

dannysauer:
The implementation looks fine, except...

I understand that we're shooting for compatibility with insecure-by-default Docker and that bsc#1142552 exists, but I'm specifically concerned by (re)adding the cap_audit_write permission to default containers.  That one means that an ostensibly unprivileged container which is either malicious or misconfigured can potentially cause host system audit logs to become unavailable -- and, because the audit log daemon operates at an elevated priority, means that a container flooding audit logs can also cause resources to be consumed on the host system which exceed the limits configured for that container.

The other capabilities can admittedly be a _part_ of a multi-step exploit in certain scenarios, but that one in particular concerns me because of how it facilitates direct impact outside the container.  I really wish we'd get an opinion from Security on that to weigh against the Sales opinion; shipping with insecure defaults can very quickly become a bigger black eye on a product than "you need to provide a command line option to retain backwards compatibility".

drpaneas:
Thanks everyone :)

qnetter:
I understand the hesitancy, especially Danny's. And we will hit the desired security target... in time.

PR #531, opened 2019-07-30
Fix aggressive cleanup exceptions in testrunner

pablochacin:
# Why is this PR needed?

The error handling logic in clean functions for terraform and
skuba generates exceptions in cases which may not be any critical
error such as trying to remove non-existing files.

## What does this PR do?

Reports exception only if critical errors occurs. Considers file removal failures as non critical, but reports them as warnings.

### DIFF
--- a/ci/infra/testrunner/platforms/terraform.py
+++ b/ci/infra/testrunner/platforms/terraform.py
@@ -32,29 +32,22 @@ class Terraform:
 
     def cleanup(self):
         """ Clean up """
-        cleanup_failure = False
         try:
             self._cleanup_platform()
         except Exception as ex:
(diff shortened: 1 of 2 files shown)


ci/infra/testrunner/platforms/terraform.py
```diff
@@ -32,29 +32,22 @@ def _get_platform_logs(self):
 
     def cleanup(self):
         """ Clean up """
-        cleanup_failure = False
         try:
             self._cleanup_platform()
         except Exception as ex:
             cleanup_failure = True
```
dannysauer: This could be removed too, right?  Maybe later.

PR #535, opened 2019-07-31
Fix env passing of TERRAFORM_STACK_NAME

MaximilianMeister:

### DIFF
--- a/ci/jenkins/pipelines/skuba-upgrade-nightly.Jenkinsfile
+++ b/ci/jenkins/pipelines/skuba-upgrade-nightly.Jenkinsfile
@@ -13,28 +13,20 @@ pipeline {
 
    stages {
         stage('Getting Ready For Cluster Deployment') { steps {
-            sh(script: 'make -f skuba/ci/Makefile pre_deployment', label: 'Pre Deployment')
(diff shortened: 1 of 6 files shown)


ci/jenkins/pipelines/skuba-upgrade-nightly.Jenkinsfile
```diff
@@ -13,28 +13,20 @@ pipeline {
 
    stages {
         stage('Getting Ready For Cluster Deployment') { steps {
-            sh(script: 'make -f skuba/ci/Makefile pre_deployment', label: 'Pre Deployment')
+            sh(script: "export TERRAFORM_STACK_NAME=${JOB_NAME}-prepare-${BUILD_NUMBER}; make -f skuba/ci/Makefile pre_deployment", label: 'Pre Deployment')
```
pablochacin: Platform should be set in the environment section , not here.

MaximilianMeister: you can only set an env once under `pipeline {}` (see https://jenkins.io/doc/pipeline/tour/environment/)

this is a workaround as we want to have parallel runs and need different `TERRAFORM_STACK_NAME`'s not to have naming clashes on openstack

PR #536, opened 2019-07-31
Fix testrunner file handling

tdaines42:
## Why is this PR needed?

We're using shell commands for things that are a part of the python standard library

Fixes https://github.com/SUSE/avant-garde/issues/644

## What does this PR do?

Switches calls to use the Python commands instead of using shell commands

### DIFF
--- a/ci/infra/testrunner/platforms/terraform.py
+++ b/ci/infra/testrunner/platforms/terraform.py
@@ -35,7 +35,6 @@ class Terraform(Platform):
     def _provision_platform(self):
         """ Create and apply terraform plan"""
         exception = None
-        print("Init terraform")
         self._check_tf_deployed()
 
         self.utils.setup_ssh()
--- a/ci/infra/testrunner/skuba/skuba.py
+++ b/ci/infra/testrunner/skuba/skuba.py
@@ -1,8 +1,8 @@
 import logging
 import os
 import platforms
+import shutil
 
-from platforms.platform import Platform
 from utils.format import Format
 from utils.utils import (step, Utils)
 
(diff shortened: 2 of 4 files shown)


ci/infra/testrunner/platforms/terraform.py
```diff
@@ -35,7 +35,6 @@ def destroy(self, variables):
     def _provision_platform(self):
         """ Create and apply terraform plan"""
         exception = None
-        print("Init terraform")
```
Klaven: nit: does not see related to the PR as a whole...

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -28,48 +28,42 @@ def _verify_bootstrap_dependency(self):
...
     @staticmethod
     @step
     def cleanup(conf):
         """Cleanup skuba working environment"""
-        utils = Utils(conf)
-        # TODO: check why (and if) the following two commands are needed
-        cmd = 'mkdir -p {}/logs'.format(conf.workspace)
-        utils.runshellcommand(cmd)
 
+        # TODO: check why (and if) the following command is needed
```
Klaven: nit: I prefer github issues over TODO's in code

Klaven: also. not sure what files this is chmod'ing BUT.... some files in skuba need to have restricted `600` permissions or skuba will reject them.

ci/infra/testrunner/utils/utils.py
```diff
@@ -33,6 +35,36 @@ class Utils:
...
+        os.chmod(directory, permissions)
+
+        for file in glob.glob(os.path.join(directory, "**/*"), recursive=True):
+            try:
+                os.chmod(file, permissions)
+            except Exception as ex:
+                logger.exception(ex)
+
+    @staticmethod
+    def cleanup_file(file):
```
pablochacin: I would have only one method: `cleanup files`.  No need for a method with a single file as argument.

ci/infra/testrunner/skuba/skuba.py
MaximilianMeister: i can also not explain this. the chmod should ideally only be files that are not tracked in the git tree, otherwise you will always have a git diff when you run testrunner within a clone. this is basically messing up the git clone and forces you to use testrunner only in a separate git clone, not to mess up the one you're mainly developing on. 

> nit: I prefer github issues over TODO's in code

:+1: an issue would be good to figure this out. when this is fixed i believe you could use testrunner also in your main git clone that you develop on, which makes things much easier

pablochacin: The `chmod` is applied to files in the `workspace` which is a working directory and should not be under git. 

Test runner can be used with the project clone from git.

ci/infra/testrunner/utils/utils.py
tdaines42: Yeah I just figured having one for a single file would be nicer in the long run. Rather than having to put a single file into an array just to remove it with the other function.

PR #540, opened 2019-08-01
AWS terraform files: refactoring

flavio:
## Why is this PR needed?

The current AWS deployment files are organized differently from the other ones. They are made by a single terraform file, instead of smaller one.

Having smaller files is better because:

  * It provides consistency with the other terraform deployments. This is good not only for developers but also for the end users.
  * Sooner or later we will have to reduce the duplication between all the different providers by abstracting common patterns. Having a similar structure would make the process easier.
  * Changes to the files are easier to inspect during review time.

## What does this PR do?

This PR splits the single `cluster.tf` file used by AWS into smaller files. This is done to be consistent with how the other deployments are structured.

Minor improvements have been done like: removing commented code that is not needed anymore, improve README and inline comments.

**Nothing** has been changed from the technical POV.

The AWS deployment is not yet available to SUSE CaaSP users. It's not even part of the official release.


inercia:
I've always thought that providing to our users one file instead of many small files would be better for the sake of clarity, but I admit this is a question of taste. Anyway, I think there is a deeper refactoring needed, and it is to move all this code to modules and stop repeating ourselves with things like cloud-init, outputs and so on. That would really enforce consistency between our deployment platforms.

flavio:
> Anyway, I think there is a deeper refactoring needed, and it is to move all this code to modules and stop repeating ourselves with things like cloud-init, outputs and so on. That would really enforce consistency between our deployment platforms.

Yes, I totally agree with that. This is a step into this direction

PR #550, opened 2019-08-05
Wrap $@ argument passing from testrunner in quotes

bergmannf:
This ensure that the parameter will be passed as initially received:

https://stackoverflow.com/questions/4824590/propagate-all-arguments-in-a-bash-shell-script/4824637#4824637


stefsuse:
@MaximilianMeister  please wait for PR manager to merge changes

MaximilianMeister:
>  please wait for PR manager to merge changes

@stefsuse is it doing any extra step that i was missing?

PR #553, opened 2019-08-05
Remove `skuba node reset` command

ereslibre:
## Why is this PR needed?

This command is not necessary to be exposed to the users. It has been
proved that it has produced some misunderstandings about its goal and
under what circumstances it should be used.

`skuba node reset` was added with the assumption that it would be
executed only on nodes that failed to `bootstrap` or `join`, so
whenever this happened, the user could execute `skuba node reset` and
retry the operation from a clean slate.

Since this goal can be done implicitly when calling to `skuba node
bootstrap` or `skuba node join`, by running a implicit `kubeadm reset`
before we attempt the `kubeadm init` or `kubeadm join` operation,
there is no need to have this explicit command exposed to the users.

There is always time to add it back in case we see it's necessary or
if users are requesting to have it.


internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -23,13 +23,11 @@ import (
 
 	"github.com/SUSE/skuba/pkg/skuba"
 	"github.com/pkg/errors"
-	"k8s.io/klog"
```
dannysauer: Is this a separate small cleanup bundled in the same PR, or was it removed for another reason I'm not seeing?

ereslibre: It's the same change, after removing the `crio.reset` here: https://github.com/SUSE/skuba/pull/553/files#diff-eb6411fdf43c95124210c3fbb0eef978L63

pkg/skuba/actions/node/join/join.go
```diff
@@ -62,6 +62,7 @@ func Join(joinConfiguration deployments.JoinConfiguration, target *deployments.T
 	}
 
 	statesToApply := []string{
+		"kubeadm.reset",
```
Klaven: just to document. I would like to see this state run dynamically on failure to clean up. :)

dannysauer: Seconded. We should always clean up after failure when possible.

jhsiaosuse:
Remind to update admin guide https://github.com/SUSE/doc-caasp

PR #554, opened 2019-08-05
Support skuba login (#bsc1140531)

jhsiaosuse:

## Why is this PR needed?

`caasp-cli login` is not available in CaaSP v4

Fixes [#bsc1140531](https://bugzilla.suse.com/show_bug.cgi?id=1140531)

## What does this PR do?

Add `skuba auth login` command porting from v3 `caasp-cli`.

We might support `skuba auth logout` in the future, so change command design from `skuba login` to `skuba auth login`.

### DIFF
--- a/pkg/skuba/actions/auth/login.go
+++ b/pkg/skuba/actions/auth/login.go
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) 2019 SUSE LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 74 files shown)


pkg/skuba/actions/auth/login.go
```diff
@@ -0,0 +1,133 @@
...
+	})
+	if err != nil {
+		return nil, errors.Wrapf(err, "auth failed")
+	}
+
+	url, err := url.Parse(cfg.DexServer)
+	if err != nil {
+		return nil, errors.Wrapf(err, "parse url")
+	}
+	apiserverURL = fmt.Sprintf("https://%s:6443", url.Hostname()) // Guess kube-apiserver on port 6443
```
c3y1huang: Not sure about hardcode the port. Does this work behind lb with port direct?

jhsiaosuse: No. Otherwise, need let the user specify the kube apiserver URL like how gangway does.

innobead: It should be fine for now, because we do the same config for gangway.

```
apiServerURL: "https://{{.ControlPlane}}:6443"
```

ereslibre: I think it would be better in this case that the user provides the host and port in a single argument; if no port is provided we can default to `6443`?

jhsiaosuse: @ereslibre @c3y1huang Agree. I will add new flag `--apiserver `

PR #557, opened 2019-08-06
Add all parser using add_parser.

bergmannf:
Instead of using add_parser for some and new ArgumentParser instances for other
parser, unify the generation of argument parsing by always using add_parser.

## Why is this PR needed?

Code cleanup in the `testparser.py` - this might help fix the invalid `arguments` problem in jenkins nightly builds.


bergmannf:
Not being able to overwrite help when using `add_parser` seems to be the reason for the current construction of parsers.

Closing as this seems to be the only way to get this behaviour.

PR #560, opened 2019-08-06
Handle node upgrade plan output better (bsc#1144914)

MaximilianMeister:
## Why is this PR needed?

When control plane nodes are not yet upgraded, the skuba node upgrade
plan command run against a worker node outputs a misleading:

"Node my-worker-0 is up to date"

In this case we propagate a custom error that will enhance and
clarify the output.

https://bugzilla.suse.com/show_bug.cgi?id=1144914

## What does this PR do?

We propagate a custom error that will enhance and
clarify the output.

Also generalize AllControlPlanesMatchVersion function.
This function can be used to check against any clusterVersion.

Unit tests will be fixed later, therefore setting WIP

## Info for QA

### Related info

* upstream issues

Create a 1.14.1 cluster and run `skuba node upgrade plan` on a worker, output will be:

`Node is up to date`

Create a 1.14.1 cluster and run `skuba node upgrade plan` on a worker, output will be:

`Unable to plan node upgrade: my-worker-0 is not upgradeable until all control plane nodes are upgraded`

## Docs

not necessary, the docs are treating `node upgrade plan` more generically

### DIFF
--- a/internal/pkg/skuba/kubernetes/node_version.go
+++ b/internal/pkg/skuba/kubernetes/node_version.go
@@ -221,24 +221,35 @@ func allWorkerNodesTolerateVersionWithVersioningInfo(allNodesVersioningInfo Node
 	return true
 }
 
-// AllControlPlanesMatchVersion checks that all control planes are on the same version, and that they match the current cluster version
-func AllControlPlanesMatchVersion(currentClusterVersion *version.Version) (bool, error) {
+// AllControlPlanesMatchVersion checks that all control planes are on the same version, and that they match a cluster version
+func AllControlPlanesMatchVersion(clusterVersion *version.Version) (bool, error) {
 	allNodesVersioningInfo, err := AllNodesVersioningInfo()
 	if err != nil {
 		return false, err
 	}
-	return AllControlPlanesMatchVersionWithVersioningInfo(allNodesVersioningInfo, currentClusterVersion), nil
+	return AllControlPlanesMatchVersionWithVersioningInfo(allNodesVersioningInfo, clusterVersion), nil
 }
 
-// AllControlPlanesMatchVersionWithVersioningInfo checks that all control planes are on the same version, and that they match the current cluster version
+// AllControlPlanesMatchVersionWithVersioningInfo checks that all control planes are on the same version, and that they match a cluster version
 // With the addition of passing in a NodeVersionInfoMap to make it easily testable
-func AllControlPlanesMatchVersionWithVersioningInfo(allNodesVersioningInfo NodeVersionInfoMap, currentClusterVersion *version.Version) bool {
+func AllControlPlanesMatchVersionWithVersioningInfo(allNodesVersioningInfo NodeVersionInfoMap, clusterVersion *version.Version) bool {
 	for _, nodeInfo := range allNodesVersioningInfo {
 		// skip workers
 		if !nodeInfo.IsControlPlane() {
 			continue
 		}
-		if !nodeInfo.ToleratesClusterVersion(currentClusterVersion) {
+		if !nodeInfo.ToleratesClusterVersion(clusterVersion) {
+			return false
+		}
+	}
+	return true
+}
+
+// AllNodesMatchVersionWithVersioningInfo returns if all nodes match a specific kubernetes version
+// This can be used to determine e.g. if an upgrade is ongoing
+func AllNodesMatchClusterVersionWithVersioningInfo(allNodesVersioningInfo NodeVersionInfoMap, clusterVersion *version.Version) bool {
+	for _, version := range allNodesVersioningInfo {
+		if !version.EqualsClusterVersion(clusterVersion) {
 			return false
 		}
 	}
--- a/internal/pkg/skuba/upgrade/node/versions.go
+++ b/internal/pkg/skuba/upgrade/node/versions.go
@@ -18,6 +18,7 @@
 package node
 
 import (
+	"fmt"
 	"reflect"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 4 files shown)


MaximilianMeister:
rebased

internal/pkg/skuba/kubernetes/node_version.go
```diff
@@ -221,26 +221,38 @@ func allWorkerNodesTolerateVersionWithVersioningInfo(allNodesVersioningInfo Node
...
+// This can be used to determine e.g. if an upgrade is ongoing
+func AllNodesMatchClusterVersionWithVersioningInfo(allNodesVersioningInfo NodeVersionInfoMap, clusterVersion *version.Version) (bool, error) {
+	match := true
+	for _, version := range allNodesVersioningInfo {
+		if !version.EqualsClusterVersion(clusterVersion) {
+			match = false
+		}
+	}
+	return match, nil
+}
```
ereslibre: This method can be simplified so we don't iterate always over all elements in case that any of them does not match. Also, no need to add error to the signature if we are not generating one inside or bubbling up.

```go
func AllNodesMatchClusterVersionWithVersioningInfo(allNodesVersioningInfo NodeVersionInfoMap, clusterVersion *version.Version) bool {
	for _, version := range allNodesVersioningInfo {
		if !version.EqualsClusterVersion(clusterVersion) {
			return false
		}
	}
	return true
}
```

internal/pkg/skuba/upgrade/cluster/versions.go
```diff
@@ -106,5 +106,13 @@ func UpgradePath() ([]*version.Version, error) {
...
 		return []*version.Version{}, err
 	}
+	upgradePath, err := upgradePathWithCurrentClusterVersion(currentClusterVersion)
+	if err != nil {
+		return []*version.Version{}, err
+	}
+	return upgradePath, nil
+}
+
+func upgradePathWithCurrentClusterVersion(currentClusterVersion *version.Version) ([]*version.Version, error) {
```
ereslibre: I don't understand this change, the logic in `UpgradePathWithAvailableVersions` was already doing this, no?, what is the intent of this changeset?

internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -152,16 +161,32 @@ func controlPlaneUpdateStatusWithAvailableVersions(currentClusterVersion *versio
...
-	controlPlanesMatchVersion := kubernetes.AllControlPlanesMatchVersionWithVersioningInfo(allNodesVersioningInfo, currentClusterVersion)
-	if !controlPlanesMatchVersion {
-		return NodeVersionInfoUpdate{}, errors.New("at least one control plane does not tolerate the current cluster version")
+	controlPlanesMatchVersion := kubernetes.AllControlPlanesMatchVersionWithVersioningInfo(allNodesVersioningInfo, clusterVersion)
+	// Check if there is a newer version
+	versionCompare, err := clusterVersion.Compare(kubernetes.LatestVersion().String())
+	if err != nil {
+		return NodeVersionInfoUpdate{}, err
+	}
+	if versionCompare != 0 && (allNodesMatchCurrentClusterVersion || !controlPlanesMatchVersion) {
```
ereslibre: If we are checking for newer let's avoid `!=` and use `<` `>` comparisons better, just to be sure.

internal/pkg/skuba/upgrade/cluster/versions.go
MaximilianMeister: i wanted to drop this, it was still from experimenting with another way of doing it...

pkg/skuba/actions/node/upgrade/plan.go
```diff
@@ -33,16 +33,17 @@ func Plan(nodeName string) error {
...
-
 	nodeVersionInfoUpdate, err := upgradenode.UpdateStatus(nodeName)
 	if err != nil {
 		return err
 	}
+
+	fmt.Printf("Current Kubernetes cluster version: %s\n", currentClusterVersion.String())
+	fmt.Printf("Latest Kubernetes version: %s\n", kubernetes.LatestVersion().String())
+	fmt.Println()
+	fmt.Printf("Current Node version: %s\n", nodeVersionInfoUpdate.Current)
```
ereslibre:
```suggestion
	fmt.Printf("Current Node version: %s\n", nodeVersionInfoUpdate.Current.KubeletVersion.String())
```

MaximilianMeister: thanks updated, however for some reason the previous way showed exactly the same end result

internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -152,16 +161,29 @@ func controlPlaneUpdateStatusWithAvailableVersions(currentClusterVersion *versio
...
+	// This can be determined by kubernetes.AllNodesMatchClusterVersion(allNodesVersioningInfo, clusterVersion)
+	allNodesMatchCurrentClusterVersion := kubernetes.AllNodesMatchClusterVersionWithVersioningInfo(allNodesVersioningInfo, clusterVersion)
 	// Check that all control plane nodes have at least the current cluster version we plan to
 	// upgrade this worker node to. If not, they need to be fully upgraded first
-	controlPlanesMatchVersion := kubernetes.AllControlPlanesMatchVersionWithVersioningInfo(allNodesVersioningInfo, currentClusterVersion)
-	if !controlPlanesMatchVersion {
-		return NodeVersionInfoUpdate{}, errors.New("at least one control plane does not tolerate the current cluster version")
+	controlPlanesMatchVersion := kubernetes.AllControlPlanesMatchVersionWithVersioningInfo(allNodesVersioningInfo, clusterVersion)
+	// Check if there is a newer version
+	versionCompare, err := clusterVersion.Compare(kubernetes.LatestVersion().String())
```
ereslibre: Can this have side effects in a case like:

* Current cluster version: `1.13.x`
* Latest version: `1.15.2`
* Versions to go through: `1.13.x -> 1.14.y -> 1.15.2`

I think with this check we'll never be able to upgrade worker nodes from `1.13.x -> 1.14.y` because `LatestVersion()` reports `1.15.2`.

ereslibre: Nevermind, it's fine, `versionCompare < 0 && (allNodesMatchCurrentClusterVersion || !controlPlanesMatchVersion)` should do the right thing.

PR #562, opened 2019-08-06
fix typo in dex clusterrole

lcavajani:
Dex can't start currently because the clusterrole is wrong.


innobead:
Thanks!

jhsiaosuse:
Thx

PR #564, opened 2019-08-06
Add testcase for bsc#1142552

drpaneas:
## Why is this PR needed?

Testcase for the 
https://bugzilla.suse.com/show_bug.cgi?id=1142552

### DIFF
--- a/ci/infra/testrunner/tests/test_dockercaps.py
+++ b/ci/infra/testrunner/tests/test_dockercaps.py
@@ -1,0 +1,82 @@
+import logging
+import os
+import pytest
+import pathlib
+import textwrap
+
+logger = logging.getLogger("testrunner")
+
+CONTENT = """---
+apiVersion: v1
+kind: Pod
+metadata:
+    name: leap
+spec:
+    containers:
+    - name: app
+      image: opensuse/leap:latest
+      command: ['/bin/sh', '-c', 'sleep 3600']
+---
+apiVersion: v1
+kind: Pod
+metadata:
+    name: sle12sp4
+spec:
+    containers:
+    - name: app
+      image: registry.suse.com/suse/sles12sp4:latest
+      command: ['/bin/sh', '-c', 'sleep 3600']
+---
+apiVersion: v1
+kind: Pod
+metadata:
+    name: sle15
+spec:
+    containers:
+    - name: app
+      image: registry.suse.com/suse/sle15:latest
+      command: ['/bin/sh', '-c', 'sleep 3600']
+---
+apiVersion: v1
+kind: Pod
+metadata:
+    name: sle15sp1
+spec:
+    containers:
+    - name: app
+      image: registry.suse.de/suse/containers/sle-server/15/containers/suse/sle15:15.1
+      command: ['/bin/sh', '-c', 'sleep 3600']"""
+
+
+@pytest.fixture
+def setup_manifest():
+    cwd = pathlib.Path.cwd()
+    p = cwd / "pods.yaml"
+    p.touch()
+    p.write_text(textwrap.dedent(CONTENT))
+    yield p
+    p.unlink()
+
+
+def test_dockercaps(deployment, kubectl, setup_manifest):
+    logger.info("Deploy testcases")
+    kubectl.run_kubectl(
+        "apply -f {}".format(setup_manifest))
+    kubectl.run_kubectl("wait --for=condition=ready pods --all --timeout=3m")
+
+    logger.info("Test: Run 'su root -c id' on the containers")
+    pods = ["sle12sp4", "leap", "sle15", "sle15sp1"]
+    for container in pods:
+        output = kubectl.run_kubectl(
+            "exec -it {} -- su root -c id".format(container))
+        assert 'uid=0' in output
+
+    logger.info("Test: Add a new user to the containers")
+    for container in pods:
+        output = kubectl.run_kubectl(
+            "exec -it {} -- useradd panos".format(container))
+        assert 'PAM' not in output
+
+    # Remove the testing pods
+    kubectl.run_kubectl(
+        "delete -f {}".format(setup_manifest))


ci/infra/testrunner/tests/test_dockercaps.py
```diff
@@ -0,0 +1,47 @@
...
+    logger.info("Deploy testcases")
+    kubectl.run_kubectl(
+        "apply -f https://pastebin.com/raw/iNA7Kh9L")
+    kubectl.run_kubectl("wait --for=condition=ready pods --all --timeout=3m")
+
+    logger.info("Test: Run 'su root -c id' on the containers")
+    pods = ["sle12sp4", "leap", "sle15", "sle15sp1"]
+    for container in pods:
+        output = kubectl.run_kubectl(
+            "exec -it {} -- {}".format(container, 'su root -c id'))
```
pablochacin: Not sure about using a constant string as a substitution in another constant string. I would put `su root -c id` as part of the `exec ....`

ci/infra/testrunner/tests/test_dockercaps.py
```diff
@@ -0,0 +1,47 @@
...
+        if not 'uid=0' in output:
+            raise NoRootUserError(container, output)
```
pablochacin: In tests `assert` should be used instead of rising exceptions.

ci/infra/testrunner/tests/test_dockercaps.py
```diff
@@ -0,0 +1,47 @@
...
+    def __init__(self, container, output):
+        msg = "Cannot add a new user in the container {}: {}".format(container, output)
+        super().__init__(msg)
+
+
+def test_dockercaps(deployment, kubectl):
+
+    logger.info("Deploy testcases")
+    kubectl.run_kubectl(
+        "apply -f https://pastebin.com/raw/iNA7Kh9L")
```
pablochacin: I will prefer to have these manifests as part of the tests and not as an external resource in pastebin. You could either embed them as a string in this test or load as a file.

ci/infra/testrunner/tests/test_dockercaps.py
```diff
@@ -0,0 +1,47 @@
...
+            "exec -it {} -- {}".format(container, 'su root -c id'))
+        if not 'uid=0' in output:
+            raise NoRootUserError(container, output)
+
+    logger.info("Test: Add a new user to the containers")
+    for container in pods:
+        output = kubectl.run_kubectl(
+            "exec -it {} -- {}".format(container, 'useradd panos'))
+        if 'PAM' in output:
+            raise CannotAddUser(container, output)
```
pablochacin: Same comment regarding using `assert`

ci/infra/testrunner/tests/test_dockercaps.py
drpaneas: so can I do something like this?

```
test_yaml = "blah blah blah"
kubectl.run_kubectl("apply -f", test_yaml)
```

ci/infra/testrunner/tests/test_dockercaps.py
drpaneas: isn't `assert` an exception of it own? Can you elaborate?

FYI: If you run `python -O -c "assert True == False"` the interpreter is _silencing_ those, which is kind of scary to me.

ci/infra/testrunner/tests/test_dockercaps.py
```diff
@@ -0,0 +1,77 @@
...
+import pytest
+import pathlib
+import textwrap
+
+logger = logging.getLogger("testrunner")
+
+
+def test_dockercaps(deployment, kubectl):
+
+    CONTENT = """---
```
bergmannf: If you extract the constant to the module-level, it won't break the visual indentation of the function.

This would help the code to be more readable.

ci/infra/testrunner/tests/test_dockercaps.py
```diff
@@ -0,0 +1,77 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+    name: sle15sp1
+spec:
+    containers:
+    - name: app
+      image: registry.suse.de/suse/containers/sle-server/15/containers/suse/sle15:15.1
+      command: ['/bin/sh', '-c', 'sleep 3600']"""
+    cwd = pathlib.Path.cwd()
```
bergmannf: Creating a new file on the filesystem during testing should always be done in a manner that will ensure it is cleaned up, after the test is done.

Usually this is achieved by using the `setUp`, `tearDown` methods in `xUnit` type of frameworks - in pytest this is not considered the goto method though (https://docs.pytest.org/en/latest/xunit_setup.html) - they advise putting code like this into a `fixture` (https://docs.pytest.org/en/latest/fixture.html#fixture)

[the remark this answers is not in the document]
drpaneas: Great, thx

drpaneas: Thanks for noticing, indeed. Now it works :)

drpaneas:
Thank everyone for their reviews :D

ci/infra/testrunner/tests/test_dockercaps.py
```diff
@@ -0,0 +1,82 @@
...
+    containers:
+    - name: app
+      image: registry.suse.de/suse/containers/sle-server/15/containers/suse/sle15:15.1
+      command: ['/bin/sh', '-c', 'sleep 3600']"""
+
+
+@pytest.fixture
+def setup_manifest():
+    cwd = pathlib.Path.cwd()
+    p = cwd / "pods.yaml"
```
dannysauer: I just want to note that this is simultaneously the ugliest and most creative overloading of the `/` operator I've seen. :D

dannysauer:
In the next revision of these tests, it might be worthwhile to enumerate (and possibly test) the capabilities in addition to running programs which might/probably exercise specific capabilities.  That's pretty easily done using `capsh`.  The primary drawbacks to that are that it'll require a little bit of parsing, and installing a package.  Here's a set of examples of just enumerating what's available:

Non-root on a regular system - note that "current" is empty and the "bounding set" is all which are available to this process:
```
sauer@thunder:~$ sudo zypper -q install -y libcap-progs
sauer@thunder:~$ /usr/sbin/capsh --print
Current: =
Bounding set =cap_chown,cap_dac_override,cap_dac_read_search,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_linux_immutable,cap_net_bind_service,cap_net_broadcast,cap_net_admin,cap_net_raw,cap_ipc_lock,cap_ipc_owner,cap_sys_module,cap_sys_rawio,cap_sys_chroot,cap_sys_ptrace,cap_sys_pacct,cap_sys_admin,cap_sys_boot,cap_sys_nice,cap_sys_resource,cap_sys_time,cap_sys_tty_config,cap_mknod,cap_lease,cap_audit_write,cap_audit_control,cap_setfcap,cap_mac_override,cap_mac_admin,cap_syslog,cap_wake_alarm,cap_block_suspend,cap_audit_read
Securebits: 00/0x0/1'b0
 secure-noroot: no (unlocked)
 secure-no-suid-fixup: no (unlocked)
 secure-keep-caps: no (unlocked)
uid=1001(sauer)
gid=100(users)
groups=100(users),468(libvirt),470(docker)
```

Root on a clean, default Docker image.  Again, notice that the bounding set is a much smaller set than the full list of supported capabilities, and that the user (root) has "current" all of the available capabilities.
```
sauer@thunder:~$ docker run --rm -it opensuse/leap:15.1
8d58c376c2a3:/ # zypper -q install -y libcap-progs
8d58c376c2a3:/ # /usr/sbin/capsh --print
Current: = cap_chown,cap_dac_override,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_net_bind_service,cap_net_raw,cap_sys_chroot,cap_mknod,cap_audit_write,cap_setfcap+eip
Bounding set =cap_chown,cap_dac_override,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_net_bind_service,cap_net_raw,cap_sys_chroot,cap_mknod,cap_audit_write,cap_setfcap
Securebits: 00/0x0/1'b0
 secure-noroot: no (unlocked)
 secure-no-suid-fixup: no (unlocked)
 secure-keep-caps: no (unlocked)
uid=0(root)
gid=0(root)
groups=
```

And, the way _I'd_ like to see things look by default... ;)
```
sauer@thunder:~$ docker run --rm -it --cap-drop all opensuse/leap:15.1
ae23ed288194:/ # zypper -q install -y libcap-progs
ae23ed288194:/ # /usr/sbin/capsh --print
Current: =
Bounding set =
Securebits: 00/0x0/1'b0
 secure-noroot: no (unlocked)
 secure-no-suid-fixup: no (unlocked)
 secure-keep-caps: no (unlocked)
uid=0(root)
gid=0(root)
groups=
```

PR #567, opened 2019-08-07
Add merge restrictions for post RC

jordimassaguerpla:
This reflects the configuration and we are writing in here to make it
explicit for the contributor.

## Why is this PR needed?

So contributors know the rules when they create a PR.

## What does this PR do?

Update the github template

This is reflecting the rules. Please do not discuss the rules in this PR.

## Info for QA

This is not a bug fix

## Docs

Already updated at https://confluence.suse.com/pages/viewpage.action?spaceKey=SUSECaaSPlatform4&title=Development+Workflow+for+CaaSP+v4+%28vnext%29+development

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -48,5 +48,24 @@ If docs need to be updated, please add a link to a PR to https://github.com/SUSE
 At the time of creating the issue, this PR can be work in progress (set its title to [WIP]),
 but the documentation needs to be finalized before the PR can be merged. 
 
+
+# Merge restrictions after RC
+
+(Please do not edit this)
+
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
> Looks good to me.
> QA is also low on people at the moment, to I expect some delays here. Pinging and assigning the squads-QA person will help.
> 
> Is there any expectation regarding the QA approval? For example, a full integration test is passing?

No explicit expectations. Whatever you think is necessary. From my experience, some times the bug is so trivial ( one line fix ) that you don't need to to anything. Other times, you will need the "Before and After" info to know how was this tested and so you know how to validate this. Other times, you would like to test it, for example a bug that requires a complex scenario or a bug that has a high chance to cause a regression... whatever you guys think is needed so when we start a release we have high chances to make it in the planned time.

PR #569, opened 2019-08-07
Init extra arguments for configuration

mjura:
## Why is this PR needed?

We would like to allow users provide their own custom extra arguments for `apiserver`, `controller-manager` and `kubelet` services.

## What does this PR do?

It is initializing empty configuration field for extra arguments.

## How to use it?

Before cluster bootstraping in `kubeadm-init.conf` additional flags for `kubelet, `apiserver`, `controller-manager` and `scheduler` can be provided:
#### kubelet
```
nodeRegistration:
  kubeletExtraArgs:
```
#### kube-apiserver
```
apiServer:
  extraArgs:
```
#### kube-controller-manager
```
controllerManager:
  extraArgs:
```
#### kube-scheduler
```
scheduler:
  extraArgs:

```

### DIFF
--- a/pkg/skuba/actions/cluster/init/manifests.go
+++ b/pkg/skuba/actions/cluster/init/manifests.go
@@ -23,6 +23,8 @@ kind: InitConfiguration
 bootstrapTokens: []
 localAPIEndpoint:
   advertiseAddress: ""
+nodeRegistration:
+  kubeletExtraArgs: {}
 ---
 apiVersion: kubeadm.k8s.io/v1beta1
 kind: ClusterConfiguration
@@ -37,6 +39,10 @@ apiServer:
     oidc-groups-claim: groups
 clusterName: {{.ClusterName}}
 controlPlaneEndpoint: {{.ControlPlane}}:6443
+controllerManager:
+  extraArgs: {}
+scheduler:
+  extraArgs: {}
 dns:
   imageRepository: {{.ImageRepository}}
   imageTag: {{.CoreDNSImageTag}}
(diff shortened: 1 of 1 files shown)


mjura:
> Generally LGTM, but for completeness - the scheduler also allows ExtraArgs that could be included:

@bergmannf thank you, you are right

bergmannf:
Ah I must have missed it the first time around it seems `etcd` also allows extraargs:

https://godoc.org/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm#LocalEtcd

mjura:
@bergmannf ok, I have added also `initConfiguration.Etcd.Local.ExtraArgs`

mjura:
@ereslibre I have added empty `extraArgs` map to `kubeadm-init.conf` and to master, worker templates as well

pkg/skuba/actions/node/bootstrap/bootstrap.go
```diff
@@ -68,6 +68,8 @@ func Bootstrap(bootstrapConfiguration deployments.BootstrapConfiguration, target
 		return errors.Wrap(err, "unable to add target information to init configuration")
 	}
 
+	setConfigurationExtraArgs(initConfiguration)
```
ereslibre: I think all the changes in `bootstrap.go` can be removed in this PR. Changing the `init` template will allow the user to easily change the `extraArgs` between `cluster init` and `node bootstrap`.

Setting defaults on `bootstrap` to empty maps has no effect.

mjura: Ok, you are right, I have checked it and it works as you expected

mjura:
According to @ereslibre comment let's close it

PR #570, opened 2019-08-07
Cleanup kubectl wrapper

chentex:
Signed-off-by: Vicente Zepeda Mas <vzepedamas@suse.com>

## Why is this PR needed?

Cleans up `kubectl` wrapper with methods that are no longer needed.

## What does this PR do?

Cleans up `kubectl` wrapper with methods that are no longer needed.

## Info for QA

Deploy test should continue to work

### DIFF
--- a/ci/infra/testrunner/kubectl/kubectl.py
+++ b/ci/infra/testrunner/kubectl/kubectl.py
@@ -1,12 +1,11 @@
 import os
 import platforms
 
-from timeout_decorator import timeout
-
 from platforms.platform import Platform
 from skuba.skuba import Skuba
 from utils.format import Format
 from utils.utils import (step, Utils)
+
 
 class Kubectl:
 
@@ -17,45 +16,6 @@ class Kubectl:
         self.skuba = Skuba(conf, platform)
 
 
-    def create_deployment(self, name, image):
-        self._run_kubectl("create deployment {name} --image={image}"
-                          .format(name=name, image=image))
-
-    def scale_deployment(self, name, replicas):
-        self._run_kubectl("scale deployment {name} --replicas={replicas}"
-                          .format(name=name, replicas=replicas))
-
-    def expose_deployment(self, name, port, nodeType="NodePort"):
-        self._run_kubectl("expose deployment {name} --port={port} --type={nodeType}"
-                          .format(name=name, port=port, nodeType=nodeType))
-
-    def wait_deployment(self, name, timeout):
-        self._run_kubectl("wait --for=condition=available deploy/{name} --timeout={timeout}m"
-                          .format(name=name, timeout=timeout))
-
-    def count_available_replicas(self, name):
-        result = self._run_kubectl("get deployment/{name} | jq '.status.availableReplicas'"
-                                   .format(name=name))
-        return int(result)
-
-    def get_service_port(self, name):
-        result = self._run_kubectl("get service/{name} | jq '.spec.ports[0].nodePort'"
-                                   .format(name=name))
-        return int(result)
-
-    def test_service(self, name, path="/", worker=0):
-        ip_addresses = self.platform.get_nodes_ipaddrs("worker")
-        if worker >= len(ip_addresses):
-            raise ValueError("Node worker-{} not deployed".format(worker))
-
-        port = self.get_service_port(name)
-        shell_cmd = "curl {ip}:{port}{path}".format(ip=ip_addresses[worker],port=port,path=path)
-        try:
-            return self.utils.runshellcommand(shell_cmd)
-        except Exception as ex:
-            raise Exception("Error testing service {} with path {} at node {}"
-                                .format(name, path, ip_address)) from ex
- 
     def run_kubectl(self, command):
         kubeconfig = self.skuba.get_kubeconfig()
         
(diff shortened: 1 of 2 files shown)


ci/infra/testrunner/tests/test_deployments.py
```diff
@@ -1,21 +1,53 @@
...
-    result = kubectl.test_service("nginx")
-    assert "Welcome to nginx" in result
+    kubectl.run_kubectl("create deployment nginx --image=nginx:stable-alpine")
+    kubectl.run_kubectl("scale deployment nginx --replicas={replicas}".format(replicas=workers))
+    kubectl.run_kubectl("expose deployment nginx --port=80 --type=NodePort")
+    kubectl.run_kubectl("wait --for=condition=available deploy/nginx --timeout={timeout}m".format(timeout=3))
+    deployment = kubectl.run_kubectl("get deployment/nginx -o json")
+    availableReplicas = readyReplicas = updatedReplicas = replicas = 0
+    try:
+        obj = json.loads(deployment)
```
pablochacin: Why you just use the jsonpath option for kubectl? 
```
kubectl get deployment/nginx -o jsonpath='{ .status.readyReplicas }'
```

ci/infra/testrunner/tests/test_deployments.py
```diff
@@ -1,21 +1,53 @@
...
+        replicas = int(obj["status"]["replicas"])
+    except ex as Exception:
+        raise ex
+
+    assert replicas == workers
+    assert availableReplicas == workers
+    assert readyReplicas == workers
+    assert updatedReplicas == workers
+
+    service = kubectl.run_kubectl("get service/nginx -o json")
```
pablochacin: same comment than above:
```
kubectl get service nginx -o jsonpath='{ .spec.ports[0].nodePort }' 
```

ci/infra/testrunner/tests/test_deployments.py
```diff
@@ -1,21 +1,53 @@
...
+import requests
+
 from skuba import Skuba
 from kubectl import Kubectl
-import pytest
-import time
-from timeout_decorator import timeout
+
 
 def test_nginx_deployment(setup, skuba, kubectl):
```
pablochacin: If you use the  `deployment` fixture there's no need to add workers manually

ci/infra/testrunner/tests/test_deployments.py
```diff
@@ -1,21 +1,53 @@
...
+    try:
+        obj = json.loads(service)
+        nodePort = int(obj["spec"]["ports"][0]["nodePort"])
+    except ex as Exception:
+        raise ex
+    
+    wrk_idx = 0
+    ip_addresses = skuba.platform.get_nodes_ipaddrs("worker")
+    if wrk_idx >= len(ip_addresses):
+        raise ValueError("Node worker-{} not deployed".format(wrk_idx))
```
pablochacin: Why using `rise` instead of an `assert? In any case,  this check should be done in the `deployment` fixture.

ci/infra/testrunner/kubectl/kubectl.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
pablochacin: I would say this is not longer needed.

ci/infra/testrunner/tests/test_deployments.py
```diff
@@ -1,21 +1,26 @@
...
-import time
-from timeout_decorator import timeout
+import json
+import requests
+
+from kubectl import Kubectl
+
+
+def test_nginx_deployment(deployment, kubectl):
+    workers = 3
```
pablochacin: Why hard code this value? you can retrieve it from skuba (I think you were doing this in a previous version)
```
def test_nginx_deployment(deployment, skuba, kubectl): 
      workers = skuba.num_of_nodes("worker")
```

ci/infra/testrunner/tests/test_deployments.py
```diff
@@ -1,21 +1,26 @@
...
-    assert "Welcome to nginx" in result
+import json
+import requests
+
+from kubectl import Kubectl
+
+
+def test_nginx_deployment(deployment, kubectl):
+    workers = kubectl.skuba.num_of_nodes("worker")
+    kubectl.run_kubectl("create deployment nginx --image=nginx:stable-alpine")
```
drpaneas: Maybe add a test making sure the nginx deployment is successfully deployed before you try to scale it up. For example as as described [here](https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/) and then `ssh sles@nodeip curl podIP:podPort` to verify nginx is working internally and it's reachable within the cluster. This is not necessary though, however if your external service test fails, you would wish to have the internal one (as debugging step).

ci/infra/testrunner/tests/test_deployments.py
```diff
@@ -1,21 +1,26 @@
...
+    kubectl.run_kubectl("scale deployment nginx --replicas={replicas}".format(replicas=workers))
+    kubectl.run_kubectl("expose deployment nginx --port=80 --type=NodePort")
```
drpaneas: I would test an extra test verifying the output of `kubectl get svc`

ci/infra/testrunner/tests/test_deployments.py
chentex: Maybe to add a more checks, but I already check that the service is running, in the last assert from line 23 - 26

drpaneas: yes, this can be done later as well, that's why I gave you already my blessing to merge this :) :P

PR #571, opened 2019-08-07
Add verbosity to CI skuba execution

dannysauer:
add verbosity argument to run_skuba

## Why is this PR needed?

This simple PR adds a verbosity parameter to the skuba execution method in testrunner with an elevated default value.

Fixes SUSE/avant-garde#547

## What does this PR do?

Increased default verbosity on skuba executions will increase the output recorded in the test cases, facilitating easier debugging when tests fail.  Passing it as an optional parameter allows specific calls to fine-tune the value as needed.

### DIFF
--- a/ci/infra/testrunner/skuba/skuba.py
+++ b/ci/infra/testrunner/skuba/skuba.py
@@ -160,7 +160,7 @@ class Skuba:
         path = "{cwd}/test-cluster/admin.conf".format(cwd=self.conf.workspace)
         return path
 
-    def _run_skuba(self, cmd, cwd=None):
+    def _run_skuba(self, cmd, cwd=None, verbosity=None):
         """Running skuba command in cwd.
         The cwd defautls to {workspace}/test-cluster but can be overrided
         for example, for the init command that must run in {workspace}
@@ -170,6 +170,19 @@ class Skuba:
         if cwd is None:
             cwd = self.cwd
 
+        if verbosity is None:
+            verbosity = self.conf.skuba.verbosity
+        try:
+            v = int(verbosity)
+        except ValueError:
+            raise ValueError(f"verbosity '{verbosity}' is not an int")
+        verbosity = v
+
         env = {"SSH_AUTH_SOCK": self.utils.ssh_sock_fn()}
 
-        return self.utils.runshellcommand(self.binpath + " " + cmd, cwd=cwd, env=env)
+
+        return self.utils.runshellcommand(
+          f"{self.binpath} -v {verbosity} {cmd}",
+          cwd=cwd,
+          env=env,
+          )
--- a/ci/infra/testrunner/utils/constants.py
+++ b/ci/infra/testrunner/utils/constants.py
@@ -83,6 +83,7 @@ class BaseConfig:
             super().__init__()
             self.binpath = None
             self.srcpath = None
+            self.verbosity = 5
 
     class Test:
         def __init__(self):
--- a/ci/infra/testrunner/vars.yaml
+++ b/ci/infra/testrunner/vars.yaml
@@ -8,6 +8,7 @@ nodeuser: "sles"  # Default node username
 skuba:
   binpath: "" # path to skuba binary
   srcpath: "" # path to skuba source
+  verbosity: 5 # default value passed to skuba with -v
 
 # platform settings
 terraform:


ci/infra/testrunner/skuba/skuba.py
```diff
@@ -187,4 +187,4 @@ def _run_skuba(self, cmd, cwd=None):
 
         env = {"SSH_AUTH_SOCK": os.path.join(self.conf.workspace, "ssh-agent-sock")}
 
-        return self.utils.runshellcommand(self.binpath + " "+ cmd, cwd=cwd, env=env)
+        return self.utils.runshellcommand(self.binpath + f" -v {verbosity} " + cmd, cwd=cwd, env=env)
```
tdaines42: maybe make this whole thing an f string? f"{self.binpath} -v {verbosity} {cmd}"

dannysauer: Doh.  I was so excited about changing as little as possible... :)

dannysauer: Updated

dannysauer:
Moved the default verbosity setting out to the config file, and added validation that the value provided is actually an integer.  Good suggestion, @tdaines42.  I also poked at the section I added it a little to make it flake8-friendly. :)

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -185,6 +185,15 @@ def _run_skuba(self, cmd, cwd=None):
         if cwd is None:
            cwd=self.cwd
 
+        if verbosity is None:
+           verbosity = self.conf.skuba.verbosity  # noqa: E111
+        try:
+            v = int(verbosity)
+        except ValueError:
+            v = 2  # default verbosity if unspecified / error
+            logger.warn(f"verbosity '{verbosity}' is not an int; using {v}")
```
pablochacin: I would not swallow the exception silently. I would rise using the message you are using for the log.

dannysauer: I guess the question is if this configuration error should halt execution.  Initially, I thought "no", but maybe you're right; maybe it should die if an invalid value is provided.

dannysauer: I switched it to log as an error and raise a generic exception.  Should it just `sys.exit` instead, as there are a couple of places where _run_skuba is called inside a try block that (mildly annoyingly) catches all exceptions?  I feel like raising `Exception` is probably the technically better route, but am open to suggestions.

dannysauer:
> You should also add the verbosity option to the [Skuba configuration class ](https://github.com/SUSE/skuba/blob/master/ci/infra/testrunner/utils/constants.py#L81)and set the default value there

Ah, I didn't know about that one.  Done.

ci/infra/testrunner/skuba/skuba.py
pablochacin: I generally against exiting from inside a library. Exception should be a prefered way. And logging and raising an exception is generally considered an anti-pattern (we have been removing this from testrunner)

dannysauer: I agree regarding log + exception (and regarding exit). However, several calls to this `catch exception` without emitting the exception text, so I did this out of concern about the cause of the error being lost.  Little is more annoying as a user than an error which doesn't say what's wrong. :)

Would it be better to leave this in and remove it in a separate PR improving the relevant exception handling, or just go ahead and improve the `catch` blocks' text in this PR?

ci/infra/testrunner/utils/constants.py
```diff
@@ -83,6 +83,7 @@ def __init__(self):
             super().__init__()
             self.binpath = None
             self.srcpath = None
+            self.verbosity = 2
```
MaximilianMeister: shouldnt this read from the skuba conf?

chentex: I think this is here in case there is an outdated `var.yaml` file that doesn't have a value.

dannysauer: Yeah; this is a default that gets overridden by the config value if present.

dannysauer:
Rebased on current master & squashed changes. :)

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -170,6 +170,21 @@ def _run_skuba(self, cmd, cwd=None):
         if cwd is None:
             cwd = self.cwd
 
+        if verbosity is None:
+           verbosity = self.conf.skuba.verbosity  # noqa: E111
```
stefsuse: indent is off, need one more space

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -170,6 +170,21 @@ def _run_skuba(self, cmd, cwd=None):
...
+        try:
+            v = int(verbosity)
+        except ValueError:
+            msg = f"verbosity '{verbosity}' is not an int"
+            logger.error(msg)
+            raise Exception(msg)
```
stefsuse: I think it makes a lot more sense to just default to 2 here like you had it in the first place, but I was late to the party..

ci/infra/testrunner/skuba/skuba.py
dannysauer: I thought it looked better with the variables vertically aligned, feels messed up with that weird one-char offset. Python needs to support `iff` :)

ci/infra/testrunner/skuba/skuba.py
dannysauer: Whoops; I forgot to change that to `logger.exception()` and just inline the `msg`.

ci/infra/testrunner/skuba/skuba.py
dannysauer: Switched to raise customized ValueError exception.

ci/infra/testrunner/skuba/skuba.py
dannysauer: Extra space added so it's wonky like the preceding block. ;)

ci/infra/testrunner/skuba/skuba.py
dannysauer: I switched it to raise a more descriptive ValueError exception.  The vars file contains a default value of "2", and the class has a default of 2.  If the config file is old and the var isn't present, it'll use "2".  If the config file is new and the value is something other than 2, then the user changed it intentionally _away from_ 2 and we should notify them that they chose an invalid value.  It might make sense to check for an empty string and use that to suppress adding the `-v`; defaulting to the program's default verbosity level - but that seems an unlikely use case.

dannysauer:
@maximilianmaher 
> however i would set the default to 5

I only messed with it a little, and sort of guessed at 2.  Is 5 "too" verbose?  I guess it's configurable if someone wants less, which is sort of the whole point of this. :D  Done.

MaximilianMeister:
@dannysauer thanks for adapting the default, however now seems that the latest commit failed the author check, can you maybe squash the commits?

dannysauer:
Whoops - I committed with wrong email (used my personal laptop to make that change). :D  Squashed and fixed author.

PR #578, opened 2019-08-08
Remove a hardcoded image reference for pause image

davidcassany:
## Why is this PR needed?

We cannot ship with references to `registry.suse.de` devel or staging projects.

## What does this PR do?

Make use of ImageRepository constant that is dependent on the build environment.

This should solve pause image references in SUSE namespace

## Info for QA

Needs to validate that pause image is coming from registry registry.suse.com and that /etc/sysconfig/crio includes the proper reference

### DIFF
--- a/pkg/skuba/actions/cluster/init/manifests.go
+++ b/pkg/skuba/actions/cluster/init/manifests.go
@@ -58,7 +58,7 @@ useHyperKubeImage: true
 ## Default        : ""
 ## ServiceRestart : crio
 #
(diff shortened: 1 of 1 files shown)


davidcassany:
@lcavajani looks like my rocket chat is frozen... see this PR, I think this should solve the registry.suse.de issue.

lcavajani:
@davidcassany this would work but could not we use https://github.com/SUSE/skuba/blob/master/internal/pkg/skuba/kubernetes/versions.go#L82 in order to build the image like we do for Cilium and co so we keep consistent ?

davidcassany:
@lcavajani  I don't see what is different in that case... in versions.go I can't see the image references, just image tags.

lcavajani:
We'd need to add `PauseImage` etc in https://github.com/SUSE/skuba/blob/master/pkg/skuba/actions/cluster/init/init.go#L34

@ereslibre can say more

davidcassany:
Superseded by #579

PR #585, opened 2019-08-09
Install the pattern instead of explicit packages (bsc#1145009)

MaximilianMeister:
## Why is this PR needed?

When upgrading from 1.14.1 to 1.15.2 we found out that cri-o has
not the right patch level version yet.

Kubernetes and cri-o are not tied to patch levels so we install
the pattern instead.

(bsc#1145009)

### DIFF
--- a/pkg/skuba/actions/node/upgrade/apply.go
+++ b/pkg/skuba/actions/node/upgrade/apply.go
@@ -120,8 +120,8 @@ func Apply(target *deployments.Target) error {
 			}
 			if nodeVersionInfoUpdate.HasMajorOrMinorUpdate() {
 				err = target.Apply(deployments.KubernetesBaseOSConfiguration{
-					KubeadmVersion:    nodeVersionInfoUpdate.Update.APIServerVersion.String(),
-					KubernetesVersion: nodeVersionInfoUpdate.Current.APIServerVersion.String(),
+					UpdatedVersion: nodeVersionInfoUpdate.Update.APIServerVersion.String(),
+					CurrentVersion: nodeVersionInfoUpdate.Current.APIServerVersion.String(),
 				}, "kubernetes.install-intermediate-node-pattern")
 				if err != nil {
 					return err
@@ -134,7 +134,7 @@ func Apply(target *deployments.Target) error {
 				return err
 			}
 			err = target.Apply(deployments.KubernetesBaseOSConfiguration{
-				KubernetesVersion: nodeVersionInfoUpdate.Update.APIServerVersion.String(),
+				CurrentVersion: nodeVersionInfoUpdate.Update.APIServerVersion.String(),
 			}, "kubernetes.install-node-pattern")
 			if err != nil {
 				return err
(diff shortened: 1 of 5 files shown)


pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -132,9 +132,8 @@ func Apply(target *deployments.Target) error {
 				return err
 			}
 			err = target.Apply(deployments.KubernetesBaseOSConfiguration{
-				KubeadmVersion:    nodeVersionInfoUpdate.Update.APIServerVersion.String(),
```
ereslibre: We might want to revisit this struct later

MaximilianMeister:
> Is this a ship stopper fro RC1?

if we don't update cri-o to 1.15.2 it's a ship stopper

MaximilianMeister:
> I believe we should also fix the kubernetesInstallNodePattern, it installs kubernetes-kubeadm which is not needed at all, we should relay only on patterns.

if i remove this explicit installation of `kubeadm-old`, e.g. we're installing kubeadm-1.15.0 within the `patterns-caasp-Node-1.14` which is wrong

MaximilianMeister:
> if i remove this explicit installation of `kubeadm-old`, e.g. we're installing kubeadm-1.15.0 within the `patterns-caasp-Node-1.14` which is wrong

turned out this was an issue in `patterns-caasp-Node-1.14` it's addressed now in https://build.suse.de/request/show/198507 so we could go ahead and remove the explicit kubeadm install within skuba if needed

MaximilianMeister:
just rebased to retrigger all CI jobs, one was stuck and not retryable

pkg/skuba/actions/node/upgrade/apply.go
ereslibre: Since we are not merging this immediately, do you think you could rename `KubeadmVersion` and `KubernetesVersion` to something more meaningful now that we are not forcing explicit package installs? something like `CurrentVersion` and `NewVersion` or something like that?

MaximilianMeister: updated in a separate commit

MaximilianMeister:
the current functionality is broken see https://github.com/SUSE/skuba/blob/master/internal/pkg/skuba/deployments/ssh/kubernetes.go#L73

it should be `kubernetesBaseOSConfiguration.ContainerRuntimeVersion`

one more reason to merge this PR, currently updates fail with:

`I0815 11:42:56.670009   22388 ssh.go:190] stderr | Package 'cri-o=1.15.2' not found.`

MaximilianMeister:
rebased

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -120,16 +120,16 @@ func Apply(target *deployments.Target) error {
...
 				if err != nil {
 					return err
 				}
 			} else {
 				err = target.Apply(deployments.KubernetesBaseOSConfiguration{
-					KubeadmVersion:    nodeVersionInfoUpdate.Update.APIServerVersion.String(),
-					KubernetesVersion: nodeVersionInfoUpdate.Current.APIServerVersion.String(),
+					UpdatedVersion: nodeVersionInfoUpdate.Update.APIServerVersion.String(),
+					CurrentVersion: nodeVersionInfoUpdate.Current.APIServerVersion.String(),
 				}, "kubernetes.install-base-packages")
```
ereslibre:
```suggestion
				}, "kubernetes.install-intermediate-node-pattern")
```

MaximilianMeister: oh thanks, i missed that. i think we can then drop the whole `if nodeVersionInfoUpdate.HasMajorOrMinorUpdate()` and always install the intermediate pattern

ereslibre:
@MaximilianMeister another suggestion, do you think we can keep everything in one single function? If `UpdatedVersion` is provided, we install `patterns-caasp-Node-%s-%s`, if no `UpdatedVersion` is provided, we install `patterns-caasp-Node-%s`, WDYT?

MaximilianMeister:
> do you think we can keep everything in one single function? If `UpdatedVersion` is provided, we install `patterns-caasp-Node-%s-%s`, if no `UpdatedVersion` is provided, we install `patterns-caasp-Node-%s`, so we can drop `kubernetes.install-intermediate-node-pattern`, WDYT?

made sense to me. thanks for the hint. updated

PR #588, opened 2019-08-09
Explicitly pass kubernetes.Interface to functions that need it

bergmannf:
## Why is this PR needed?

Supersedes #429 (it seems after the skuba rename, fork <-> origin connection is lost).

Instead of having every function retrieving the `GetAdminClientSet` explicitly this will dependency inject this into every function that needs it.

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/cni.go
+++ b/internal/pkg/skuba/deployments/ssh/cni.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/SUSE/skuba/internal/pkg/skuba/cni"
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
 	"github.com/SUSE/skuba/pkg/skuba"
 )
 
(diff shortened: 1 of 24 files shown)


internal/pkg/skuba/deployments/ssh/cni.go
```diff
@@ -33,10 +34,14 @@ func init() {
 }
 
 func cniDeploy(t *Target, data interface{}) error {
```
ereslibre: With this refactor I think we can also consider to pass the interface to all states as an argument along with the target and data.

ereslibre: Or maybe as a function/attribute on `Target`; maybe it doesn't make sense (depends on how often we use the clientset on the states themselves)

bergmannf:
I 100% agree on not merging, I just reopened it after realizing my rebase didn't end up in the old PR.

Before I forget about it again, I wanted to make sure I have a new one open.

jenting:
Use context and put the clientset into context is a good way.

BTW, it would be nice to align the import ordering.
// standard package

// 3rd-party package

// skuba package

inercia:
> BTW, it would be nice to align the import ordering.
> // standard package
> 
> // 3rd-party package
> 
> // skuba package

I completely agree, I think we should standarize the order of imports in our code.

jenting:
> > BTW, it would be nice to align the import ordering.
> > // standard package
> > // 3rd-party package
> > // skuba package
> 
> I completely agree, I think we should standarize the order of imports in our code.

I have VScode setting for auto reordering.

I think I could add CI check on this, post-GA :)

ereslibre:
This PR has many conflicts, should it be revisited again, or closed?

suseclee:
I am working on it with https://github.com/SUSE/avant-garde/issues/989. 
I will close it after I correct all info from this PR.

jenting:
Close this one, another [PR](https://github.com/SUSE/skuba/pull/806) address this got merged.

PR #589, opened 2019-08-09
Change order of arguments passed to testrunner

ereslibre:
## Why is this PR needed?

Also, explicitly specify `nargs` for `platform` argument in
testrunner.

This ensures that at most 1 argument will be consumed (if present).

### DIFF
--- a/ci/infra/testrunner/testrunner.py
+++ b/ci/infra/testrunner/testrunner.py
@@ -92,6 +92,7 @@ def main():
     parser.add_argument("-p", "--platform",
                         default="openstack",
                         choices=["openstack", "vmware", "bare-metal", "libvirt"],
+                        nargs="?",
                         help="The platform you're targeting. Defaults to openstack")
     parser.add_argument("-l", "--log-level", dest="log_level", default=None, help ="log level",
                         choices=["DEBUG", "INFO", "WARNING", "ERROR"])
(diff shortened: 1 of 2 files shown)


ci/Makefile
```diff
@@ -67,20 +67,20 @@ post_run: gather_logs cleanup
...
 
 .PHONY: test_upgrade_apply_from_previous
 test_upgrade_apply_from_previous:
-	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} test --verbose --suite test_skuba_upgrade.py --test test_upgrade_apply_from_previous
+	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} --verbose --suite test_skuba_upgrade.py --test test_upgrade_apply_from_previous test
 
 .PHONY: test_upgrade_apply_user_lock
 test_upgrade_apply_all_fine:
-	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} test --verbose --suite test_skuba_upgrade.py --test test_upgrade_apply_user_lock
+	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} --verbose --suite test_skuba_upgrade.py --test test_upgrade_apply_user_lock test
```
chentex: I think the `test` subcommand should be before the `--verbose` flag otherwise it breaks I think

PR #590, opened 2019-08-09
Upgrade: do not check container runtime version

ereslibre:
## Why is this PR needed?

Fixes: https://github.com/SUSE/avant-garde/issues/729

When upgrading do not check the full container runtime version.

cri-o releases are not up to speed with Kubernetes releases, and while
we have already a new Kubernetes version to upgrade to, cri-o may be
lagging behind yet.

Since the patchlevel version drift is tolerable, do not take
explicitly into account the exact version when determining if a node
is up to date when it comes to checking the container runtime version.

### DIFF
--- a/internal/pkg/skuba/kubernetes/versions.go
+++ b/internal/pkg/skuba/kubernetes/versions.go
@@ -83,7 +83,7 @@ var (
 			},
 			ComponentsVersion: ComponentsVersion{
 				KubeletVersion:          "1.15.2",
-				ContainerRuntimeVersion: "1.15.2",
+				ContainerRuntimeVersion: "1.15.0",
 			},
 			AddonsVersion: AddonsVersion{
 				CiliumVersion:  "1.5.3",
--- a/internal/pkg/skuba/upgrade/node/versions.go
+++ b/internal/pkg/skuba/upgrade/node/versions.go
@@ -36,7 +36,13 @@ type NodeVersionInfoUpdate struct {
 }
 
 func (nviu NodeVersionInfoUpdate) IsUpdated() bool {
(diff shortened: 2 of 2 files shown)


ereslibre:
Had to force push, too long commit title :)

internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -83,7 +83,7 @@ var (
 			},
 			ComponentsVersion: ComponentsVersion{
 				KubeletVersion:          "1.15.2",
-				ContainerRuntimeVersion: "1.15.2",
+				ContainerRuntimeVersion: "1.15.0",
```
lcavajani: would it be sufficient to just have this version set to `1.15.0` ?

ereslibre: @lcavajani not sure I'm following, it's set to `1.15.0`, am I missing something?

lcavajani: sorry I mean, would it have worked if we had here `1.15.0` from the beginning ?

ereslibre: Yes, it would, but we'd have risked this happening in the future...

ereslibre:
> What is the likelihood that the same problem happens with etcd in the future?

Low, I'd say, as its versions are not strictly tied to Kubernetes' as cri-o ones.

internal/pkg/skuba/kubernetes/versions.go
lcavajani: Thanks, I just wanted to be sure I understand.

PR #591, opened 2019-08-09
Upgrade: install intermediate pattern when needed

ereslibre:
## Why is this PR needed?

Do not install the intermediate pattern unless the upgrade of
Kubernetes is a major or minor one.

### DIFF
--- a/internal/pkg/skuba/upgrade/node/versions.go
+++ b/internal/pkg/skuba/upgrade/node/versions.go
@@ -35,6 +35,17 @@ type NodeVersionInfoUpdate struct {
 	Update  kubernetes.NodeVersionInfo
 }
 
+func (nviu NodeVersionInfoUpdate) HasMajorOrMinorUpdate() bool {
+	if nviu.Current.IsControlPlane() {
+		if nviu.Update.APIServerVersion.Major() > nviu.Current.APIServerVersion.Major() ||
+			nviu.Update.APIServerVersion.Minor() > nviu.Current.APIServerVersion.Minor() {
+			return true
+		}
+	}
+	return nviu.Update.KubeletVersion.Major() > nviu.Current.KubeletVersion.Major() ||
+		nviu.Update.KubeletVersion.Minor() > nviu.Current.KubeletVersion.Minor()
+}
+
 func (nviu NodeVersionInfoUpdate) IsUpdated() bool {
 	return reflect.DeepEqual(nviu.Current, nviu.Update)
 }
--- a/pkg/skuba/actions/node/upgrade/apply.go
+++ b/pkg/skuba/actions/node/upgrade/apply.go
@@ -118,12 +118,22 @@ func Apply(target *deployments.Target) error {
 			if err != nil {
 				return err
 			}
-			err = target.Apply(deployments.KubernetesBaseOSConfiguration{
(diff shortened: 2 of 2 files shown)


pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -118,12 +118,14 @@ func Apply(target *deployments.Target) error {
 			if err != nil {
 				return err
 			}
-			err = target.Apply(deployments.KubernetesBaseOSConfiguration{
-				KubeadmVersion:    nodeVersionInfoUpdate.Update.APIServerVersion.String(),
-				KubernetesVersion: nodeVersionInfoUpdate.Current.APIServerVersion.String(),
-			}, "kubernetes.install-intermediate-node-pattern")
-			if err != nil {
-				return err
+			if nodeVersionInfoUpdate.HasMajorOrMinorUpdate() {
```
MaximilianMeister: do we also have to put this around the `install-node-pattern` later after `kubeadm.upgrade.apply`? it's going to be introduced by https://github.com/SUSE/skuba/pull/585 

also possibly some other states don't have to be run if we don't have a major or minor update, e.g. restarting services?

ereslibre: I prefer to play safe and improve later :)

danielorf:
We seem to be running into LB pool quota:
```
1 error occurred:\n\t* openstack_lb_pool_v2.gangway_pool: Error creating pool: Expected HTTP response code [] when accessing [POST https://engcloud.prv.suse.net:9696/v2.0/lbaas/pools], but got 409 instead\n{"NeutronError": {"message": "Quota exceeded for resources: [\'pool\'].", "type": "OverQuota", "detail": ""}}\n
```

I am able to manually create loadbalancers in ECP though, so I'm not sure where this error is originating from.

internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -35,6 +35,17 @@ type NodeVersionInfoUpdate struct {
 	Update  kubernetes.NodeVersionInfo
 }
 
+func (nviu NodeVersionInfoUpdate) HasMajorOrMinorUpdate() bool {
+	if nviu.Current.IsControlPlane() {
+		if nviu.Update.APIServerVersion.Major() > nviu.Current.APIServerVersion.Major() ||
+			nviu.Update.APIServerVersion.Minor() > nviu.Current.APIServerVersion.Minor() {
```
MaximilianMeister: it's probably a very minor enhancement but as the major updates are very infrequent, we could check for minor first to speed up the check a little bit

ereslibre: We can change that later, however that would mean negative logic on the pattern installation (`if !isPatchUpdate { installIntermediatePattern }`).

MaximilianMeister: it shouldn't change any outside logic i meant like this:

```go
if nviu.Update.APIServerVersion.Minor() > nviu.Current.APIServerVersion.Minor() ||
        nviu.Update.APIServerVersion.Major() > nviu.Current.APIServerVersion.Major()
```

ereslibre: Ah misunderstood. Did it in level of importance but yes I don't even know if Kubernetes 2.0 is planned :)

PR #592, opened 2019-08-09
Print argument array of testrunner.

bergmannf:
This should help to debug the nightly job.


ci/infra/testrunner/testrunner.py
```diff
@@ -169,6 +169,7 @@ def main():
     cmd_test = commands.add_parser("test", parents=[test_args], help="execute tests")
     cmd_test.set_defaults(func=test)
 
+    print("Received arguments: {}".format(sys.argv))
```
stefsuse: why not use logger?

bergmannf:
The problem was not related to the passed arguments, but to the `pytest.main` call being made in the wrong directory.

PR #595, opened 2019-08-09
Fix output for upgrade plan (bsc#1144914)

suseclee:
When worker node plans out for upgrade, worker node's version is
dependent upon cluster version instead of the latest cluster version
available.

## Why is this PR needed?
```skuba node plan``` for worker node is misleading users for planning of upgrading worker nodes.
```
skuba node upgrade plan caasp-worker-clee-0 
skuba version: 0.9.3 (development) 4f6122f 20190809 go1.12.7
Current Kubernetes cluster version: 1.14.1

Node caasp-worker-clee-0 is up to date
```
The above info is correct for master node upgrade. However,  worker node upgrade is dependent upon current cluster version. 

The fix is change of the output for worker node
``` 
skuba node upgrade plan caasp-worker-clee-0 
skuba version: 0.9.3 (development) 4f6122f 20190809 go1.12.7
Current worker node version: 1.14.1
Current cluster version: 1.14.1

Node caasp-worker-clee-0 is up to date
```

Fixes #560

### DIFF
--- a/pkg/skuba/actions/node/upgrade/plan.go
+++ b/pkg/skuba/actions/node/upgrade/plan.go
@@ -20,8 +20,6 @@ package upgrade
 import (
 	"fmt"
 
-	"github.com/SUSE/skuba/internal/pkg/skuba/kubeadm"
-	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
 	upgradenode "github.com/SUSE/skuba/internal/pkg/skuba/upgrade/node"
 	"github.com/SUSE/skuba/pkg/skuba"
 )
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
This looks like an alternative to #560 the description should say "Fixes https://github.com/SUSE/avant-garde/issues/694"

It's another way of addressing the problem, however it does only hide the fact to the user that there is an update. with creating this bug I had more in mind to tell the user that the control plane nodes need to be updated before, but still keeping the info that there is a pending update.

And changing the output string would also mean that the `node upgrade plan` output is not uniform anymore, which would require adapting the e2e tests.

Nevertheless thanks for looking into this, and sorry I should have marked the bug in progress when I started to work on it

ereslibre:
I think a mix of https://github.com/SUSE/skuba/pull/560 and this PR would be the ideal solution.

https://github.com/SUSE/skuba/pull/560 is also addressing some cryptic output that we had on `skuba` regarding the upgrade, so I think this commit could be integrated in this PR in the way that all `skuba node upgrade plan` would always show:

* Current cluster version
* Latest kubernetes version
* Current node version

The latest one would basically show the `kubelet` version on that specific node as well, for all nodes (workers and control planes). So we have this PR's main feature as well there.

WDYT?

MaximilianMeister:
> `Current node version` would basically show the `kubelet` version on that specific node as well, for all nodes (workers and control planes). So we have this PR's main feature as well there.
> 
> WDYT?

sounds good, I have adapted https://github.com/SUSE/skuba/pull/560 now to include your commit @suseclee and adapt it to @ereslibre's comment - I'll close this PR then

PR #596, opened 2019-08-12
Rename generate-cloud-provider-conf file, bsc#1145082

mjura:
## Why is this PR needed?

Rename generate-cloud-provider-conf script to generate-cpi-conf.sh,
because of regex in package spec file.
It will include generate-cpi-conf.sh in the package.

Fixes # bsc#1145082

## What does this PR do?

It will rename generate-cloud-provider-conf script to generate-cpi-conf.sh,
because of regex in package spec file.

### DIFF
--- a/ci/infra/openstack/cloud-provider.tf
+++ b/ci/infra/openstack/cloud-provider.tf
@@ -14,6 +14,6 @@ resource "null_resource" "generate_cloud_provider_conf" {
       OS_PUBLIC_NET_ID     = "${data.openstack_networking_network_v2.external_network.id}"
     }
 
-    command = "bash generate-cloud-provider-conf"
+    command = "bash generate-cpi-conf.sh"
   }
 }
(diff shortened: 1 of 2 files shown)


ci/infra/openstack/cloud-provider.tf
```diff
@@ -14,6 +14,6 @@ resource "null_resource" "generate_cloud_provider_conf" {
       OS_PUBLIC_NET_ID     = "${data.openstack_networking_network_v2.external_network.id}"
     }
 
-    command = "bash generate-cloud-provider-conf"
+    command = "bash generate-cpi-conf.sh"
   }
```
jordimassaguerpla: Sorry I don't get this. Only changing this makes generate-cpi-conf.sh file be in the package?

kravciak:
Which `regex in package spec file`? Can you please elaborate how renaming `cloud-provider-conf` to `cpi-conf.sh` fixes the bug?

ci/infra/openstack/cloud-provider.tf
mjura: yes, this is because
```
for file in ci/infra/openstack/*.*; do
  install -p -m 0644 $file %{buildroot}/%{_datadir}/caasp/terraform/openstack/
done
```
so it means that only files `*.*` will be installed

PR #601, opened 2019-08-12
Use run_kubectl within upgrade e2e tests

MaximilianMeister:
# Why is this PR needed?

To fix the `test_upgrade_apply_user_lock` e2e test

## What does this PR do?

It's not necessary to have functions as they're only used once.
Also the outputformat option got removed for some reason and that
broke the kured lock e2e test.


MaximilianMeister:
part of https://github.com/SUSE/skuba/pull/603 now

PR #603, opened 2019-08-13
Create parametrized e2e pipeline

MaximilianMeister:
## Why is this PR needed?

We had massive trouble getting parallel e2e tests to work with https://ci.suse.de/job/caasp-jobs/job/caasp-v4-upgrade-nightly/

it's just not possible to run parallel tests at the moment in one single workspace.

1. testrunner only creates one `test-cluster` init folder
2. terraform creates one statefile

to fix this we would need to refactor the testrunner and also it's cleanup logic

## What does this PR do?

* Add a new pipeline to allow passing the name of the e2e test to run within a single e2e pipeline
* Drop the old upgrade specific pipeline

The benefit is also that we can easily add new e2e tests


MaximilianMeister:
the `jenkins/skuba-jjb-validation` job fails because it can't find the new template, but it's added within this PR, so i'm not sure if that's ok or not

```
jenkins_jobs.errors.JenkinsJobsException: Failed to find suitable template named '{name}-e2e-nightly'
```

PR #605, opened 2019-08-13
Print different versioning warnings depending on tag name

ereslibre:
## Why is this PR needed?

Depending on the tag that is pointing to a given commit, `skuba`
will give different notices to the user, depending on the maturity
of the release.

If the tag is "final" (no `alpha`, `beta` or `rc`), but the build
mode is not release, also print a warning about it.


ereslibre:
@jordimassaguerpla with this approach we can even get rid of the `VERSION` file; the version can be read from the tag directly on build time. I can improve this PR to handle that case as well.

Makefile
```diff
@@ -22,11 +22,13 @@ GOBINPATH    := $(shell $(GO) env GOPATH)/bin
 VERSION      := $(shell cat VERSION)
 COMMIT       := $(shell git rev-parse --short HEAD 2>/dev/null)
 BUILD_DATE   := $(shell date +%Y%m%d)
+TAG          := $(shell git show-ref --tags | grep $(COMMIT) | awk '{print $2}' | cut -d/ -f3)
```
nirmoy: This should be `git describe --tags` I think because 
```
nirmoy@chapman:~/go/src/github.com/SUSE/skuba> git show-ref --tags
c30833c271fdbd935ed159969032fa4715382395 refs/tags/v0.1.0
96a14d2d2c12a08871f25773feaf392dc5a701ca refs/tags/v0.2.0
d4343eb8035b3adafcb34604114f3f28d24a46dc refs/tags/v0.2.1
f55963df90dab3bfd33c977f239c7db31820e011 refs/tags/v0.3.0
eed514f2e5c868ea66b7eeef3442392ae628010e refs/tags/v0.3.1
d3698ec303c0e80f27dfefe4acb1d9a2ff6e3c20 refs/tags/v0.3.1.1
dd96fed5c766de8c3c845b23f761fdf41b4f4030 refs/tags/v0.4.0
67e823908db569d2992fc5bc1907663d21e2e393 refs/tags/v0.5.0
e3b493b0151fbbe5349cb10fb2a8b58160002b26 refs/tags/v1.4.0
851cca8e52c49fcb5e20bfb17e3e3b3611e6a6f9 refs/tags/v1.4.1
nirmoy@chapman:~/go/src/github.com/SUSE/skuba> git describe --tags
v1.4.1
nirmoy@chapman:~/go/src/github.com/SUSE/skuba> make
GO111MODULE=on go install -mod=vendor -ldflags "-X=github.com/SUSE/skuba/pkg/skuba.Version= -X=github.com/SUSE/skuba/pkg/skuba.Commit=1dd16815 -X=github.com/SUSE/skuba/pkg/skuba.BuildDate=20190813 -X=github.com/SUSE/skuba/pkg/skuba.Tag=" -tags development ./cmd/...
rm -f /home/nirmoy/go/bin/kubectl-caasp
ln -s /home/nirmoy/go/bin/skuba /home/nirmoy/go/bin/kubectl-caasp
nirmoy@chapman:~/go/src/github.com/SUSE/skuba> skuba version 
skuba version: untagged (development) 1dd16815 20190813 go1.12.6

```

ereslibre: I think you didn't tag a new version on that specific commit, that's why it's untagged. If you try the PR, and then:

`~/p/g/s/g/S/skuba (dynamic-version-warning) > git tag v0.1.0-alpha2
~/p/g/s/g/S/skuba (dynamic-version-warning) > make
GO111MODULE=on go install -mod=vendor -ldflags "-X=github.com/SUSE/skuba/pkg/skuba.Version=0.1.0 -X=github.com/SUSE/skuba/pkg/skuba.Commit=44589957 -X=github.com/SUSE/skuba/pkg/skuba.BuildDate=20190813 -X=github.com/SUSE/skuba/pkg/skuba.Tag=v0.1.0-alpha2" -tags development ./cmd/...
rm -f /home/ereslibre/projects/go/bin/kubectl-caasp
ln -s /home/ereslibre/projects/go/bin/skuba /home/ereslibre/projects/go/bin/kubectl-caasp
~/p/g/s/g/S/skuba (dynamic-version-warning) > skuba version
** This is an ALPHA release and NOT intended for production usage. **
skuba version: 0.1.0 (development) 44589957 (tagged as "v0.1.0-alpha2") 20190813 go1.12.7
`

It should work with any kind of tag, like `v0.1.0-alpha`, `v0.1.0-beta`, `v0.1.0-rc`, `v0.1.0`. Without a tag on the current commit, it will assume it's an untagged version and report it.

`git describe --tag` always returns a tag, but we only want to report as tagged commits that are really tagged I'd say.

pkg/skuba/version.go
```diff
@@ -26,26 +26,36 @@ var (
...
+	skubaVersion := SkubaVersion{
 		Version:   Version,
 		BuildType: BuildType,
 		BuildDate: BuildDate,
 		Commit:    Commit,
+		Tag:       Tag,
 		GoVersion: runtime.Version(),
 	}
+	if skubaVersion.Version == "" {
+		skubaVersion.Version = "untagged"
```
stefsuse: should this just unset skubaVersion.Tag? I don't see this checked anywhere and the version check for untagged release is:

```	
case skubapkg.Tag == "":
    fmt.Println("** This is an UNTAGGED version and NOT intended for production usage. **")
```

cmd/skuba/skuba.go
```diff
@@ -49,14 +51,29 @@ func newRootCmd() *cobra.Command {
...
 	if err := cmd.Execute(); err != nil {
 		os.Exit(1)
 	}
 }
 
+func printVersionStatement() {
+	switch {
+	case skubapkg.Tag == "":
+		fmt.Println("** This is an UNTAGGED version and NOT intended for production usage. **")
+	case strings.Contains(skubapkg.Tag, "alpha"):
```
stefsuse: right now we have tagged releases without suffixes that will report as "production" releases. 
Maybe we can also check for versions starting with 0 (ie 0.9.4) and report those as alpha as a catch all, then release GMC as v1.0.0.

pkg/skuba/version.go
ereslibre: Tag will already be empty in this case; if `Version` was empty is because `Tag` was (it couldn't be inferred).

cmd/skuba/skuba.go
ereslibre: I didn't want to care about the past because this PR will be effective once it merges; at that point is when we can start tagging with the new method. Am I misunderstanding?

ereslibre: In fact I did misunderstand, sorry. We are about to release a couple of RC's and after that we'll go with a "final" (no suffix) version -- that doesn't need to be `1.0.0`.

All things considered I don't think the benefit for this intermediate time is too much to explicitly catch the `Major() == 0` case (besides, we have released `0.X.Y` as alphas, betas and rc's at this time, so `Major() == 0` is not giving a "maturity" level).

Makefile
ereslibre: Thank you @nirmoy, updated the PR so if the current HEAD is not a tag we will use the `git describe --tag` output instead, like this:

```
~/p/g/s/g/S/skuba (dynamic-version-warning) > skuba version
skuba version: untagged (v0.9.4-17-g1602d982) (development) 20190814 go1.12.7
```

jordimassaguerpla:
> @jordimassaguerpla with this approach we can even get rid of the `VERSION` file; the version can be read from the tag directly on build time. I can improve this PR to handle that case as well.

How will skuba know which version it is when you run "skuba version" without the VERSION file?

jordimassaguerpla:
I like it but I won't approve it until we are done with CaaSPv4 RC1 just in case we need to create a new tag for RC1. Thanks.

jordimassaguerpla:
> > @jordimassaguerpla with this approach we can even get rid of the `VERSION` file; the version can be read from the tag directly on build time. I can improve this PR to handle that case as well.
> 
> How will skuba know which version it is when you run "skuba version" without the VERSION file?

I see it is done in build time. Forget about my comment ;)

ereslibre:
@jordimassaguerpla feel free to remove the `do not merge` label when appropriate.

PR #606, opened 2019-08-13
[WIP] put ssh agent sock in $TMP and verify length

dannysauer:
## Why is this PR needed?

Long test case names result in a full path which exceed the 107 character Linux (and general UNIX) socket path length.  This causes the ssh agent to not work, as it can't bind to the socket.

## What does this PR do?

This PR generates an md5sum of the unique workspace name, and uses that as a directory name in $TMP.  This creates a fixed-length path for the socket to live in despite test names of arbitrary length.  This PR also adjusts the cleanup process (to remove teh socket and, if otherwise empty, the new tempdir) and unifies calls to a utils method which gets the ssh socket name.  The sock calculating function also checks the max length and raises an exception if allowed length is exceeded.

I think the code speaks for itself; if not, that's a code problem which should be fixed. :)

## Info for QA

N/A

### Related info

Why does a Unix socket's path have to be under 100-ish chars: https://unix.stackexchange.com/a/367012/48650

Most nightly tests don't run

Tests should all work

## Docs

N/A; this only impacts internal testing

# Merge restrictions after RC

(Please do not edit this)

        3 approvals:
            1 developer
            1 QA

### DIFF
--- a/ci/infra/testrunner/utils/utils.py
+++ b/ci/infra/testrunner/utils/utils.py
@@ -1,9 +1,11 @@
 import glob
+import hashlib
 import logging
 import os
 import shutil
 import subprocess
 from functools import wraps
+from tempfile import gettempdir
 
 import requests
 from timeout_decorator import timeout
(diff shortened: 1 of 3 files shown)


ci/infra/testrunner/utils/utils.py
```diff
@@ -65,6 +67,23 @@ def cleanup_files(files):
         for file in files:
             Utils.cleanup_file(file)
 
+    def ssh_cleanup(self):
+        """Remove ssh sock files"""
+        # TODO: also kill ssh agent here? maybe move pkill to kill_ssh_agent()?
```
MaximilianMeister: i guess the [WIP] is because of this TODO here? let's follow up on that in a separate PR to unblock upgrade testing

dannysauer: That, and because I hadn't validated that it actually ran at the time I created the PR.  So mostly because I forgot to remove the WIP. :D

PR #608, opened 2019-08-14
(bsc#1145599) Explicitly enable ip forward in autoyast

lcavajani:
## Why is this PR needed?

This commit avoids disabling ip_forward
in /etc/sysctl.conf because this file takes
precedence over skuba kernel configuration files
in /etc/sysctl.d.

Without this commit, ip forward is disabled after
the reboot of a node.

Fixes bsc#1145599

The following is in `/etc/sysctl.conf`

net.ipv4.ip_forward = 0
net.ipv6.conf.all.forwarding = 0

The kernel options are not even defined in `/etc/sysctl.conf`

# Merge restrictions after RC

(Please do not edit this)

        3 approvals:
            1 developer
            1 QA


lcavajani:
@kravciak done, Thanks for testing !

danielorf:
This needs to be rebased to fix CI failures.  This is due to changes in https://github.com/SUSE/skuba/pull/611/.  Anything older requires a rebase.

PR #610, opened 2019-08-14
Set current kubernetes version in the upgrade e2e tests

MaximilianMeister:
Ideally we have another way of getting the latest version,
e.g. let the testrunner parse the versions.go file


MaximilianMeister:
> I don't like hardcoding it.

me neither, created https://github.com/SUSE/avant-garde/issues/767 for this

PR #612, opened 2019-08-14
Modify upgrade test to use in HA env

drpaneas:
## Why is this PR needed?

The only supported scenario in CaaSP is the multi-master multi-node one.  The current test for upgrades is testing an unsupported scenario. 

Fixes https://github.com/SUSE/avant-garde/issues/768

## What does this PR do?

This PR makes the test to run in an HA environment.

The test is failing already with one master/worker because of an actual product bug https://bugzilla.suse.com/show_bug.cgi?id=1145630 . T

## Info for QA

This PR is not fixing the bug, it just makes the test run into HA env (so maybe there are more bugs hidden here). There's no need to validate anything here.

### Related info

/nothing

The logs of the nightly jenkins runs should show this is using 1 master 1 worker

The logs of the nightly jenkins runs should show this is using 3 masters and 3 workers

## Docs

/nothing

# Merge restrictions after RC

(Please do not edit this)

        3 approvals:
            1 developer
            1 QA

### DIFF
--- a/ci/infra/testrunner/tests/test_skuba_upgrade.py
+++ b/ci/infra/testrunner/tests/test_skuba_upgrade.py
@@ -8,7 +8,7 @@ CURRENT_VERSION = "1.15.0"
 
 @pytest.fixture
 def setup(request, platform, skuba):
-    platform.provision(num_master=1, num_worker=1)
+    platform.provision(num_master=3, num_worker=3)
 
     def cleanup():
         platform.cleanup()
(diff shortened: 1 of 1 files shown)


ci/infra/testrunner/tests/test_skuba_upgrade.py
```diff
@@ -8,7 +8,7 @@
 
 @pytest.fixture
 def setup(request, platform, skuba):
-    platform.provision(num_master=1, num_worker=1)
+    platform.provision(num_master=3, num_worker=3)
```
MaximilianMeister: you could also drop the args, as it should be the default, but it's a minor detail

drpaneas: I would prefer to have the args there, so it's obvious to the reader/code-reviewer/qa how many workers and masters are needed.

jordimassaguerpla:
@danielorf This can be merged when the tests pass.

danielorf:
@MaximilianMeister Where are the hardcoded node names?  I can't find them anywhere.

ci/infra/testrunner/tests/test_skuba_upgrade.py
```diff
@@ -133,17 +133,21 @@ def test_upgrade_apply_user_lock(setup, platform, kubectl, skuba):
     setup_kubernetes_version(skuba, PREVIOUS_VERSION)
 
     # lock kured
-    kubectl.run_kubectl("-n kube-system annotate ds kured weave.works/kured-node-lock='{\"nodeID\":\"manual\"}'")
+    kubectl.run_kubectl(
+        "-n kube-system annotate ds kured weave.works/kured-node-lock='{\"nodeID\":\"manual\"}'")
 
     outs = {}
     for (r, n) in [("master", 0), ("worker", 0)]:
```
MaximilianMeister: @danielorf here e.g. the numbers should be applied from a CONSTANT, e.g. `MASTER_NUM` and `WORKER_NUM`

otherwise we'll only upgrade one worker and one master

pablochacin: Event better, you can retrieve the number of masters and workers using skuba: 
https://github.com/SUSE/skuba/blob/master/ci/infra/testrunner/tests/conftest.py#L41-L48

pablochacin:
I marked this PR as `do not merge` as it conflicts with https://github.com/SUSE/skuba/pull/628/files which deprecates the `setup` fixture in favor of the most complete `bootstrap`.  So the change in the number of nodes should be introduced in `bootstrap'. However, such change would make all tests to use that configuration.

MaximilianMeister:
@drpaneas I've followed up on this here https://github.com/SUSE/skuba/pull/635 - feel free to close this one

PR #614, opened 2019-08-15
Run testrunner scenarios instead of ginko

chentex:
Ginkos test provide a limited test suite, having the PRs and nightlys
run testrunner's cases actually gives a greatear view in terms of
product stability

Adds new parameter module to test command

This will allow to specify which folder containing test to execute.

Also created a fodler called prs in the tests folder that contains the
test that need to stay green in each PR that is created.

The nightly job will run all of the tests.


drpaneas:
We need to document how people can add scenarios there as well.

ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
```diff
@@ -49,26 +51,26 @@ pipeline {
...
         } }
 
-        stage('Cluster Deployment') { steps {
-            sh(script: 'make -f skuba/ci/Makefile deploy', label: 'Deploy')
-            archiveArtifacts("skuba/ci/infra/${PLATFORM}/terraform.tfstate")
-            archiveArtifacts("skuba/ci/infra/${PLATFORM}/terraform.tfvars.json")
-        } }
+        stage('Run test workers') { steps {
+            sh(script: "make -f skuba/ci/Makefile test_prs", label: "test_prs")
+            sh(script: "make --keep-going -f skuba/ci/Makefile cleanup", label: 'Cleanup')
```
tdaines42: If you move the clean up here instead of in the post block and the `test_prs` step fails we're going to end up with leftover clusters.

ci/jenkins/pipelines/skuba-nightly.Jenkinsfile
```diff
@@ -6,39 +6,41 @@ pipeline {
...
         } }
 
-        stage('Cluster Deployment') { steps {
-            sh(script: 'make -f skuba/ci/Makefile deploy', label: 'Deploy')
-            archiveArtifacts("skuba/ci/infra/${PLATFORM}/terraform.tfstate")
-            archiveArtifacts("skuba/ci/infra/${PLATFORM}/terraform.tfvars.json")
-        } }
+        stage('Run test Nightly') { steps {
+            sh(script: "make -f skuba/ci/Makefile test_nightly", label: "test_nightly")
+            sh(script: "make --keep-going -f skuba/ci/Makefile cleanup", label: 'Cleanup')
```
tdaines42: Same as the last comment

ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
chentex: I will add it also in the post section.

ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
```diff
@@ -52,28 +51,28 @@ pipeline {
...
             sh(script: 'make -f skuba/ci/Makefile pre_deployment', label: 'Pre Deployment')
             sh(script: 'make -f skuba/ci/Makefile pr_checks', label: 'PR Checks')
-            sh(script: 'cd skuba; make install; cd ../', label: 'Install skuba')
+            sh(script: "pushd skuba; make -f Makefile install; popd", label: 'Build Skuba')
         } }
 
-        stage('Cluster Deployment') { steps {
-            sh(script: 'make -f skuba/ci/Makefile deploy', label: 'Deploy')
-            archiveArtifacts("skuba/ci/infra/${PLATFORM}/terraform.tfstate")
-            archiveArtifacts("skuba/ci/infra/${PLATFORM}/terraform.tfvars.json")
```
tdaines42: We're no longer logging files that could help with debugging

ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
```diff
@@ -52,28 +51,28 @@ pipeline {
...
-           dir("skuba") {
-             sh(script: 'make build-ginkgo', label: 'build ginkgo binary')
-             sh(script: "make setup-ssh", label: "Setup SSH")
-             sh(script: "make test-e2e", label: "End-to-end tests")
-       } } }
     }
     post {
         always {
-            sh(script: 'make --keep-going -f skuba/ci/Makefile post_run', label: 'Post Run')
-            zip(archive: true, dir: 'testrunner_logs', zipFile: 'testrunner_logs.zip')
```
tdaines42: These logs should still be archived

drpaneas:
Vicente is gone for long summer vacations and he will be back before the workshop at September. In the meantime, we need to activate this. Since you are quite knowledgeable about the CI, it would be really great and awesome if you can help further with this. Any commits to this PR would be really appreciated :)

pablochacin:
I have a big concern about this PR. It assumes that we are replacing integration tests with e2e tests which is neither the intention not necessarily a good idea due to the implicit overhead of e2e tests. Integration and e2e tests should have their own separate pipelines. 

Another discussion is if all integration tests are going to be developed using the testrunner test library. There are good arguments to keep unit/integration tests in ginkgo. This topic has been discussed extensively but seems that there's still not a common understanding about this.

pablochacin:
I had missed the fact that this PR introduces a test suite for PRs but I'm not sure about the criteria to put tests in that suite. For example, the cilium tests and the update tests.

tdaines42:
> I had missed the fact that this PR introduces a test suite for PRs but I'm not sure about the criteria to put tests in that suite. For example, the cilium tests and the update tests.

@pablochacin Do you think we should leave this one for now? Then I can create a separate PR for a pipeline for the testrunner tests?

pablochacin:
@tdaines42 I think we should put this on hold until we clarify the points I mentioned in the first comment.

tdaines42:
Rebased on to master, squashed commits that made sense, made it so tests marked disruptive or flaky do not run for prs. 

Now any test added to `ci/infra/testrunner/tests/`  that doesn't have `@pytest.mark.flaky` or `@pytest.mark.disruptive` will run during a pr

tdaines42:
Marked cilium and dockercaps tests as flaky since they weren't always passing. Maybe these can be fixed later.

pablochacin:
@tdaines42 Overall I like the PR but I wasn't expecting to run ALL tests for PR just those marked as for PR . I think this gives more control on selecting which one we actually want to run. 

Also, marking the `provision` fixture as `session` makes that all tests that use that fixture will share one instance. That's not a problem with non-disruptive tests, but then disruptive tests must use another fixture. I think this is problematic (this is the problem I mentioned in the call that I haven't found a way for selection the scope of a fixture dynamically) .

bergmannf:
I think making all the fixtures `session` breaks the `update` tests - they are based around the fact that the cluster is bootstrapped with an older version of `k8s` initially.

I guess the testrunner has to provide a set of fixtures for each `scope` of `pytest` - and maybe make the `session`-wide ones not prefixed.

Then tests that need their cluster newly generated still have the chance to get one.

pablochacin:
@tdaines42  I have found a way for dynamically setting the fixture scope. I have a [POC](https://gist.github.com/pablochacin/23ce8ddee5cf9c6efb6bcfb3e267f0f8)  and also a [WIP](https://github.com/pablochacin/skuba/commit/370dfb864bae3d8bcc960013bfad4d086f591044)  for implementing in testurunner. I have some issues with python imports, as I had to slightly refactor the code (tl;dr: fixtures must be in a separate file from `conftest.py`). Hope to have a working version soon.

tdaines42:
Not really sure if we need to set the scope dynamically aren't the upgrade tests all being ran on their own in separate pipelines?

pablochacin:
@tdaines42  I think that we need this. Otherweise we are forcing the scope session for all tests that use the fixture. with this change, we can select if the tests will share the fixture or not. It is IMHO safest because unless you explicitly reusing the cluster, one new cluster will be provisioned. The best part is that we don't have to modify either the fixture nor the tests. It is only a cli parameter.

I don't see which drawback will have introducing this feature and see many advantages.

tdaines42:
I removed the session scope from the fixtures and made a few fixes

tdaines42:
@pablochacin At this point we only really have one test running so I think we can add more marks later and change the filter when we get more. This PR is already getting out of control so I'd rather do follow ups to make any further changes unless they are necessary.

pablochacin:
@tdaines42 I agree we must limit the scope of this PR. Now that we have reverted the session scope, I think we can move forward with this PR and solve that cluster reuse. 

Regarding the markers, I'm not 100% sure about using this implicit rule "if is not market as such and such" then it will be run. I would have prefered an explicit marke for PR. But I think we can revise that also later. However that will change how developers develop.  I leave others to decide on this.

ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
```diff
@@ -52,32 +51,36 @@ pipeline {
...
+            archiveArtifacts(artifacts: "skuba/ci/infra/${PLATFORM}/terraform.tfstate", allowEmptyArchive: true)
+            archiveArtifacts(artifacts: "skuba/ci/infra/${PLATFORM}/terraform.tfvars.json", allowEmptyArchive: true)
+            archiveArtifacts(artifacts: 'testrunner_logs/**/*', allowEmptyArchive: true)
+            junit('skuba/ci/infra/testrunner/*.xml')
```
pablochacin: We should also collect `testrunner.log` which has the DEBUG level output from the testrunner

ci/jenkins/pipelines/skuba-nightly.Jenkinsfile
```diff
@@ -18,33 +16,36 @@ pipeline {
...
-       }
-       cleanup {
-           dir("${WORKSPACE}") {
-               deleteDir()
-           }
-           sh(script: "rm -f ${SKUBA_BIN_PATH}; ", label: 'Remove built skuba')
+        always {
+            archiveArtifacts(artifacts: "skuba/ci/infra/${PLATFORM}/terraform.tfstate", allowEmptyArchive: true)
+            archiveArtifacts(artifacts: "skuba/ci/infra/${PLATFORM}/terraform.tfvars.json", allowEmptyArchive: true)
+            archiveArtifacts(artifacts: 'testrunner_logs/**/*', allowEmptyArchive: true)
```
pablochacin: same here we should also collect testrunner.log file

ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
```diff
@@ -52,32 +51,36 @@ pipeline {
...
+            sh(script: "make --keep-going -f skuba/ci/Makefile cleanup", label: 'Cleanup')
         }
         cleanup {
+            dir("${WORKSPACE}@tmp") {
+                deleteDir()
+            }
+            dir("${WORKSPACE}@script") {
+                deleteDir()
+            }
+            dir("${WORKSPACE}@script@tmp") {
```
pablochacin: maybe I'm not understanding what's going on here but if we have already deleted `WORKSPACE/script` why delete `WORKSPACE/script/tmp`?

MaximilianMeister: AFAIR what @chentex told me, it was necessary because it had the `@` in its name, or something. it was found during manual runs on the jenkins worker itself that only like that the dirs get cleaned up

hwoarang:
PR needs to be rebased due to conflicts

ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
tdaines42: Done

ci/jenkins/pipelines/skuba-nightly.Jenkinsfile
tdaines42: done

ci/jenkins/pipelines/skuba-nightly.Jenkinsfile
```diff
@@ -18,33 +16,37 @@ pipeline {
    stages {
         stage('Getting Ready For Cluster Deployment') { steps {
             sh(script: 'make -f skuba/ci/Makefile pre_deployment', label: 'Pre Deployment')
-            sh(script: 'cd skuba; make install; cd ../', label: 'Install skuba')
+            sh(script: "pushd skuba; make -f Makefile install; popd", label: 'Build Skuba')
```
dannysauer: Same deal with `bash-specific code :/

ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
```diff
@@ -52,32 +51,37 @@ pipeline {
         stage('Getting Ready For Cluster Deployment') { steps {
             sh(script: 'make -f skuba/ci/Makefile pre_deployment', label: 'Pre Deployment')
             sh(script: 'make -f skuba/ci/Makefile pr_checks', label: 'PR Checks')
-            sh(script: 'cd skuba; make install; cd ../', label: 'Install skuba')
+            sh(script: "pushd skuba; make -f Makefile install; popd", label: 'Build Skuba')
```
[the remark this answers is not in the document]
tdaines42: Yeah I think it was copied from other pipelines

PR #616, opened 2019-08-15
Make sure nodes are ready and upgraded

MaximilianMeister:
## Why is this PR needed?

Adding 2 assertions:

* Node must be ready before attempting an upgrade
* Node must be upgraded to a specific version (additional version check)

These 2 conditions are somewhat crucial to assert the right behavior
during upgrades

## What does this PR do?

adds some more assertions at the right time

# Merge restrictions after RC

(Please do not edit this)

        3 approvals:
            1 developer
            1 QA

### DIFF
--- a/ci/infra/testrunner/tests/test_skuba_upgrade.py
+++ b/ci/infra/testrunner/tests/test_skuba_upgrade.py
@@ -27,6 +27,14 @@ def setup_kubernetes_version(skuba, kubectl, kubernetes_version=None):
     kubectl.run_kubectl("wait --for=condition=ready nodes --all --timeout=5m")
 
 
+def node_is_upgraded(kubectl, node_name):
+    return kubectl.run_kubectl("get nodes {} -o jsonpath='{{.status.nodeInfo.kubeletVersion}}'".format(node_name)).find(CURRENT_VERSION) != -1
+
+
+def node_is_ready(kubectl, node_name):
+    return kubectl.run_kubectl("get nodes {} -o jsonpath='{{range @.status.conditions[*]}}{{@.type}}={{@.status}};{{end}}'".format(node_name)).find("Ready=True") != -1
+
+
 def test_upgrade_plan_all_fine(bootstrap, skuba, kubectl):
     """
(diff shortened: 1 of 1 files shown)


ci/infra/testrunner/tests/test_skuba_upgrade.py
```diff
@@ -116,7 +124,9 @@ def test_upgrade_apply_from_previous(setup, platform, skuba):
     outs = {}
     for (r, n) in [("master", 0), ("worker", 0)]:
         node = "my-{}-{}".format(r, n)
+        assert node_is_ready(kubectl, node)
         outs[node] = skuba.node_upgrade("apply", r, n)
+        assert node_is_upgraded(kubectl, node)
```
stefsuse: In the normal update case we want to check that the update timer is still running, right? (sorry if this is already in another test, I didn't see it)
`assert platform.ssh_run(r, n, "sudo systemctl is-enabled skuba-update.timer || :").find("disabled") == -1`

MaximilianMeister: skuba-update.timer being enabled is the default, if everything goes fine we safely can assume that it is enabled (if not the enable would have failed during the upgrade)

we have `test_upgrade_apply_user_lock` that disables skuba-update (and kured) and only then we explicitly check that it stays disabled

so all in all we don't need to check this specific scenario, as we would also have to check if kubelet and cri-o is running e.g. among all the other things we do during the upgrade

stefsuse: My thinking is that since there is logic to disable timers, it could happen that we disable something automatically. :) In the event of a refactor or skuba-update pkg update, if we have a test that checks for the timer errors will get caught right away. 

> [...] we would also have to check if kubelet and cri-o is running e.g. among all the other things we do during the upgrade 

Now that you mention it, that's a pretty good idea, too!

MaximilianMeister:
had to rebase due to some conflict, @stefsuse @lcavajani @mssola please approve again

PR #617, opened 2019-08-16
Add new wrapper fixture to create a cluster.

bergmannf:
The new fixture will create the cluster and all configuration in a /tmp
directory. This way it will be possible to have multiple clusters at the
same time, without them interfering with each other.

## Why is this PR needed?

I makes it possible to run tests in parallel, as long as they use the
new fixture, as they will have all their configuration in their own
directories.

This PR only impact tests.

## Info for QA

This PR only impacts the tests and testrunner, no QA impact.

# Merge restrictions after RC

(Please do not edit this)

        3 approvals:
            1 developer
            1 QA

### DIFF
--- a/ci/infra/testrunner/tests/conftest.py
+++ b/ci/infra/testrunner/tests/conftest.py
@@ -1,8 +1,19 @@
+import contextlib
+import copy
+import logging
+import os
 import pytest
 import platforms
+import tempfile
+import shutil
 from skuba import Skuba
 from kubectl import Kubectl
 from utils import BaseConfig
+
+
+LOGGER = logging.getLogger('testrunner')
+LOGGER.setLevel(logging.DEBUG)
+
 
 def pytest_addoption(parser):
     """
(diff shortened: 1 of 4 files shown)


ci/infra/testrunner/tests/conftest.py
```diff
@@ -12,6 +23,97 @@ def pytest_addoption(parser):
...
+    >>>     c.run("kubectl ...")
+
+    Once 'c' goes out of scope the cluster directory from skuba will be removed and
+    the terraform deployment will be destroyed.
+    """
+    def __init__(self, conf: BaseConfig, masters=1, workers=1, platform='openstack'):
+        self.masters = masters
+        self.workers = workers
+        self.workspace = Workspace()
+        self.config = BaseConfig(conf.yaml_path)
```
pablochacin: I Think we are breaking a lot the encapsulation of concerns by instantiating the BaseConfig class here. Not sure if this is necessary. The configuration is available by means of the `conf` fixture.

ci/infra/testrunner/tests/conftest.py
```diff
@@ -1,9 +1,20 @@
...
 import pytest
 import platforms
+import tempfile
+import shutil
 from skuba import Skuba
 from kubectl import Kubectl
 from utils import BaseConfig
 
+
+LOGGER = logging.getLogger('conftest')
```
pablochacin: This will make test logs go to a different logger than the testrunner's log. Not sure this is convenient.

ci/infra/testrunner/tests/conftest.py
```diff
@@ -12,6 +23,97 @@ def pytest_addoption(parser):
...
+        """Reset the config workspace, to not create all cluster resources in the same
+        directory.
+
+        """
+        LOGGER.debug(f"Adjusting workspace to {self.workspace.temporary_directory}")
+        self.config.workspace = self.workspace.temporary_directory
+        LOGGER.debug(f"Adjusting terraform directory to {self.workspace.temporary_directory}")
+        self.config.terraform.tfdir = self.workspace.temporary_directory
+
+    def __copy_configuration(self, original_config: BaseConfig):
```
pablochacin: I think the name is misleading: `copy_tf_config` maybe would be more appropriated.

ci/infra/testrunner/tests/conftest.py
```diff
@@ -12,6 +23,97 @@ def pytest_addoption(parser):
...
+        shutil.copytree(src, tgt)
+
+
+@pytest.fixture
+def cluster(conf: BaseConfig):
+    """Wrap the Cluster creation class in a function, so the tests don't have to
+    get the `conf` file.
+
+    """
+    def _(masters=1, workers=1, platform='openstack'):
```
pablochacin: I think platform shouldn't have a default value and should always be passed from the test using the `target` fixture. This guarantees the test uses the platform defined in the environment.

ci/infra/testrunner/tests/conftest.py
```diff
@@ -12,6 +23,97 @@ def pytest_addoption(parser):
...
+
+class Workspace:
+    def __init__(self):
+        self.temporary_directory = tempfile.mkdtemp(prefix="skuba-tests")
+
+    def cleanup(self):
+        shutil.rmtree(self.temporary_directory)
+
+
+class ClusterFixture(contextlib.AbstractContextManager):
```
pablochacin: I'm not sure using python's context has any advantage over using pytest fixture cleanup menchanis, which is simpler and more idiomatic to pytest.

pablochacin:
Overall, I'm not quite sure the changes should be implemented in the test library. I think the changes should be in the core testrunner. I think test library should be limited to test- related logic. Here there are a lot of setup and configuration logic.

ci/infra/testrunner/tests/conftest.py
bergmannf: Well the current testrunner setup breaks, when attempting to run tests in parallel (in Jenkins), because they all attempt to use the same workspace - that's why this is make a shallow copy of the configuration and after that resets the configuration paths for the workspace to point to a temporary directory.

If I just reset the directories on the initial `conf` reference, obviously all usages would get the same temporary directory and it would be useless.

ci/infra/testrunner/tests/conftest.py
bergmannf: I am using https://docs.pytest.org/en/latest/fixture.html#factories-as-fixtures - mostly because it seems that otherwise tests will end up implementing the same fixtures again and again themselves like in the upgrade test: 

https://github.com/SUSE/skuba/blob/master/ci/infra/testrunner/tests/test_skuba_upgrade.py#L9-L15

Which is just a duplicate of: https://github.com/SUSE/skuba/blob/master/ci/infra/testrunner/tests/conftest.py#L26-L31

Because I don't see a way that a single test could easily say it needs a 3 master/3 worker cluster right now, if they have to use the configuration file values.

pablochacin:
> I can't follow what you mean with the differentiation between `test library` and `core testrunner` here.

Tests should only have test related code and use the library provided by testrunner for setup (platform, skuba, kubectl, logs, ssh, ,,). It a simple separation of concerns in the code which hopefully will prevent the test code to grow exponentially. 

> I am fine using a different approach - but what we experienced when attempting to run the nightly tests in jenkins was that parallel tests won't work in the testrunner due to the single workspace.

This was a known issue and actually there was a card to prevent the cluster-init code to be hardcoded. What was missing was the tf files.  IMHO these changes should go to the testrunner in the proper place (`skuba` for cluster directory and `terraform` for copying the tf files)

> also some tests might need bigger cluster than others - e.g. the update test is currently using 1 master and 1 worker, but experience problems with 3 master / 3 workers [@drpaneas I think you found that])

Sure, that's something that can be implemented in the testlibrary. I like the idea of the `cluster'  fixture and the way you solved passing parameters is idiomatic in pytest. 

By the way there is already another fixture which provides a fully deployed cluster, which your cluster fixture should replace:

https://github.com/SUSE/skuba/blob/master/ci/infra/testrunner/tests/conftest.py#L41-L48

ci/infra/testrunner/tests/conftest.py
pablochacin: Yes, Factory fixture is a great idea and definitely something other tests can use. my comment is about using abstract contentect manager for automatic tear down.

ci/infra/testrunner/tests/conftest.py
```diff
@@ -12,6 +23,97 @@ def pytest_addoption(parser):
...
+        self.platform.cleanup()
+        LOGGER.info("Cleanup workspace")
+        self.workspace.cleanup()
+
+    def __adjust_directories(self):
+        """Reset the config workspace, to not create all cluster resources in the same
+        directory.
+
+        """
+        LOGGER.debug(f"Adjusting workspace to {self.workspace.temporary_directory}")
```
pablochacin: This should not be required. Calling the testrunner with a different workspace (can be defined as an environment variable) is preferable option

ci/infra/testrunner/tests/conftest.py
```diff
@@ -12,6 +23,97 @@ def pytest_addoption(parser):
...
+        self.config.workspace = self.workspace.temporary_directory
+        LOGGER.debug(f"Adjusting terraform directory to {self.workspace.temporary_directory}")
+        self.config.terraform.tfdir = self.workspace.temporary_directory
+
+    def __copy_tf_configuration(self, original_config: BaseConfig):
```
pablochacin: This logic should be implemented in [terraform.py](https://github.com/SUSE/skuba/blob/master/ci/infra/testrunner/platforms/terraform.py) see https://github.com/SUSE/avant-garde/issues/783

bergmannf:
I also realized now that this PR probably would have to rename the `stack` for *actually* running them in parallel.

Right now this would get conflicts when attempting to actually create a cluster in parallel.

jordimassaguerpla:
Hi!
Happy hacking

drpaneas:
> Hi!
> New code has been merged in master regarding terraform 0.12. Please rebase your branch to take this changes in.
> Happy hacking

I am pretty sure this is an automated message coming from Jordi's account. Anyway, this PR is no longer valid. Please close it if you think so as well.

jordimassaguerpla:
You are right. This is old and outdated. Closing.

PR #622, opened 2019-08-19
Make libvirt tfvars example more complete

rhafer:
## Why is this PR needed?

For making it easier to adapt the libvirt terraform files to own needes.

## What does this PR do?

Mention the "lb_repositories" setting, which allows to specific repos
specifically for the load balancer node. The loadbalancer node needs at
least the SLE-15 Basesystem and the HA addon (for haproxy).


MaximilianMeister:
now that the master branch is unblocked we can merge this, sorry for the delay

PR #623, opened 2019-08-19
Add e2e pipeline

tdaines42:
## Why is this PR needed?

We want to run the testrunner pytest tests against prs. At this time though we're not sure if we want to disable the Ginkgo ones.

Fixes #

## What does this PR do?

Creates a new pipeline to run the testrunner pytest tests.

### DIFF
--- a/ci/Makefile
+++ b/ci/Makefile
@@ -80,3 +80,9 @@ test_upgrade_apply_from_previous:
 .PHONY: test_upgrade_apply_user_lock
 test_upgrade_apply_user_lock:
 	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} test --verbose --suite test_skuba_upgrade.py --test test_upgrade_apply_user_lock
+
+.PHONY: test_pr_e2e
+test_pr:
+	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} test --verbose --suite test_cilium.py
+	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} test --verbose --suite test_deployments.py
+	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} test --verbose --suite test_skuba_upgrade.py
--- a/ci/infra/testrunner/tests/conftest.py
+++ b/ci/infra/testrunner/tests/conftest.py
@@ -27,6 +27,7 @@ def setup(request, platform, skuba):
 def provision(request, platform):
     platform.provision()
     def cleanup():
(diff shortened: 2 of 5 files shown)


ci/Makefile
```diff
@@ -80,3 +80,9 @@ test_upgrade_apply_from_previous:
 .PHONY: test_upgrade_apply_user_lock
 test_upgrade_apply_user_lock:
 	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} test --verbose --suite test_skuba_upgrade.py --test test_upgrade_apply_user_lock
+
+.PHONY: test_pr_e2e
+test_pr:
+	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} test --verbose --suite test_cilium.py
+	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} test --verbose --suite test_deployments.py
+	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} test --verbose --suite test_skuba_upgrade.py
```
MaximilianMeister: if this runs against every PR i would only run 1 upgrade test from the suite:

```
${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} test --verbose --suite test_skuba_upgrade.py --test test_upgrade_plan_from_previous
```

kravciak: `test_cilium.py` is not 100% stable until https://github.com/SUSE/skuba/blob/master/ci/infra/testrunner/tests/test_cilium.py#L16 is replaced.

pablochacin:
> this might take a bit of time and i'm not really sure if we should always run the tests for every PR

@MaximilianMeister I agree we shouldn't. This e2e pipeline should be run periodically against master, not with each PR. 

As we don't have at this point narrow scope integration tests [1] we should probably use one or  two e2e tests which offer a broad code coverage as part of the PR pipeline. We already have one in ginkgo. We can keep using it or replace with something similar in testrunner like the deployments tests. IMHO this is not important. 

[1]  https://martinfowler.com/bliki/IntegrationTest.html

MaximilianMeister:
> I think this pipeline should run nightly. Not per PR.

nightly e2e tests are already running here https://ci.suse.de/view/CaaSP/job/caasp-jobs/job/e2e/

if there are missing ones, they need to be added to this folder

MaximilianMeister:
> if there are missing ones, they need to be added to this folder

not all of them were enabled yet, i have done that now in https://github.com/SUSE/skuba/pull/624

PR #630, opened 2019-08-22
Don't use bootstrap fixture on upgrades

MaximilianMeister:
## Why is this PR needed?

The bootstrap fixture bootstraps a more recent cluster than we need in the upgrade tests.

We need to bootstrap a specific version of the cluster here `1.14.1`

## What does this PR do?

Don't use bootstrap fixture on upgrades

## Anything else a reviewer needs to know?

follow up from https://github.com/SUSE/skuba/pull/628

PR #631, opened 2019-08-22
Don't use bootstrap fixture on upgrades

MaximilianMeister:
## Why is this PR needed?

The bootstrap fixture bootstraps a more recent cluster than we need in the upgrade tests.

We need to bootstrap a specific version of the cluster here `1.14.1`

## What does this PR do?

Don't use bootstrap fixture on upgrades

follow up from https://github.com/SUSE/skuba/pull/628

fixing the glitch of https://github.com/SUSE/skuba/pull/630


MaximilianMeister:
rebase was necessary due to some conflict

PR #632, opened 2019-08-22
Addon structure

ereslibre:
## Why is this PR needed?

Fixes https://github.com/SUSE/avant-garde/issues/780

* Adds an addon structure where each addon will register itself
  against an addon map.

  * Each addon can register optional callbacks that will be called
    if present before and after applying the addon to the cluster.

  * These callbacks can be used to renew certificates, or perform
    any other actions.

  * By refactoring the addon struct to an addon map, remove several
    ad-hoc switches regarding addon logic, that checked what version
    should be returned depending on the requested addon.

* Refactors the code so each addon file contains everything related
  to that addon.

* Addons are applied locally as opposed to the previous general
  approach of scp'ing the manifests and applying them remotely.

  * While on `bootstrap` we have SSH access to the machine, during
    an addon upgrade SSH access is not desired, we should use the
    cluster as another consumer; for this reason, render the addons
    locally and apply them on the local machine with `kubectl`.

* `skuba cluster init` now iterates on the list of addons for the
  current Kubernetes version, and creates the addons that are
  registered for that version.

* Adds a `skuba-config` ConfigMap with an entry `SkubaConfiguration`,
  that holds a struct that gets (un)marshaled. This struct only
  contains one field right now: `AddonsVersion`, that is the map
  containing the current addons version for all deployed addons on
  the cluster. This is what is used to determine whether an addon
  should be updated.

  * Without this method, it's not trivial to decide in what exact
    version an addon is: consider an addon with several resources,
    should all resources be annotated? What happens if some of them
    are out of date while others are up to date, what version we
    choose? For this reason, a `skuba-config` ConfigMap holds the
    information about every addon version and manifest version.

  * Addons are always applied, unless the version hardcoded in skuba
    is less than the one present in the cluster (this can happen if
    a user is using an outdated `skuba` in a cluster where addons were
    upgraded with a newer `skuba`).

  * The ConfigMap is created upon `bootstrap` with the current version
    of all addons matching the Kubernetes version used to boostrap the
    cluster.

  * Every time an addon is applied, the ConfigMap gets updated with the
    new version and manifest version.

* Since the `tooling` image is not strictly an addon, move it to the
  `ComponentsVersion` struct, so we can still use different `tooling`
  images depending on the current cluster version if needed.

* Add `PSP` as a first class citizen addon: use an arbitrary "1.0.0"
  version since no images will be pulled in any case for these
  manifests, so the image version is not relevant.

* Addons have two Apply modes: `AlwaysRender` and
  `SkipRenderIfConfigFilePresent`.

  * `AlwaysRender` will always get the addon manifest template and
    render the last version of it, this is the desired behavior
    always except for a `skuba node bootstrap` operation.

  * `SkipRenderIfConfigFilePresent` will render the addon manifest
    template only if the config file is not present in disk. This
    is useful because we allow the user to run `skuba cluster init`,
    tweak the configuration if they want, and run
    `skuba node bootstrap` afterwards. In this case, we want to apply
    the user modified settings, instead of forcefully rerender.

Note well:

* When an addon has a new version or manifest revision (manifest
  revisions should *always* be bumped), it will be reapplied on the
  cluster when running `skuba addon upgrade apply`. This means that
  any user customization will be lost when performing this action.

  * A way to tackle this problem would be to allow for user provided
    `kustomize` patches that live inside the cluster; when we upgrade
    our manifests (the base), we run the user `kustomize` patches on
    top of them, applying the patched version to the cluster. This
    would ensure that we would be able to evolve our base manifests
    while keeping the user changes on every upgrade. Also, it should
    be possible to detect manifest conflicts early, by running the
    patch operation beforehand in a sandbox.


internal/pkg/skuba/kubernetes/resources.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+// DoesResourceExist check the given error from a client-go API call and returns
+// whether the resource exists. If no error was given, it will return true, if
+// the given error was a `IsNotFound` it will return false; otherwise it will
+// return an error
+func DoesResourceExist(err error) (bool, error) {
```
MaximilianMeister: can we name this a bit more clear, that it's about error resources?

internal/pkg/skuba/kured/kured.go
```diff
@@ -18,20 +18,20 @@
...
-		kubernetes.CurrentAddonVersion(kubernetes.Kured))
-}
+const (
+	kuredDSName             = "kured"
+	kuredLockAnnotationJson = `{"metadata":{"annotations":{"weave.works/kured-node-lock":"'{\"nodeID\":\"manual\"}'"}}}`
+)
 
-func KuredLockExists() (bool, error) {
+func LockExists(client clientset.Interface) (bool, error) {
 	clientSet, err := kubernetes.GetAdminClientSet()
```
MaximilianMeister: you can drop this `clientSet` instantiation here, and use the passed `client` below instead

internal/pkg/skuba/addons/addons.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "could not parse manifest template")
+	}
+	var rendered bytes.Buffer
+	err = template.Execute(&rendered, addonRenderContext{
+		Addon:  addon,
+		Config: addonConfiguration,
+	})
+	if err != nil {
+		return "", errors.Wrapf(err, "could not render manifest template")
```
mssola: nipicking: change the `Wrapf` for a `Wrap`

internal/pkg/skuba/addons/addons.go
```diff
@@ -0,0 +1,265 @@
...
+	if err := ioutil.WriteFile(addon.manifestPath(), []byte(addonManifest), 0600); err != nil {
+		return errors.Wrapf(err, "unable to write %s addon rendered template", addon.addon)
+	}
+	return nil
+}
+
+func (addon Addon) Apply(addonConfiguration AddonConfiguration, skubaConfiguration *skuba.SkubaConfiguration, applyBehavior ApplyBehavior) error {
+	klog.V(1).Infof("applying %q addon", addon.addon)
+	if addon.callbacks != nil {
+		if err := addon.callbacks.BeforeApply(addonConfiguration, skubaConfiguration); err != nil {
```
mssola: If I understand this correctly, if one addon fails but another doesn't, then we will get anyway an error but at least the successful addons won't have to be re-applied. Is that correct? Or this is an "all or nothing" situation? I'm a bit lost here :sweat_smile:

internal/pkg/skuba/addons/dex.go
```diff
@@ -0,0 +1,259 @@
...
+	dexCertExists, err := dex.DexCertExists(client)
+	if err != nil {
+		return errors.Wrap(err, "unable to determine if dex certificate exists")
+	}
+	err = dex.CreateCert(client, skubaconstants.PkiDir(), skubaconstants.KubeadmInitConfFile())
+	if err != nil {
+		return errors.Wrap(err, "unable to create dex certificate")
+	}
+	if dexCertExists {
+		if err := dex.RestartPods(client); err != nil {
```
mssola: As I've seen in `dex.RestartPods`, this function will not wait for the pods to reappear. Should we do this so this addon has a known state when `BeforeApply` is done? As in, should we double-check that restart really went fine before returning? Or this is done in another stage? Same for the gangway addon.

internal/pkg/skuba/addons/dex.go
```diff
@@ -0,0 +1,259 @@
...
+		}
+	}
+	return nil
+}
+
+func (DexCallbacks) AfterApply(addonConfiguration AddonConfiguration, skubaConfiguration *skuba.SkubaConfiguration) error {
+	return nil
+}
+
+const (
```
mssola: Wondering if we should also add CaaSP metadata into the different objects :thinking: Same for the gangway addon.

internal/pkg/skuba/addons/gangway.go
```diff
@@ -0,0 +1,191 @@
...
+		kubernetes.AddonVersionForClusterVersion(kubernetes.Gangway, addonRenderContext.Config.ClusterVersion).Version)
+}
+
+func renderGangwayTemplate(addonConfiguration AddonConfiguration) string {
+	return gangwayManifest
+}
+
+type GangwayCallbacks struct{}
+
+func (GangwayCallbacks) BeforeApply(addonConfiguration AddonConfiguration, skubaConfiguration *skuba.SkubaConfiguration) error {
```
mssola: The implementation of this function looks really similar to the dex one (with the difference on the secrets part). I'm wondering if we could share more code. This can be done on a later PR whenever we have more addons, though.

internal/pkg/skuba/gangway/gangway.go
```diff
@@ -96,21 +90,30 @@ func CreateCert(
...
-// GetGangwayImage returns gangway image registry
-func GetGangwayImage() string {
-	return images.GetGenericImage(skuba.ImageRepository, imageName,
-		kubernetes.CurrentAddonVersion(kubernetes.Gangway))
+func GangwaySecretExists(client clientset.Interface) (bool, error) {
+	_, err := client.CoreV1().Secrets(metav1.NamespaceSystem).Get(secretKeyName, metav1.GetOptions{})
+	return kubernetes.DoesResourceExist(err)
+}
+
+func GangwayCertExists(client clientset.Interface) (bool, error) {
```
mssola: `GangwayCertExists` and `RestartPods` are identical to the ones from the `dex` package. Can't we have a shared package for these two addons or something like that?

internal/pkg/skuba/kubernetes/resources.go
```diff
@@ -0,0 +1,36 @@
...
+	switch {
+	case err == nil:
+		return true, nil
+	case apierrors.IsNotFound(err):
+		return false, nil
+	}
```
mssola: Nitpicking: what's the benefit of using a `switch` statement here instead of mere `if` statements?

internal/pkg/skuba/skuba/configmap.go
```diff
@@ -0,0 +1,85 @@
...
+type SkubaConfiguration struct {
+	AddonsVersion kubernetes.AddonsVersion
+}
+
+func GetSkubaConfiguration() (*SkubaConfiguration, error) {
+	clientSet, err := kubernetes.GetAdminClientSet()
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting client set")
+	}
+	skubaConfiguration := SkubaConfiguration{}
```
mssola: Why don't you simply declare this object as a heap pointer, so then you don't have to derefence it all the time. Plus, at least from a C background, as it is it looks as if you are returning a reference to a local variable :grin:

internal/pkg/skuba/addons/addons.go
ereslibre: Exactly, if we get an error applying one addon, we will still continue trying to apply the rest, `DeployAddons` has the loop, we will `Errorf` but continue with the loop.

internal/pkg/skuba/addons/dex.go
ereslibre: Hm, I copied this from the previous behavior but I'll double check because I don't think it's even longer necessary as we are adding an annotation inside the pod templates, and thus, if we get any upgrade on those addons they will restart themselves.

internal/pkg/skuba/addons/dex.go
ereslibre: I added the caasp metadata on the places necessary for the pods to automatically restart if something changes. The real information is saved into the `skuba-config` ConfigMap.

internal/pkg/skuba/addons/gangway.go
ereslibre: Could be, however I wanted as a feature on this PR to leave every addon logic separate from each other. We can improve in future PR's though.

internal/pkg/skuba/kubernetes/resources.go
ereslibre: Thought it was more clear, nothing special. If you think it's more clear with `if` I can change it.

internal/pkg/skuba/skuba/configmap.go
ereslibre: Do you mean `skubaConfiguration := &SkubaConfiguration{}`?

ereslibre:
Wow, thanks for the fantastic reviews @MaximilianMeister and @mssola! I still have to review some things because PSP needs to be strictly applied before anything else so it's a special case.

I will check this on monday as well as address all your comments.

internal/pkg/skuba/kubernetes/resources.go
mssola: I think it's more clear as `if`, but I don't mind. It's really a nitpick

internal/pkg/skuba/skuba/configmap.go
mssola: Yeah

internal/pkg/skuba/kubernetes/resources.go
ereslibre: I have been trying to come up with a better name but I couldn't find one, given the transformation this function performs, do you have a better suggestion?

MaximilianMeister: i'm not very creative here either, but it would help if we know that it requires an error to be passed, maybe `DoesResourceExistWithError`?

my confusion is that the name `DoesResourceExist` implies passing a resource, but maybe it's just me

internal/pkg/skuba/addons/addons.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+}
+
+func addonsByPriority() []Addon {
+	sortedAddons := make([]Addon, len(Addons))
+	i := 0
+	for _, addon := range Addons {
+		sortedAddons[i] = addon
+		i++
+	}
```
mssola: Nitpicking: just:

```go
	for i, addon := range Addons {
		sortedAddons[i] = addon
	}
```

Since both slices have the same length, there is no need for that `i` variable outside of the for loop.

internal/pkg/skuba/addons/addons.go
```diff
@@ -0,0 +1,288 @@
...
+	cmd := exec.Command("kubectl", "apply", "--kubeconfig", skubaconstants.KubeConfigAdminFile(), "-f", "-")
+	stdin, err := cmd.StdinPipe()
+	if err != nil {
+		return err
+	}
+	if _, err := stdin.Write([]byte(renderedManifest)); err != nil {
+		return err
+	}
+	stdin.Close()
```
mssola: I'd move this line as a defer before this `if` statement. The reason being: what happens if we couldn't write for whatever reason but the stdin is not really closed? Then with the current code we'd get out with an error but stdin wouldn't be closed.

internal/pkg/skuba/addons/addons.go
ereslibre: But `Addons` is a map though, the first argument would be the `kubernetes.Addon` instead of the iteration.

internal/pkg/skuba/addons/addons.go
ereslibre: Let me check, because I think we want to close the `stdin` before executing the command (before the `CombinedOutput()` call), and the `defer` would be executed after that.

internal/pkg/skuba/addons/addons.go
mssola: Ah, missed that part. Sorry...

internal/pkg/skuba/addons/addons.go
mssola: In that case, you need to also add `stdin.Close()` before the `return err` above :smile:

ereslibre: Updated to not need a `StdinPipe` but pass the input on `cmd.Stdin` instead :)

internal/pkg/skuba/skuba/configmap.go
```diff
@@ -0,0 +1,85 @@
...
+const (
+	ConfigMapName             = "skuba-config"
+	SkubaConfigurationKeyName = "SkubaConfiguration"
+)
+
+type SkubaConfiguration struct {
+	AddonsVersion kubernetes.AddonsVersion
+}
+
+func GetSkubaConfiguration() (*SkubaConfiguration, error) {
```
jenting: Naming as GetConfiguration is better.
Then use it, is `skuba.GetConfiguration`.

internal/pkg/skuba/skuba/configmap.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
+)
+
+const (
+	ConfigMapName             = "skuba-config"
+	SkubaConfigurationKeyName = "SkubaConfiguration"
+)
+
+type SkubaConfiguration struct {
```
jenting: Naming as `Configuration` is better.

internal/pkg/skuba/skuba/configmap.go
```diff
@@ -0,0 +1,85 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
+)
+
+const (
+	ConfigMapName             = "skuba-config"
+	SkubaConfigurationKeyName = "SkubaConfiguration"
```
jenting: Naming as `ConfigurationKeyName` is better.

internal/pkg/skuba/skuba/configmap.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+		return nil, err
+	}
+	if err := yaml.Unmarshal([]byte(configMap.Data[SkubaConfigurationKeyName]), skubaConfiguration); err != nil {
+		return nil, errors.Wrap(err, "error unmarshaling SkubaConfiguration")
+	}
+	return skubaConfiguration, nil
+}
+
+func UpdateSkubaConfiguration(skubaConfiguration *SkubaConfiguration) error {
```
jenting: Same, prefer `UpdateConfiguration`

internal/pkg/skuba/kured/kured.go
```diff
@@ -18,28 +18,44 @@
...
+	"k8s.io/klog"
 )
 
-func GetKuredImage() string {
-	return images.GetGenericImage(skuba.ImageRepository, "kured",
-		kubernetes.CurrentAddonVersion(kubernetes.Kured))
-}
+const (
+	kuredDSName             = "kured"
+	kuredLockAnnotationJson = `{"metadata":{"annotations":{"weave.works/kured-node-lock":"'{\"nodeID\":\"manual\"}'"}}}`
```
jenting: Awesome :100:

internal/pkg/skuba/addons/addons.go
```diff
@@ -0,0 +1,281 @@
...
+func (addon Addon) manifestPath() string {
+	return filepath.Join(addon.addonPath(), fmt.Sprintf("%s.yaml", addon.addon))
+}
+
+func (addon Addon) Write(addonConfiguration AddonConfiguration) error {
+	addonManifest, err := addon.Render(addonConfiguration)
+	if err != nil {
+		return errors.Wrapf(err, "unable to render %s addon template", addon.addon)
+	}
+	os.MkdirAll(addon.addonPath(), 0700)
```
jenting: Add error check would better

internal/pkg/skuba/addons/addons.go
```diff
@@ -0,0 +1,281 @@
...
+var Addons = map[kubernetes.Addon]Addon{}
+
+type Addon struct {
+	addon         kubernetes.Addon
+	templater     addonTemplater
+	callbacks     addonCallbacks
+	addonPriority addonPriority
+}
+
+type addonCallbacks interface {
```
jenting: Awesome!

internal/pkg/skuba/addons/dex.go
```diff
@@ -0,0 +1,262 @@
...
+func init() {
+	registerAddon(kubernetes.Dex, renderDexTemplate, dexCallbacks{}, normalPriority)
+}
+
+func (renderContext renderContext) DexImage() string {
+	return images.GetGenericImage(skubaconstants.ImageRepository, "caasp-dex",
+		kubernetes.AddonVersionForClusterVersion(kubernetes.Dex, renderContext.config.ClusterVersion).Version)
+}
+
+func (renderContext renderContext) GangwayClientSecret() string {
```
jenting: Cool way, thanks for teach me. Now I know rendering could pass function, not just variable. :1st_place_medal:

internal/pkg/skuba/addons/addons.go
```diff
@@ -0,0 +1,281 @@
...
+			klog.Errorf("failed on %q addon AfterApply callback: %v", addon.addon, err)
+			return err
+		}
+	}
+	return updateSkubaConfigMapWithAddonVersion(addon.addon, addonConfiguration.ClusterVersion, skubaConfiguration)
+}
+
+func addonVersionLower(version1 *kubernetes.AddonVersion, version2 *kubernetes.AddonVersion) bool {
+	// If we don't have a version to compare to, assume it's not lower
+	if version2 == nil {
```
jenting: Should check on both right?

internal/pkg/skuba/skuba/configmap.go
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package skuba
```
jenting: Have any other package naming for this? I saw `internal/pkg/skuba/skuba`

internal/pkg/skuba/skuba/configmap.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func GetSkubaConfiguration() (*SkubaConfiguration, error) {
+	clientSet, err := kubernetes.GetAdminClientSet()
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting client set")
+	}
+	skubaConfiguration := &SkubaConfiguration{}
+	configMap, err := clientSet.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(ConfigMapName, metav1.GetOptions{})
+	if err != nil {
```
jenting: There is a function `DoesResourceExistWithError`

internal/pkg/skuba/kured/kured.go
```diff
@@ -18,28 +18,44 @@
...
+)
 
-func KuredLockExists() (bool, error) {
-	clientSet, err := kubernetes.GetAdminClientSet()
-	if err != nil {
-		return false, errors.Wrap(err, "unable to get admin client set")
-	}
-	kuredDaemonSet, err := clientSet.AppsV1().DaemonSets(metav1.NamespaceSystem).Get("kured", metav1.GetOptions{})
+func LockExists(client clientset.Interface) (bool, error) {
```
jenting: Better name to `IsLocked` or `IsLockExists`

ereslibre:
@jenting Thank you for your feedback! I'll address most (if not all) your comments in a follow-up PR.

PR #633, opened 2019-08-22
Added ability to skip deployment steps

tdaines42:
## Why is this PR needed?

Using the `--skip-setup` arg when running tests you can now skip different deployment
 stages during setup.

Previously the test would setup and teardown a cluster each time.

Fixes https://github.com/SUSE/avant-garde/issues/789

## What does this PR do?

Adds the `--skip-setup` arg

I manually tested all the different choices

# Merge restrictions after RC

(Please do not edit this)

        3 approvals:
            1 developer
            1 QA

### DIFF
--- a/ci/infra/testrunner/testrunner.py
+++ b/ci/infra/testrunner/testrunner.py
@@ -170,7 +170,12 @@ def main():
                            help="only list tests to be executed")
     test_args.add_argument("-v", "--verbose", dest="verbose", action="store_true", default=False,
                            help="show all output")
-    test_args.add_argument("--skip-setup", help="setup step to skip")
+    test_args.add_argument("--skip-setup",
+                           choices=['provisioned', 'bootstrapped', 'deployed'],
+                           help="Skip the given setup step.\n"
+                                "'provisioned' For when you have already provisioned the nodes.\n"
+                                "'bootstrapped' For when you have already bootstrapped the cluster.\n"
+                                "'deployed' For when you already have a fully deployed cluster.")
     cmd_test = commands.add_parser(
         "test", parents=[test_args], help="execute tests")
     cmd_test.set_defaults(func=test)
--- a/ci/infra/testrunner/testrunner.py
+++ b/ci/infra/testrunner/testrunner.py
@@ -74,7 +74,8 @@ def node_upgrade(options):
 
 def test(options):
     TestDriver(options.conf, options.platform).run(test_suite=options.test_suite, test=options.test,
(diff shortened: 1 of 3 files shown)


ci/infra/testrunner/testrunner.py
```diff
@@ -74,7 +74,8 @@ def node_upgrade(options):
 
 def test(options):
     TestDriver(options.conf, options.platform).run(test_suite=options.test_suite, test=options.test,
-                                                   verbose=options.verbose, collect=options.collect)
+                                                   verbose=options.verbose, collect=options.collect,
+                                                   skip_setup=options.skip_setup)
```
dannysauer: This change being 2 lines is why I really prefer the closing brace/paren to be on a line by itself. ;)

tdaines42: I really dislike trailing commas not sure why 😂

dannysauer: Probably because they often break lesser languages. :p

stefsuse: Check it out, never forget a comma again! ;)
```
run(test_suite=options.test_suite, test=options.test
                                                 , verbose=options.verbose
                                                 , collect=options.collect
                                                 , verbose=options.verbose
                                                 , collect=options.collect
                                                 , skip_setup=options.skip_setup
                                                 )
```

dannysauer: Augh, my eyes! :p

ci/infra/testrunner/testrunner.py
```diff
@@ -169,6 +170,7 @@ def main():
                            help="only list tests to be executed")
     test_args.add_argument("-v", "--verbose", dest="verbose", action="store_true", default=False,
                            help="show all output")
+    test_args.add_argument("--skip-setup", help="setup step to skip")
```
pablochacin: shouldn't we add here the options for this argument so that it is documented and is not a free text?

ci/infra/testrunner/tests/conftest.py
```diff
@@ -11,9 +11,18 @@ def pytest_addoption(parser):
     """
     parser.addoption("--vars", action="store",help="vars yaml" )
     parser.addoption("--platform", action="store",help="target platform" )
+    parser.addoption("--skip-setup",
+                     choices=['provisioned', 'bootstrapped', 'deployed'],
```
pablochacin: I would move this to the test runner so that this is visible to the testrunner's help

PR #634, opened 2019-08-22
Added reboot node tests

tdaines42:
## Why is this PR needed?

We want to test that rebooting a node doesn't break the cluster

Fixes https://github.com/SUSE/avant-garde/issues/264

## What does this PR do?

Adds tests for the different node types that reboots them then waits to see if they come back up and are ready again.

# Merge restrictions after RC

(Please do not edit this)

        3 approvals:
            1 developer
            1 QA

### DIFF
--- a/ci/infra/testrunner/tests/test_node_reboot.py
+++ b/ci/infra/testrunner/tests/test_node_reboot.py
@@ -1,0 +1,71 @@
+import logging
+import json
+import time
+
+import pytest
+from timeout_decorator import timeout
+
+logger = logging.getLogger("testrunner")
+
+
+@timeout(120)
+def wait_for_node_reboot(node_type, node_number, platform):
+    while True:
+        try:
+            platform.ssh_run(node_type, node_number, 'echo hello')
+        except Exception as ex:
+            logger.debug(ex)
+        else:
+            break
+
+        time.sleep(5)
+
+
+@timeout(120)
+def check_nodes_ready(kubectl):
+    while True:
+        try:
+            kubectl.run_kubectl('wait --for=condition=Ready node --all')
+        except Exception as ex:
+            logger.debug(ex)
+        else:
+            break
+
+        time.sleep(5)
+
+
+@timeout(120)
+def check_pods_ready(kubectl):
+    while True:
+        try:
+            pods = json.loads(kubectl.run_kubectl('get pods --all-namespaces -o json'))['items']
+
+            for pod in pods:
+                pod_status = pod['status']['phase']
+                assert pod_status in ['Running', 'Completed'], f'Pod {pod["metadata"]["name"]} status {pod_status} != Running or Completed'
+        except Exception as ex:
+            logger.debug(ex)
+        else:
+            break
+
+        time.sleep(5)
+
+
+@pytest.mark.parametrize('node_type,node_number', [('master', 1), ('worker', 0)])
+def test_hard_reboot(deployment, platform, skuba, kubectl, node_type, node_number):
+    assert skuba.num_of_nodes(node_type) > node_number
+
+    logger.info('Wait for all the nodes to be ready')
+    kubectl.run_kubectl('wait --for=condition=Ready node --all --timeout=5m')
+
+    logger.info(f'Rebooting {node_type} {node_number}')
+
+    platform.ssh_run(node_type, node_number, 'sudo reboot &')
+
+    # Give the reboot a bit to start
+    logger.info(f'Waiting 30s...')
+    time.sleep(30)
+
+    wait_for_node_reboot(node_type, node_number, platform)
+    check_nodes_ready(kubectl)
+    check_pods_ready(kubectl)
--- a/ci/infra/testrunner/tests/driver.py
+++ b/ci/infra/testrunner/tests/driver.py
@@ -10,6 +10,8 @@ class PyTestOpts:
 
     SHOW_OUTPUT = "-s"
 
+    VERBOSE = "-v"
+
     COLLECT_TESTS = "--collect-only"
 
     JUNIT = f"--junitxml={TESTRUNNER_DIR}/results.xml"
@@ -33,6 +35,7 @@ class TestDriver:
 
         if verbose:
             opts.append(PyTestOpts.SHOW_OUTPUT)
+            opts.append(PyTestOpts.VERBOSE)
 
         if collect:
             opts.append(PyTestOpts.COLLECT_TESTS)


ci/infra/testrunner/tests/test_node_reboot.py
```diff
@@ -0,0 +1,37 @@
...
+    assert skuba.num_of_nodes('master') > 1
+
+    logger.info('Wait for all the nodes to be ready')
+    kubectl.run_kubectl('wait --for=condition=Ready node --all --timeout=5m')
+
+    logger.info('Rebooting master 1')
+    platform.ssh_run('master', 1, 'sudo reboot')
+
+    # Give the reboot a bit to start
+    time.sleep(5)
```
pablochacin: I thin using a fixed time will introduce too much flakynes. We could use the same approach than it used in terraform and retry:
```
for retry in range(10):
     try: 
           platform.ssh_run( 'master', 1 , '/usr/bin/true'):
           break   
     except:
           time.sleep(5)
```

pablochacin: Maybe this is not needed as we are waiting for the kubectl as long as we have more than one master ...

ci/infra/testrunner/tests/test_node_reboot.py
```diff
@@ -0,0 +1,37 @@
...
+
+    kubectl.run_kubectl('wait --for=condition=Ready node --all --timeout=5m')
```
kravciak: This condition does not verify anything, it takes some time until node becomes `NotReady`.
```bash
# ssh sles@10.86.2.99 'sudo reboot'

# kubectl wait --for=condition=Ready node --all
node/caasp-master-kravciak-0 condition met
node/caasp-worker-kravciak-0 condition met # rebooted

# ssh 10.86.2.99
ssh: connect to host 10.86.2.99 port 22: Connection refused
```

tdaines42: Yeah I haven't had a chance to run the test yet just ran the steps locally. That's why this is labeled wip

MaximilianMeister: Also be aware that the `--all` flag doesn't work yet as we're still on kubectl 1.13.2 on the jenkins workers, but i'll try to redeploy them today

MaximilianMeister:
can we name the tests `test_hardreboot_master` and so on? just thinking we might want to add a controlled reboot test via kured at some point, so we have some clarity in the naming

ci/infra/testrunner/tests/test_node_reboot.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Give the reboot a bit to start
+    time.sleep(5)
+
+    start_time = time.time()
+    run_time = 0
+    while run_time < 300:
+        try:
+            platform.ssh_run('master', 1, 'echo hello')
+            kubectl.run_kubectl('wait --for=condition=Ready node --all')
```
kravciak: When you reboot a node it won't go into `NotReady` state that fast. This test will always pass (if ssh starts).
I had this in mind for testing reboot, but I don't have code ready yet (did not test it):
```
# we should somehow check that nodes are really rebooting
1. reboot node && sleep 30

# first problem is that nodes won't go to NotReady during reboot (it's too fast)
# second problem is that pods show error and start restarting only !AFTER! ssh is back
2. wait until ssh comes back && sleep 30

# verify that reboot did not break things
3. check all pods are running
```

ci/infra/testrunner/tests/test_node_reboot.py
```diff
@@ -0,0 +1,71 @@
...
+
+@timeout(120)
+def check_pods_ready(kubectl):
+    while True:
+        try:
+            pods = json.loads(kubectl.run_kubectl('get pods --all-namespaces -o json'))['items']
+
+            for pod in pods:
+                pod_status = pod['status']['phase']
+                assert pod_status == 'Running', f'Pod {pod["metadata"]["name"]} status {pod_status} != Running'
```
kravciak: Not all pods have to be running, some can be completed - I think this will always fail on cilium-operator, or not?
Btw there is kubectl wait for pods too, but I am not sure if it's reliable.

ci/infra/testrunner/tests/test_node_reboot.py
pablochacin: As for checking the sss is back you can use the approach I suggested before, similar to the one used in tf during deployment

ci/infra/testrunner/tests/test_node_reboot.py
```diff
@@ -0,0 +1,71 @@
...
+        except Exception as ex:
+            logger.debug(ex)
+        else:
+            break
+
+        time.sleep(5)
+
+
```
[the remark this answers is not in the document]
pablochacin: This is a great suggestion. This has been asked by some developers for other tests. I think we could add this to the development library so that it can be used by all tests.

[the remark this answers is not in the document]
MaximilianMeister: yes this would be very helpful for some scenarios in the e2e tests, it would be great if we could have a generic repeater function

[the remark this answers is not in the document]
pablochacin: It should only require to move these two functions to either ' testconf` file or create another ' testlib.py'  file.

ci/infra/testrunner/tests/test_node_reboot.py
```diff
@@ -0,0 +1,71 @@
...
+logger = logging.getLogger("testrunner")
+
+
+@timeout(120)
+def wait_for_node_reboot(node_type, node_number, platform):
+    while True:
+        try:
+            platform.ssh_run(node_type, node_number, 'echo hello')
+        except Exception as ex:
+            logger.debug(ex)
```
pablochacin: We are logging here errors that are expected to happen as we are waiting for the node to reboot. That may not be necessary.

ci/infra/testrunner/tests/test_node_reboot.py
```diff
@@ -0,0 +1,71 @@
...
+        else:
+            break
+
+        time.sleep(5)
+
+
+@timeout(120)
+def check_nodes_ready(kubectl):
```
pablochacin: I think this function can be useful for other tests. I would consider to make it part of the test library maybe in a new `testlib.py` file

ci/infra/testrunner/tests/test_node_reboot.py
```diff
@@ -0,0 +1,71 @@
...
+import time
+
+import pytest
+from timeout_decorator import timeout
+
+logger = logging.getLogger("testrunner")
+
+
+@timeout(120)
+def wait_for_node_reboot(node_type, node_number, platform):
```
pablochacin: I think the logic for rebooting a node (and waiting for it to come back alive) could be moved to the platform to make it reusable.

PR #640, opened 2019-08-27
Add cluster images command

sigsteve:
## Why is this PR needed?
This solves the question of what images are necessary for a deployment, when internet access is unavailable.

## What does this PR do?

This PR implements a 'skuba cluster images' command that displays the images, versions and locations of all containers skuba will need for a cluster. We can then take this information and populate a local registry without need for internet access from the CaaSP nodes directly.

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions after RC

(Please do not edit this)

        3 approvals:
            1 developer
            1 QA


pkg/skuba/actions/cluster/images/images.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"fmt"
+	"k8s.io/kubernetes/cmd/kubeadm/app/images"
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
+	"github.com/SUSE/skuba/pkg/skuba"
+)
+
+// Print out list of images that will be pulled
+// This can be used as input to skopeo for mirroring in air-gapped scenarios
+func Images() error {
```
MaximilianMeister: I'm not fully sure if that's compatible with what we do in [`cluster status`](https://github.com/SUSE/skuba/blob/master/pkg/skuba/actions/cluster/status/status.go#L47-L56) but adding column titles would make it look better

also I would put all image names into an array/map, so we can loop through them and don't repeat the `fmt.Println` so much in the code

pkg/skuba/actions/cluster/images/images.go
```diff
@@ -0,0 +1,56 @@
...
+	for _, ver := range kubernetes.AvailableVersions() {
+		fmt.Println(ver,images.GetGenericImage(skuba.ImageRepository, "pause",
+			kubernetes.ComponentVersionForClusterVersion(kubernetes.Pause,ver)))
+		fmt.Println(ver,images.GetGenericImage(skuba.ImageRepository, "hyperkube",
+			kubernetes.ComponentVersionForClusterVersion(kubernetes.Hyperkube,ver)))
+		fmt.Println(ver,images.GetGenericImage(skuba.ImageRepository, "etcd",
+			kubernetes.ComponentVersionForClusterVersion(kubernetes.Etcd,ver)))
+		fmt.Println(ver,images.GetGenericImage(skuba.ImageRepository, "coredns",
+			kubernetes.ComponentVersionForClusterVersion(kubernetes.CoreDNS,ver)))
+		fmt.Println(ver,images.GetGenericImage(skuba.ImageRepository, "kured",
```
ereslibre: Addons can now be iterated over.

internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -163,7 +163,11 @@ func CurrentComponentVersion(component Component) string {
 }
 
 func CurrentAddonVersion(addon Addon) string {
-	currentKubernetesVersion := Versions[LatestVersion().String()]
+	return AddonVersionForClusterVersion(addon, LatestVersion())
+}
+
+func AddonVersionForClusterVersion(addon Addon, clusterVersion *version.Version) string {
+  currentKubernetesVersion := Versions[clusterVersion.String()]
```
ereslibre: These changes can be dropped here in `versions.go`, now you can iterate over addons and use `AddonVersionForClusterVersion`.

internal/app/skuba/cluster.go
```diff
@@ -34,6 +34,7 @@ func NewClusterCmd() *cobra.Command {
 		cluster.NewInitCmd(),
 		cluster.NewStatusCmd(),
 		cluster.NewUpgradeCmd(),
+		cluster.NewImagesCmd(),
```
ereslibre: We should document this command here: https://github.com/SUSE/skuba/tree/master/docs/man, creating its own documentation man page, and also changing https://github.com/SUSE/skuba/blob/master/docs/man/skuba.1.md `See Also` section.

pkg/skuba/actions/cluster/images/images.go
sigsteve: I have updated the output to match that of the status command and now iterate over addons and components.

internal/pkg/skuba/kubernetes/versions.go
sigsteve: I've removed the function and use the new one to iterate over addons. Thank you!

internal/app/skuba/cluster.go
sigsteve: I have created a man page and updated skuba.1.md.

ereslibre:
I would prefer if we add this command on `development` build only first, would that be fine?

sigsteve:
Thank you. I couldn't see the lint results, but now I've made the suggested changes.

MaximilianMeister:
> I would prefer if we add this command on `development` build only first, would that be fine?

@ereslibre Depends, can you explain the reasoning behind keeping it out of production builds? At least I can't see any reason to hide this in production.

sigsteve:
> I would prefer if we add this command on `development` build only first, would that be fine?

Releasing to development first is fine. We won't be able to test it though, as we lack the access to pull from the registry.
What would the schedule be for having this deployed into the release build?
We have documentation for air gap that makes use of this and we can align its release with the schedule.

ereslibre:
> @ereslibre Depends, can you explain the reasoning behind keeping it out of production builds? At least I can't see any reason to hide this in production.

The main problem with the current approach is the scope of this new command. For whom is it?

If it's for the final user, then I see a little misleading that we are printing all versions of Kubernetes when you can only deploy the latest one (`--kubernetes-version` is only exposed in the `development` build). This command should then probably take a `--kubernetes-version` as well in development mode and print only the images related to that version. If none provided, in `development` we could print all of them, and on `staging` and `release` only print the latest one.

I think there are some details to polish on the scope of this command, that's why I'd prefer to not expose it directly right now in the `release` build until we figure out the use cases.

pkg/skuba/actions/cluster/images/images.go
```diff
@@ -0,0 +1,59 @@
...
+func Images() error {
+	components := map[kubernetes.Component]string{
+		kubernetes.Hyperkube: "hyperkube",
+		kubernetes.Etcd:      "etcd",
+		kubernetes.CoreDNS:   "coredns",
+		kubernetes.Pause:     "pause",
+		kubernetes.Tooling:   "tooling",
+	}
+
+	addons := map[kubernetes.Addon]string{
```
ereslibre: This map should not be necessary.

pkg/skuba/actions/cluster/images/images.go
```diff
@@ -0,0 +1,59 @@
...
+		kubernetes.PSP:     "psp",
+	}
+
+	fmt.Printf("VERSION    IMAGE\n")
+	for _, cv := range kubernetes.AvailableVersions() {
+		for component, componentName := range components {
+			fmt.Printf("%-10v %v\n", cv, images.GetGenericImage(skuba.ImageRepository, componentName,
+				kubernetes.ComponentVersionForClusterVersion(component, cv)))
+		}
+		for addon, addonName := range addons {
```
ereslibre:
```suggestion
		for addonName, addonVersion := range Versions[cv.String()].AddonsVersion {
```

sigsteve: Ok. I've cleaned this up and now use addonsVersion instead of a separate map.

sigsteve:
@ereslibre 

> The main problem with the current approach is the scope of this new command. For whom is it?

This command is intended for anyone deploying in an air-gapped environment, or wanting to localize their images. We are doing deployments/demos in environments without internet access. This feature allows us a simple way to find the right images to pull and provide the user a simple solution.

> If it's for the final user, then I see a little misleading that we are printing all versions of Kubernetes when you can only deploy the latest one (`--kubernetes-version` is only exposed in the `development` build).

I didn't realize we weren't going to propagate this option to release.

The thought around having multiple versions, would be to allow a user to pull images/deploy an older version, should they have to rebuild or extend a cluster that's not at the latest version.

If this isn't going to be supported, I can change the output to only include LatestVersion, as we will prevent confusion.

> I think there are some details to polish on the scope of this command, that's why I'd prefer to not expose it directly right now in the `release` build until we figure out the use cases.

We'd be happy to get together and provide feedback on what's been driving our use case here.

jenting:
Just a little picky, repeated same version makes the output hard to read.

ereslibre:
> Just a little picky, repeated same version makes the output hard to read.

I agree, however without it being json or yaml output makes it easy to do something like `skuba cluster images | grep "1.15.0" | awk '{print $2}'`.

PR #642, opened 2019-08-27
Fix setup teardow

pablochacin:
## Why is this PR needed?

Fixes # https://github.com/SUSE/avant-garde/issues/810

appear in the changelog.

## What does this PR do?

Ensure finalizer function are registered before the platform
provisioning is started, to allow cleanup of failed provisionings.

## Info for QA

This PR only affects tests.

### DIFF
--- a/ci/infra/testrunner/tests/README.md
+++ b/ci/infra/testrunner/tests/README.md
@@ -35,13 +35,25 @@ The example below shows a fixture that provides a bootstrapped cluster. It also 
 ```
 @pytest.fixture
 def setup(request, platform, skuba):
-    platform.provision()
     def cleanup():
         platform.cleanup()
     request.addfinalizer(cleanup)
 
+    platform.provision()
     skuba.cluster_init()
     skuba.node_bootstrap()
+```
+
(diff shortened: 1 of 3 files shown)


ci/infra/testrunner/tests/README.md
```diff
@@ -35,13 +35,25 @@ The example below shows a fixture that provides a bootstrapped cluster. It also
...
     def cleanup():
         platform.cleanup()
     request.addfinalizer(cleanup)
 
+    platform.provision()
+    skuba.cluster_init()
+    skuba.node_bootstrap()
+```
+
+Note: pytest also allow a more idiomatic way of defining teardown logic in fixtures by using python's `yield` statement instead of registering a finalizer, as whown in the code bellow. However, finalizer functions have the advantage that they will always be called regardless if the fixture setup code raises an exception, provided they are registered before the exception occurs. Therefore, testrunner encourages using finalizer functions.
```
MaximilianMeister: `as shown in the code below`

pablochacin:
PR closed by mistake. New PR opened that supersedes it https://github.com/SUSE/skuba/pull/660

PR #644, opened 2019-08-27
Deployment test now cleans up after itself

tdaines42:
## Why is this PR needed?

The deployments test was not cleaning up after itself

Fixes #

## What does this PR do?

Removes the deployment and the service at the end of the test

# Merge restrictions after RC

(Please do not edit this)

        3 approvals:
            1 developer
            1 QA

### DIFF
--- a/ci/infra/testrunner/tests/test_deployments.py
+++ b/ci/infra/testrunner/tests/test_deployments.py
@@ -1,8 +1,5 @@
 import pytest
-import json
 import requests
-
-from kubectl import Kubectl
 
 
 def test_nginx_deployment(deployment, kubectl):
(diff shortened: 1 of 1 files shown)


pablochacin:
@stefsuse the problem is that this test is not marked as disruptive and therefore it should be possible to run another test re-using the same cluster. I think ensuring proper cleanup is required, not optional.

ci/infra/testrunner/tests/test_deployments.py
```diff
@@ -24,3 +21,13 @@ def test_nginx_deployment(deployment, kubectl):
...
+
+    # Cleanup
+    kubectl.run_kubectl("delete --wait --timeout=60s service/nginx")
+    kubectl.run_kubectl("delete --wait --timeout=60s deployments/nginx")
+
+    with pytest.raises(Exception):
+        kubectl.run_kubectl("get service/nginx")
+
+    with pytest.raises(Exception):
+        kubectl.run_kubectl("get deployments/nginx")
```
pablochacin: I would prefer to check directly if the service/deployment is present and not rely on exceptions (which may have a different cause) 

```
services = kubectl get services  -o jsonpath=' { .items[*].metadata.name }'.`
assert 'nginx not in services`
```

PR #645, opened 2019-08-27
Print stderr when running shell commands

pablochacin:
## Why is this PR needed?

Some commands, like skuba, print usefull debug level
information to stderr instead of stdout. 

## What does this PR do?

When running those commands, testrunner should print stderr
regardless there was an error executing the command or not.


jordimassaguerpla:
Anyone can now merge on master if you have the needed approvals given we have a specific branch for CaaSPv4 GMC release.

PR #651, opened 2019-08-28
Easy recovery from bootstrap failures bsc#1148524

flavio:
## Why is this PR needed?

The bootstrap of the first master node is made by two macro phases:

  1. Bootstrapping or core components: operations like installing missing
     packages, enabling system services, running kubeadm
  2. Define core resources on top of the kubernetes cluster

At the end of phase 1 we have a working control plane, we have even a
`kubeconfig` file. However core add-ons like network (cilium), coredns,
kured and similar are not yet deployed.

Unfortunately there's no easy way to recover from a failure that happens at
the end of phase 2.

It occurred to me several times that the AWS load balancer would have problems
seeing the "just started" api-server as active. That would cause a failure at
the early beginning of phase 2.

Recovering from this failure requires a lot of manual clean up, both inside of
the cluster folder (removing CA and other files) and on the remote node.

We can allow an easy recovery from these situation: we can keep applying
phase 2 as many times as we want because the deployment of kubernetes
resources (which is the only kind of operation done during phase 2) is
idempotent.

Note well: this issue happened to me with AWS, but this kind of glitches can
happen also with other deployments.

fixes bsc#1148524

## What does this PR do?

This PR changes the code that does the initial bootstrap by splitting the code in two parts:
  * Setup of core components: install missing packages, enable services, run `kubadm`
  * Deployment of core add-ons on top of the freshly deployed control plane node: stuff like cilium, kured, coredns, PSP

Nothing special to report.

## Info for QA

Validation should ensure:

  * Bootstrap works as expected when no glitches are introduced between phase 1 and 2.
  * Bootstrap can be resumed when something goes wrong at the beginning of phase 2.

The tricky part is finding a way to introduce a glitch, I see two possible solutions:
  * Alter the cilium deployment manually, introduce some mistake. See the bootstrap fail, fix the error you introduced, run the same command again -> everything should work fine
  * Stop the load balancer or instruct it to point to another master node, watch the deployment fail, fix the load balancer, try again -> everything should work

I'm a bit skeptical about the 2nd approach, I fear `kubeadm` tries to go through the load balancer at some point... but I might be wrong. @ereslibre what do you think?

Before this patch an user cannot recover from a failure during phase 2 without a lot of manual work.

Validation has been described above.

## Docs

No need to update the docs I guess.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/pkg/skuba/actions/node/bootstrap/bootstrap.go
+++ b/pkg/skuba/actions/node/bootstrap/bootstrap.go
@@ -43,10 +43,13 @@ import (
 // FIXME: being this a part of the go API accept the toplevel directory instead
 //        of using the PWD
 func Bootstrap(bootstrapConfiguration deployments.BootstrapConfiguration, target *deployments.Target) error {
+	coreBootstrapDone := false
+
 	if clientSet, err := kubernetes.GetAdminClientSet(); err == nil {
 		_, err := clientSet.ServerVersion()
 		if err == nil {
-			return errors.New("cluster is already bootstrapped")
+			fmt.Printf("[bootstrap] node %q has already the core components bootstrapped\n", target.Target)
+			coreBootstrapDone = true
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


ereslibre:
This needs a rebase, the addons logic got in.

flavio:
@ereslibre: I rebased against master

pablochacin:
@flavio maybe manually introducing an error in some of the addon yaml files (for instance, changing kured's command) before bootstrapping the cluster may be sufficient, to force the failure, doesn't it?

PR #652, opened 2019-08-28
Fix tag detection

ereslibre:
## Why is this PR needed?
The previous approach for detecting if we were in a tag was not
supporting annotated tags. With this method we are supporting both
lightweight and annotated tags.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,18 +22,18 @@ GOBINPATH     := $(shell $(GO) env GOPATH)/bin
 COMMIT        := $(shell git rev-parse HEAD)
 BUILD_DATE    := $(shell date +%Y%m%d)
 # TAG can be provided as an envvar (provided in the .spec file)
-TAG           ?= $(shell git show-ref --tags | grep $(COMMIT) | awk '{print $2}' | cut -d/ -f3)
-# ANNOTATED_TAG can be provided as an envvar (provided in the .spec file)
-ANNOTATED_TAG ?= $(shell git describe --tag)
-# VERSION is inferred from TAG
+TAG           ?= $(shell git describe --tags --exact-match HEAD 2> /dev/null)
+# CLOSEST_TAG can be provided as an envvar (provided in the .spec file)
+CLOSEST_TAG   ?= $(shell git describe --tags)
+# VERSION is inferred from CLOSEST_TAG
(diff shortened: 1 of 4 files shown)


jenting:
I make a lightweight tag `git tag test123` and run `make`. Both _skuba.Tag_ and _skuba.AnnotatedTag_ are `test123`. Is this an expected behavior?

ereslibre:
> Is this an expected behavior?

Yes, both with a lightweight and annotated tag, when running `make`, `skuba.Tag` and `skuba.AnnotatedTag` should be the tag contents.

When we are not in a tag, when running `make`, `skuba.Tag` should be empty, and `skuba.AnnotatedTag` should contain something of the form `v1.0.0-34-ga75eaaaf` (last tag-number of commits since last tag-HEAD).

jenting:
> > Is this an expected behavior?
> 
> Yes, both with a lightweight and annotated tag, when running `make`, `skuba.Tag` and `skuba.AnnotatedTag` should be the tag contents (in your case `test123`).
> 
> When we are not in a tag, when running `make`, `skuba.Tag` should be empty, and `skuba.AnnotatedTag` should contain something of the form `v1.0.0-34-ga75eaaaf` (last tag-number of commits since last tag-HEAD).

Hmmm, but `git tag test123` is a lightweight tag, right?

I am not sure why provides both lightweight and annotated tag in Makefile. Maybe we could restrict to use annotated tag only in GitHub setting.

ereslibre:
> Hmmm, but git tag test123 is a lightweight tag, right?

Yes, maybe the confusion comes from the `ANNOTATED_TAG` env name, this will have whatever `git describe --tags` return (it always return something), I didn't have a better name, but it certainly has no direct relationship with `git tag -a` kind of annotated tag.

The idea is that we always have `ANNOTATED_TAG` populated, and `TAG` only when we are in a lightweight or annotated tag.

ereslibre:
@jenting, I think it's more clear now. `ANNOTATED_TAG` was renamed to `CLOSEST_TAG`.

Now we always pass the `skuba.ClosestTag` and `skuba.Version` (inferred from `CLOSEST_TAG`) to the go build environment.

`skuba.Tag` is only passed to the go build environment if we are in a lightweight or annotated tag. With this change, we can always build a valid package with `make suse-package`, because `Version` will always be populated.

PR #653, opened 2019-08-28
Remove `ginkgo`

ereslibre:
## Why is this PR needed?

As part of the latest discussions on how to perform e2e tests,
remove ginkgo for e2e testing.

It can be added later if we want to perform integration testing with
it, or if we want to use the gomega matchers for any kind of tests.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


ereslibre:
Adding `wip` because the vmware pipeline was calling to ginkgo.

ereslibre:
> removing ci/tasks/rebase.py would break ci/Makefile:git_rebase_check

Ouch, good catch, will update.

jenting:
Where is the latest discussion? Any link?

ereslibre:
> Where is the latest discussion? Any reference link?

During the testrunner ad-hoc meeting and https://confluence.suse.com/display/SUSECaaSPlatform4/Testing+Tools+and+Responsibilities

ereslibre:
I apologize in advance for the verbosity of this message, and if I'm
stating obvious things, but I find it important that we are talking
about the same concepts. Please, if there's anything that I say that
is wrong, tell me.

> Integration and e2e tests are different things so not sure we should rename that

I'll use the classic acronyms that do not fit 1:1 with golang entirely.

| Type        | Also known as                                               | System Under Test (SUT)                                                  | Current Framework      |
|-------------|-------------------------------------------------------------|--------------------------------------------------------------------------|------------------------|
| Unit        |                                                             | Class Under Test (CUT), Object Under Test (OUT), Method Under Test (MUT) | golang's "testing"     |
| Integration | Narrow integration tests                                    | Module(s) Under Test                                                        | golang's "testing" (*) |
| End to End  | Broad integration tests, Functional tests, Acceptance tests | Application Under Test (AUT) or a major subsystem of it                  | testrunner/pytest      |

(*) I'm not aware of any integration tests in the product right now.

* Unit
  * Test small logic chunks and are very specific to the current
    module being tested.
  * They must not span other modules.
  * They are very fast.
  * They report very localized errors.

* Integration
  * Test the integration between two different modules work as
    expected.
  * They can span different modules, testing specific integration
    points.
  * It is very common for them to still set the SUT as one module,
    what usually leads to having Depended-on Component (DOC). DOCs
    are usually implemented in the integration test itself, with Test
    Doubles, Mocks or Fakes. They allow to focus on the SUT, and how
    it integrates with the mocked responses of the DOCs.
  * They should be fast, because DOCs are usually doubled, mocked, or
    faked.
  * They report very localized errors in general, but if real modules
    are used instead of doubles, mocks or fakes the error could be on
    any of the modules (SUT, or DOC).

* End to End
  * Test a big chunk of logic, that can comprise a number of modules
    and/or components.
  * Black box testing is a very common approach.
  * They report whether an operation failed or succeeded, but they
    don't point to the specific component that caused it, hence they
    are harder to debug.

So, from my point of view, our current testrunner/pytest are strict e2e tests
since they call to `skuba` in a black box fashion, and check if the outcome
is the desired one.

| Type                 | System Under Test (SUT) | Main Goal     | Current Framework   |
|----------------------|-------------------------|---------------|---------------------|
| Skuba e2e tests      | skuba                   | QA validation | testrunner/pytest   |
| Kubernetes e2e tests | Kubernetes              | Conformance   | testrunner/sonobuoy |


> Also have we decided that we're not using ginkgo for unit tests or does this not affect those?

Before this PR we had two different places to put e2e tests: `/test`
and `/ci/infra/testrunner/tests`, what lead to some confusion. The
goal of this PR is to remove `/test` (the e2e tests implemented with
ginkgo). As a result of this, the ginkgo dependency goes away because
nothing else in the product is depending on it.

This doesn't mean that ginkgo shouldn't be used, only that we are
cleaning it because it's a no longer required dependency at the
moment. If we want to reintroduce ginkgo or gomega in the future I'm
perfectly fine with it, maybe targeting unit or integration tests.

pablochacin:
I think the commend from @ereslibre explaining the different testing strategies should be migrated to a confluence part so that it outlives this PR.

ereslibre:
Merging because this PR does not impact the product in a direct way.

PR #658, opened 2019-08-29
Relax the upgrade e2e tests

MaximilianMeister:
## Why is this PR needed?

During the upgrade of a control plane we depend on a correctly
configured load balancer to handle request passing.

## What does this PR do?

Adding a function to check if all control planes have a
responding apiserver.

## Docs

PR to docs will follow


MaximilianMeister:
rebased

PR #660, opened 2019-08-29
Fix setup teardown

pablochacin:
## Why is this PR needed?

Fixes # https://github.com/SUSE/avant-garde/issues/810

appear in the changelog.

## What does this PR do?

Ensure finalizer function are registered before the platform
provisioning is started, to allow cleanup of failed provisionings.

## Info for reviwers

This PR supperseds https://github.com/SUSE/skuba/pull/642 closed by mistake. 

## Info for QA

This PR only affects tests.

### DIFF
--- a/ci/infra/testrunner/tests/README.md
+++ b/ci/infra/testrunner/tests/README.md
@@ -35,13 +35,25 @@ The example below shows a fixture that provides a bootstrapped cluster. It also 
 ```
 @pytest.fixture
 def setup(request, platform, skuba):
-    platform.provision()
     def cleanup():
         platform.cleanup()
     request.addfinalizer(cleanup)
 
+    platform.provision()
     skuba.cluster_init()
     skuba.node_bootstrap()
+```
+
+Note: pytest also allow a more idiomatic way of defining teardown logic in fixtures by using python's `yield` statement instead of registering a finalizer, as shown in the code below. However, finalizer functions have the advantage that they will always be called regardless if the fixture setup code raises an exception, provided they are registered before the exception occurs. Therefore, testrunner encourages using finalizer functions.
+
+```
+@pytest.fixture
+def setup(request, platform, skuba):
+    platform.provision()
+    skuba.cluster_init()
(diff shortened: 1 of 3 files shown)


pablochacin:
Apparently there's an incompatibility between the changes introduced in https://github.com/SUSE/skuba/pull/639 and the automatic tear down of resources in fixtures. 

From what I've seen now a cleanup is executed as part of the ci. The assumption for tests is that the cleanup is done by the tests.  @tdaines42

MaximilianMeister:
> Apparently there's an incompatibility between the changes introduced in #639 and the automatic tear down of resources in fixtures.

can you explain more details? i can't see any relation to the other PR. i ran into a failed test -> no cleanup again just today running testrunner on my machine, so i believe we still need this change

also can you rebase again? looks like there's an extra commit now

pablochacin:
> can you explain more details? i can't see any relation to the other PR.

It seems there's a divergence in the way the resource life-cycle is managed. I opened a discussion here: https://github.com/SUSE/avant-garde/issues/817

LT;DR the recent changes are moving the logic for cluster setup to the ci and testrunner instead of relaying on the pythest fixture mechanism (which I personally advocate) therefore, I'm no longer sure this PR is even valid.

MaximilianMeister:
> It seems there's a divergence in the way the resource life-cycle is managed. I opened a discussion here: [SUSE/avant-garde#817](https://github.com/SUSE/avant-garde/issues/817)
> 

thanks, this should indeed be separated from what we try to fix here 

> LT;DR the recent changes are moving the logic for cluster setup to the ci and testrunner instead of relaying on the pythest fixture mechanism (which I personally advocate) therefore, I'm no longer sure this PR is even valid.

at least registering the cleanup finalizer before the deployment is still needed for the time being, can we have that change now?

pablochacin:
> can we have that change now?
Sure, it's not like I have not been trying to move this forward, PR tests were failing and I was unable to find the reason. Rebasing magically solved it.

PR #661, opened 2019-08-29
OpenStack security group: fix assignment to instances [bsc#1148711]

flavio:
## Why is this PR needed?

This PR fixes bsc#1148711 which currently is blocking the GMC release.

## What does this PR do?

Change we way we reference security groups inside of OpenStack's instance definitions. We have to rely on the security group name instead of its ID.
    
Referencing a security group by ID is confusing terraform when `terraform apply` is executed against an already (or partially) deployed cluster.

Without this patch terraform thinks the security group associated with the instances changed and tries to change the instance <-> security group association against a live instance. This is something not supported by OpenStack, which leads to an internal server error.
    
While doing this change I also spotted two different security groups had the same name. This has been fixed as well.

Nothing else

## Info for QA

The issue has been reported by QA. @thehejik worked with me on that.

Growing an OpenStack cluster, or doing a `terraform apply` against an already existing one will lead to a server error inside of OpenStack.

It's possible to grow clusters and run `terraform apply` against an already deployed cluster.

## Docs

No change needed.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/openstack/master-instance.tf
+++ b/ci/infra/openstack/master-instance.tf
@@ -66,8 +66,8 @@ resource "openstack_compute_instance_v2" "master" {
   }
 
   security_groups = [
-    "${openstack_networking_secgroup_v2.common.id}",
-    "${openstack_networking_secgroup_v2.master_nodes.id}",
+    "${openstack_networking_secgroup_v2.common.name}",
+    "${openstack_networking_secgroup_v2.master_nodes.name}",
   ]
 
   user_data = "${data.template_file.master-cloud-init.rendered}"
--- a/ci/infra/openstack/security-groups-load-balancer.tf
+++ b/ci/infra/openstack/security-groups-load-balancer.tf
@@ -1,5 +1,5 @@
 resource "openstack_networking_secgroup_v2" "load_balancer" {
-  name        = "${var.stack_name}-caasp_common_secgroup"
+  name        = "${var.stack_name}-caasp_lb_secgroup"
   description = "Common security group for CaaSP load balancer"
 }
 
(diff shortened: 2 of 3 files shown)


thehejik:
Doesn't work, now `terraform apply` complains on attempt to deploy even only 1:1 cluster:
```
Error: Error applying plan:

1 error(s) occurred:

* openstack_lb_loadbalancer_v2.lb: 1 error(s) occurred:

* openstack_lb_loadbalancer_v2.lb: Bad request with: [PUT https://engcloud.prv.suse.net:9696/v2.0/ports/a23ec3ad-54ec-4f45-a563-4cdc93c8b6cd], error message: {"NeutronError": {"message": "Invalid input for security_groups. Reason: 'th-gcc-bug-caasp_lb_secgroup' is not a valid UUID.", "type": "HTTPBadRequest", "detail": ""}}
```

flavio:
Note well: the test failed because our quota on openstack was exceeded.

thehejik:
The issue seems to be fixed by applying code from this PR (inc. 2nd commit). Also Dirk Mueller from #engcloud-help somehow fixed ECP (terraform was getting EOF when asking for network resources).

I was able to deploy 1:1 cluster by `terraform apply` and then let the cluster successfully grow to 3:3 by changing `masters` and `workers` values in `terraform.tfvars` file and reapplying `terraform apply`.

flavio:
> /lgtm I'm assuming that this is not needed for VMware? I looked but am not an expert on that platform.

This is not needed on VMware, they don't have security groups defined as far as I can see

PR #662, opened 2019-08-29
Remove MustParseSemantic wherever is not needed

mssola:
## Why is this PR needed?

There are two relevant functions when parsing Kubernetes versions: ParseSemantic and MustParseSemantic. They are identical except that the former returns an error and the latter panics on error.

Before this commit we were abusing of MustParseSemantic, and this in turn blurred some error logs that we have recently faced.

Fixes SUSE/avant-garde#818

## What does this PR do?

This commit changes MustParseSemantic with ParseSemantic wherever possible: whenever the version does not come strictly from the skuba source code.

Last but not least, in some functions it's hard to tell whether the given version comes from a safe place or not (e.g. the version is in a string format). In these cases I've kept the usage of MustParseSemantic and I've written a documentation comment explaining the possibility of the function panicking.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubeadm/configmap.go
+++ b/internal/pkg/skuba/kubeadm/configmap.go
@@ -54,7 +54,7 @@ func GetCurrentClusterVersion(client clientset.Interface) (*version.Version, err
 	if err != nil {
 		return nil, err
 	}
-	return version.MustParseSemantic(initCfg.KubernetesVersion), nil
+	return version.ParseSemantic(initCfg.KubernetesVersion)
 }
 
 // GetKubeadmApisVersion returns the api version to use in the kubeadm-init.conf
--- a/internal/pkg/skuba/addons/generic.go
+++ b/internal/pkg/skuba/addons/generic.go
@@ -21,6 +21,9 @@ import (
 	versionutil "k8s.io/apimachinery/pkg/util/version"
 )
 
+// KubernetesVersionAtLeast will panic if the given version is not a semantic
+// one. Thus, before calling this function, make sure that this version comes
+// from a safe place (i.e skuba source code).
 func (renderContext renderContext) KubernetesVersionAtLeast(version string) bool {
 	return renderContext.config.ClusterVersion.AtLeast(versionutil.MustParseSemantic(version))
 }
--- a/internal/pkg/skuba/deployments/ssh/kubernetes.go
+++ b/internal/pkg/skuba/deployments/ssh/kubernetes.go
@@ -66,14 +66,18 @@ func kubernetesInstallNodePattern(t *Target, data interface{}) error {
 		return errors.New("couldn't access kubernetes base OS configuration")
 	}
 
-	currentVersion := kubernetes.MajorMinorVersion(version.MustParseSemantic(kubernetesBaseOSConfiguration.CurrentVersion))
+	v, err := version.ParseSemantic(kubernetesBaseOSConfiguration.CurrentVersion)
+	if err != nil {
+		return err
+	}
+	currentVersion := kubernetes.MajorMinorVersion(v)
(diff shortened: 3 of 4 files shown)


internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -67,7 +68,11 @@ func (nviu NodeVersionInfoUpdate) IsUpdated() bool {
 
 func (nviu NodeVersionInfoUpdate) IsFirstControlPlaneNodeToBeUpgraded() bool {
 	isControlPlane := nviu.Current.IsControlPlane()
-	currentClusterVersion, _ := kubeadm.GetCurrentClusterVersion()
+	currentClusterVersion, err := kubeadm.GetCurrentClusterVersion()
+	if err != nil {
+		klog.Error(err.Error())
+		return false
```
ereslibre: I'd change the signature of this function to `(bool, error)` and bubble up the error here... I think with this current behavior we could hit weird logic errors. WDYT?

MaximilianMeister: it would likely give false logic results in https://github.com/SUSE/skuba/blob/master/pkg/skuba/actions/node/upgrade/apply.go

but I would rather change the function like this:

```go
func (nviu NodeVersionInfoUpdate) IsFirstControlPlaneNodeToBeUpgraded(currentClusterVersion *version.Version) bool {
	isControlPlane := nviu.Current.IsControlPlane()
	allControlPlanesMatchVersion, _ := kubernetes.AllControlPlanesMatchVersion(currentClusterVersion)
	matchesClusterVersion := (currentClusterVersion.String() == nviu.Current.KubeletVersion.String())

	return isControlPlane && allControlPlanesMatchVersion && matchesClusterVersion
}
```

and pass the `currentClusterVersion` which is [already instantiated during the upgrade apply](https://github.com/SUSE/skuba/blob/master/pkg/skuba/actions/node/upgrade/apply.go#L50)

internal/pkg/skuba/kubeadm/configmap.go
```diff
@@ -60,7 +60,7 @@ func GetCurrentClusterVersion() (*version.Version, error) {
 	if err != nil {
 		return nil, err
 	}
-	return version.MustParseSemantic(initCfg.KubernetesVersion), nil
+	return version.ParseSemantic(initCfg.KubernetesVersion)
```
jenting: nil missing

MaximilianMeister: `ParseSemantic` returns (version, err) so this is just passing through

internal/pkg/skuba/upgrade/node/versions.go
ereslibre: Also, looking at that snippet, isn't `matchesClusterVersion := (currentClusterVersion.String() == nviu` redudant? I think `AllControlPlanesMatchVersion` already checks that, right?

So this could be refactored as `return isControlPlane && allControlPlanesMatchVersion` if I'm not mistaken.

mssola: Done!

ereslibre:
Note, this PR and https://github.com/SUSE/skuba/pull/669 are going to conflict. @mssola, do you think we could keep the logic part of `IsFirstControlPlaneNodeToBeUpgraded()` as it is there? Reusing the variable as it's there. I think this function without any argument looks more clear (a version argument on that function IMHO looks a little weird in terms of semantics).

mssola:
Sure. Let's get #669 in and then I'll adapt this PR :+1:

c3y1huang:
Thanks. #669 merged :)

mssola:
@ereslibre review again :wink:

PR #665, opened 2019-08-29
updated to install recommended zypper packages

Klaven:
when installing the node package zypper will now also install the
recommended packages

## Why is this PR needed?

Fixes #https://github.com/SUSE/avant-garde/issues/728

appear in the changelog.

## What does this PR do?

adds the --recommends flag to zypper so that recommended packages get installed with it.

A simple test was added to ensure the packages in question on the issue where installed.

## Info for QA

Please review the test cases and see if there is anything you would like.

### Related info

NONE

use skuba bootstrap, then ask zypper if ceph-common is installed

use skuba bootstrap, then ask zypper if ceph-common is installed

## Docs

None

# Merge restrictions

(Please do not edit this) then why is it in the PR template?

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/kubernetes.go
+++ b/internal/pkg/skuba/deployments/ssh/kubernetes.go
@@ -73,7 +73,7 @@ func kubernetesInstallNodePattern(t *Target, data interface{}) error {
 		updatedVersion := kubernetes.MajorMinorVersion(version.MustParseSemantic(kubernetesBaseOSConfiguration.UpdatedVersion))
 		patternName = fmt.Sprintf("patterns-caasp-Node-%s-%s", currentVersion, updatedVersion)
 	}
-	_, _, err := t.ssh("zypper", "--non-interactive", "install", "--force", patternName)
+	_, _, err := t.ssh("zypper", "--non-interactive", "install", "--recommends", "--force", patternName)
 	return err
 }
 
--- a/ci/infra/testrunner/tests/test_installed_pkg.py
+++ b/ci/infra/testrunner/tests/test_installed_pkg.py
@@ -1,0 +1,24 @@
+import logging
+import time
+
+logger = logging.getLogger("testrunner")
+
+
+def test_installed_pkg(deployment, platform, skuba, kubectl):
+    assert skuba.num_of_nodes('master') > 0
+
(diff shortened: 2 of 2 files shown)


internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -73,7 +73,7 @@ func kubernetesInstallNodePattern(t *Target, data interface{}) error {
 		updatedVersion := kubernetes.MajorMinorVersion(version.MustParseSemantic(kubernetesBaseOSConfiguration.UpdatedVersion))
 		patternName = fmt.Sprintf("patterns-caasp-Node-%s-%s", currentVersion, updatedVersion)
 	}
-	_, _, err := t.ssh("zypper", "--non-interactive", "install", "--force", patternName)
+	_, _, err := t.ssh("zypper", "--non-interactive", "install", "--recommends", "--force", patternName)
```
ereslibre: Now that we are at it, do you think it would make sense to remove `--force` here? I know @davidcassany has recommended against using `--force` several times, maybe we could tackle this as well here, @davidcassany WDYT?

fgerling:
Will fix https://bugzilla.suse.com/show_bug.cgi?id=1148988

internal/pkg/skuba/deployments/ssh/kubernetes.go
Klaven: I would feel better about doing this in a separate PR. I think it warrants a larger talk about why we wanted it there in the first place.

ereslibre: I'm fine with another PR, why it was here was mainly because we started off without a pattern (with packages directly) and then we moved to the pattern. When we started using a pattern, @davidcassany has recommended several times to not use `--force`.

Klaven: I will create a new PR today to address this.

ereslibre:
After checking what really gets installed I'd consider to revert this change, please check https://github.com/SUSE/avant-garde/issues/728#issuecomment-527096311.

cc/ @jordimassaguerpla

PR #673, opened 2019-09-02
Fix make clean/distclean failed

jenting:
## Why is this PR needed?

Run `make clean` or `make distclean`, error message bumps out
```
can't load package: package github.com/SUSE/skuba: unknown import path "github.com/SUSE/skuba": cannot find module providing package github.com/SUSE/skuba
```

## What does this PR do?

Not use `-mod=vendor` flag when clean/disclean.

N/A

## Info for QA

### Related info

N/A

Run `make clean` or `make distclean`, error message bumps out

Run `make clean` or `make distclean`, error message bump out, no error message bumps out

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,12 +64,12 @@ install: go-version-check
 
 .PHONY: clean
 clean:
-	$(GO) clean -i
+	$(GO) clean -i ./...
 	$(RM) -f ./skuba
 
 .PHONY: distclean
 distclean: clean
-	$(GO) clean -i -cache -testcache -modcache
+	$(GO) clean -i -cache -testcache -modcache ./...
 
 .PHONY: staging
 staging:


Makefile
```diff
@@ -64,12 +64,12 @@ install: go-version-check
 
 .PHONY: clean
 clean:
-	$(GO) clean -i
+	go clean -i
```
innobead: This is not working when the project is outside GOPATH/src. Suggest to use `$(GO) clean -i ./...`, then it will work in both places.

Makefile
```diff
@@ -64,12 +64,12 @@ install: go-version-check
...
 	$(RM) -f ./skuba
 
 .PHONY: distclean
 distclean: clean
-	$(GO) clean -i -cache -testcache -modcache
+	go clean -i -cache -testcache -modcache
```
innobead: `$(GO) clean -i -cache -testcache -modcache ./...`

Makefile
jenting: 💯

Makefile
jenting: 💯

PR #678, opened 2019-09-03
testrunner: do not rely on shebang when running `pip3`

ereslibre:
## Why is this PR needed?

When running `pip3` directly, the shebang command can exceed the 128
limit very easily depending on the job name that is being run on CI.

To avoid this limit, call to `python3` directly, that will run the
`pip3` command, instead of relying on the `pip3` venv script, that
contains the shebang that can exceed 128 characters.

More information: https://github.com/pypa/pip/issues/1773

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/testrunner
+++ b/ci/infra/testrunner/testrunner
@@ -16,7 +16,9 @@ setup_python_env() {
     python3 -m venv $VENVDIR
   fi
   source ${VENVDIR}/bin/activate
-  pip3 install -r "${SDIR}/requirements.txt"
+  # Not calling to `pip3` directly because its shebang can exceed 128 characters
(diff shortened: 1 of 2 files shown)


ereslibre:
An alternative option is to run `pip` as a module (`python3 -m pip`) instead.

ereslibre:
Updated `pr_manager` as well, and used `python -m pip` instead on both.

ereslibre:
As an aside I find it a little weird to `activate` the venv, install requirements and `deactivate` right after that? When we run the script we are in no specific environment, but we installed the requirements in the `$VENVDIR` environment.

Am I misunderstanding how python does this behind the covers?

PR #679, opened 2019-09-03
removed force from options in node install

Klaven:
no longer force package install for the node package

## Why is this PR needed?

Per our discussions here https://github.com/SUSE/skuba/pull/665 and elsewhere, is not needed or wanted anymore to use the flag force.
Fixes #

## What does this PR do?

it removes the `--force` flag from the node install.

no.

## Info for QA
I thought about making a test where it would fail without force? but that seemed convoluted at best and prone to breaking. if anyone wants to hit me up with an idea awesome.
### Related info

* upstream issues

this issue is not fixed.

you would need to make the node package pattern install fail if it does not use the --force flag. 

## Docs

NONE

# Merge restrictions

(Please do not edit this) then why is it in the PR template?!

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


ereslibre:
This PR has conflicts now.

PR #680, opened 2019-09-03
testrunner: stream output from subprocesses

ereslibre:
## Why is this PR needed?

testrunner: stream output from subprocesses

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/testrunner
+++ b/ci/infra/testrunner/testrunner
@@ -22,6 +22,5 @@ setup_python_env() {
 }
 
 setup_python_env
-# use unbuffered output
-stdbuf -i0 -o0 -e0 python -u ${SDIR}/testrunner.py "$@"
+python -u ${SDIR}/testrunner.py "$@"
 deactivate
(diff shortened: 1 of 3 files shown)


ci/infra/testrunner/testrunner
```diff
@@ -22,6 +22,5 @@ setup_python_env() {
 }
 
 setup_python_env
-# use unbuffered output
-stdbuf -i0 -o0 -e0 python -u ${SDIR}/testrunner.py "$@"
```
dannysauer: Ugly as it is, I think we might like to leave the `stdbuf` commands. There's some desire (SUSE/avant-garde#670) to have tests running within Jenkins or similar generate synchronous output.  Imagine something like wget which prints progress output before a newline happens, for example.  That, of course, will also require a tweak to how the `read_fd` threads read from the pipe, but we can deal with that later.  This gets us 90% there.

ereslibre: I would like to only keep things we really need, otherwise we can just keep certain changes because they were there before without a real use. Honestly, if we need to have a complete unbuffered output let's do that in the future, in my opinion.

PR #690, opened 2019-09-06
Add testrunner with cloud-provider option

suseclee:
Testrunner can init, bootstrap, join a cluster with cpi option in Openstack.

1. Added the cloud-provider option for testrunner for deployment
`./testrunner bootstrap --cloud-provider openstack`
2. Added a full test for openstack cpi. This will trigger testcases in pytest after cloud-provider cluster is deployed 
`make test_cpi_openstack`

## Why is this PR needed?
https://github.com/SUSE/avant-garde/issues/582

## What does this PR do?
This PR can verify a regression for cloud-provider in Openstack. 

## Docs
Added in testrunner/README.md

### DIFF
--- a/ci/infra/testrunner/requirements.txt
+++ b/ci/infra/testrunner/requirements.txt
@@ -1,5 +1,6 @@
 timeout-decorator
 requests
 pyhcl
-pytest
+pytest==5.1.0
 pyyaml
+pytest-ordering
--- a/ci/infra/testrunner/skuba/skuba.py
+++ b/ci/infra/testrunner/skuba/skuba.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import stat
 
 import platforms
 from utils.format import Format
@@ -45,26 +46,35 @@ class Skuba:
         Utils.cleanup_files(dirs)
 
     @step
-    def cluster_init(self, kubernetes_version=None):
+    def cluster_init(self, kubernetes_version=None, cloud_provider=None):
         logger.debug("Cleaning up any previous test-cluster dir")
         self.utils.cleanup_file(self.cwd)
 
-        k8s_version_option = ""
+        k8s_version_option, cloud_provider_option = "", ""
         if kubernetes_version:
             k8s_version_option = "--kubernetes-version {}".format(kubernetes_version)
-        cmd = "cluster init --control-plane {} {} test-cluster".format(self.platform.get_lb_ipaddr(), k8s_version_option)
+        if cloud_provider:
+            cloud_provider_option = "--cloud-provider {}".format(type(self.platform).__name__.lower())
+
+        cmd = "cluster init --control-plane {} {} {} test-cluster".format(self.platform.get_lb_ipaddr(),
+                                                                          k8s_version_option, cloud_provider_option)
         # Override work directory, because init must run in the parent directory of the
         # cluster directory
         self._run_skuba(cmd, cwd=self.conf.workspace)
 
     @step
-    def node_bootstrap(self):
+    def node_bootstrap(self, cloud_provider=None):
         self._verify_bootstrap_dependency()
 
+        if cloud_provider:
+            self.platform.setup_cloud_provider()
+
         master0_ip = self.platform.get_nodes_ipaddrs("master")[0]
-        cmd = "node bootstrap --user {username} --sudo --target \
-                 {ip} my-master-0".format(ip=master0_ip, username=self.conf.nodeuser)
+        cmd = "node bootstrap --user {nodeuser} --sudo --target \
+                 {ip} caasp-master-{username}-0".format(ip=master0_ip, nodeuser=self.conf.nodeuser,
+                                                        username=self.conf.username)
         self._run_skuba(cmd)
+
 
     @step
     def node_join(self, role="worker", nr=0):
@@ -79,9 +89,9 @@ class Skuba:
             raise Exception(Format.alert("Node {role}-{nr} is not deployed in "
                                          "infrastructure".format(role=role, nr=nr)))
 
-        cmd = "node join --role {role} --user {username} --sudo --target {ip} \
-               my-{role}-{nr}".format(role=role, ip=ip_addrs[nr], nr=nr,
-                                      username=self.conf.nodeuser)
+        cmd = "node join --role {role} --user {nodeuser} --sudo --target {ip} \
+               caasp-{role}-{username}-{nr}".format(role=role, ip=ip_addrs[nr], nr=nr,
+                                      nodeuser=self.conf.nodeuser, username=self.conf.username)
         try:
             self._run_skuba(cmd)
         except Exception as ex:
--- a/ci/infra/testrunner/testrunner.py
+++ b/ci/infra/testrunner/testrunner.py
@@ -36,8 +36,8 @@ def provision(options):
 
 def bootstrap(options):
     skuba = Skuba(options.conf, options.platform)
-    skuba.cluster_init(kubernetes_version=options.kubernetes_version)
-    skuba.node_bootstrap()
+    skuba.cluster_init(kubernetes_version=options.kubernetes_version, cloud_provider=options.cloud_provider)
+    skuba.node_bootstrap(cloud_provider=options.cloud_provider)
 
 
 def cluster_status(options):
@@ -103,7 +103,7 @@ def main():
                         default="openstack",
                         choices=["openstack", "vmware",
                                  "bare-metal", "libvirt"],
-                        help="The platform you're targeting. Defaults to openstack")
+                        help="The platform you're targeting. Default is openstack")
     parser.add_argument("-l", "--log-level", dest="log_level", default=None, help="log level",
                         choices=["DEBUG", "INFO", "WARNING", "ERROR"])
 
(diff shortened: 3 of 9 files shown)


suseclee:
Test Result Summary:
~/go/src/github.com/SUSE/skuba/ci$ make test_cpi_openstack
=================== test session starts ============================
platform linux -- Python 3.6.5, pytest-5.1.0, py-1.8.0, pluggy-0.12.0 -- /home/chang/py3venv/bin/python
cachedir: .pytest_cache
rootdir: /home/chang/go/src/github.com/SUSE/skuba/ci/infra/testrunner/tests, inifile: pytest.ini
plugins: ordering-0.6, timeout-1.3.3
collected 6 items

tests/test_cpi_openstack.py::test_init_cpi_openstack_PASSED
tests/test_cpi_openstack.py::test_bootstrap_cpi_openstack_cluster PASSED
tests/test_cpi_openstack.py::test_node_join_cpi_openstack_cluster PASSED
tests/test_cpi_openstack.py::test_create_cinder_storage PASSED
tests/test_cpi_openstack.py::test_wrtite_cinder_storage PASSED
tests/test_cpi_openstack.py::test_read_cinder_storage PASSED
Cleaning up /home/chang/go/src/github.com/SUSE/tfout
Cleaning up /home/chang/go/src/github.com/SUSE/tfout.json
Cleaning up /tmp/9dcb844899ac746ca7848197d4efd049/ssh-agent-sock

====================== 6 passed in 886.27s (0:14:46) ======================

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -45,27 +46,81 @@ def cleanup(conf):
...
 
+    @step
+    def setup_config_cpi(self, cloud_provider):
+        if cloud_provider == "openstack":
+            self._setup_openstack_cpi()
+        else:
+            raise ValueError(f"{cloud_provider} is not an option for --cloud-provider")
+
+    @step
+    def _setup_openstack_cpi(self):
```
pablochacin: This code is specific to openstack and shouldn't be located in the skuba class. I should better be located in the platform class, as any other platform specific logic. Please move there. And implement the logic for openstack in the openstack class.

ci/infra/testrunner/requirements.txt
```diff
@@ -1,5 +1,6 @@
 timeout-decorator
 requests
 pyhcl
-pytest
+pytest==5.1.0
```
pablochacin: why this specific version?

pablochacin:
Large PR should be split in multiple commits makes review easier, making evident which changes are related.

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -45,27 +46,81 @@ def cleanup(conf):
...
         # cluster directory
         self._run_skuba(cmd, cwd=self.conf.workspace)
 
     @step
-    def node_bootstrap(self):
+    def node_bootstrap(self, cloud_provider=None):
         self._verify_bootstrap_dependency()
 
+        if cloud_provider:
+            self.setup_config_cpi(cloud_provider)
```
pablochacin: this should be `self.platform.conifig_cpi()`

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -45,27 +46,81 @@ def cleanup(conf):
...
         master0_ip = self.platform.get_nodes_ipaddrs("master")[0]
-        cmd = "node bootstrap --user {username} --sudo --target \
-                 {ip} my-master-0".format(ip=master0_ip, username=self.conf.nodeuser)
+        cmd = "node bootstrap --user {nodeuser} --sudo --target \
+                 {ip} caasp-master-{username}-0".format(ip=master0_ip, nodeuser=self.conf.nodeuser,
+                                                        username=self.conf.username)
         self._run_skuba(cmd)
 
+    @step
+    def setup_config_cpi(self, cloud_provider):
```
pablochacin: This code is specific to openstack and shouldn't be located in the `skuba` class. I should better be located in the `platform` class, as any other platform specific logic. Please move there.

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -45,27 +46,81 @@ def cleanup(conf):
...
+        openrc_vars["PRIVATE_SUBNET_ID"] = ""
+        openrc_vars["PUBLIC_NET_ID"] = ""
+        with open(self.conf.openstack.openrc, mode="rt") as openrc:
+            for line in openrc:
+                try:
+                    name, _, var = line.split()[1].partition("=")
+                    if var:
+                        openrc_vars[name] = var.strip('"').strip("'")
+                except:
+                    pass
```
pablochacin: It is safe to ignore exceptions here?

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -45,27 +46,81 @@ def cleanup(conf):
...
+    @step
+    def setup_config_cpi(self, cloud_provider):
+        if cloud_provider == "openstack":
+            self._setup_openstack_cpi()
+        else:
+            raise ValueError(f"{cloud_provider} is not an option for --cloud-provider")
+
+    @step
+    def _setup_openstack_cpi(self):
+        openstack_temp_conf = os.path.join(self.cwd, "cloud", "openstack", "openstack.conf.template")
```
pablochacin: To improve legibility maybe rename as `openstack_conf_template`. I had to read the code three times to realize it wasn't a temporary file.

ci/infra/testrunner/requirements.txt
suseclee: > why this specific version?

If version specific is better than blind version. If someone already had python virtual env a long time back, It would not use without upgrading the packages.  If we add a version, all the package version will be same in all dev machines. Plus, some pytest plugins requires high version of pytest.

ci/infra/testrunner/skuba/skuba.py
suseclee: > It is safe to ignore exceptions here?

Yes, this code parse variables from openrc.sh. If there are other than variable assignment (=) like comments, shebang, and etc, will be in pass, which is same as continue.

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -45,27 +46,42 @@ def cleanup(conf):
...
-        cmd = "node bootstrap --user {username} --sudo --target \
-                 {ip} my-master-0".format(ip=master0_ip, username=self.conf.nodeuser)
+        cmd = "node bootstrap --user {nodeuser} --sudo --target \
+                 {ip} caasp-master-{username}-0".format(ip=master0_ip, nodeuser=self.conf.nodeuser,
+                                                        username=self.conf.username)
         self._run_skuba(cmd)
 
+    @step
+    def setup_config_cpi(self, cloud_provider):
+        if cloud_provider == "openstack":
```
pablochacin: This can further improved the decoupling of skuba class from this logic if you implement the `setup_cloud_provider`method in the `Platform class and by default it return an error and overrides it in the openstack class, as you did:`
```
class Platform:

    def setup_cloud_provider():
          raise Exception("Cloud provider not supported for this platform")
```

Unfortunately, here we miss the name of the platform (because how Platform class is presently implemented) but I prefer this approach.

ci/infra/testrunner/testrunner.py
```diff
@@ -132,6 +132,10 @@ def main():
                         deployed nodes in your platform")
     cmd_bootstrap.add_argument("-k", "--kubernetes-version", help="kubernetes version",
                                dest="kubernetes_version", default=None)
+    cmd_bootstrap.add_argument("-c", "--cloud-provider",
```
pablochacin: Do we need to have the cloud provider here? the test runner already knows the platform by means of the `--platform` argument. Does it make sense that `--platform` and `--cloud-provider` has different values? I think not. Therefore, I would make this option to store a boolean.

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -45,27 +46,42 @@ def cleanup(conf):
...
         logger.debug("Cleaning up any previous test-cluster dir")
         self.utils.cleanup_file(self.cwd)
 
-        k8s_version_option = ""
+        k8s_version_option, cloud_provider_option = "", ""
         if kubernetes_version:
             k8s_version_option = "--kubernetes-version {}".format(kubernetes_version)
-        cmd = "cluster init --control-plane {} {} test-cluster".format(self.platform.get_lb_ipaddr(), k8s_version_option)
+        if cloud_provider:
+            cloud_provider_option = "--cloud-provider {}".format(cloud_provider)
```
pablochacin: If you remove this from the `testrunner` cli you can retrieve the cloud provider from the `Skuba`  by adding this to the `__init__` method:

```
    def __init__(self, conf, platform):
          self.target_platform = platform
```
You can now use `self.target_platform`  as the name of the cloud provider.

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -45,27 +46,36 @@ def cleanup(conf):
...
-    def node_bootstrap(self):
+    def node_bootstrap(self, cloud_provider=None):
         self._verify_bootstrap_dependency()
 
+        if cloud_provider:
+            self.platform.setup_cloud_provider()
+
         master0_ip = self.platform.get_nodes_ipaddrs("master")[0]
-        cmd = "node bootstrap --user {username} --sudo --target \
-                 {ip} my-master-0".format(ip=master0_ip, username=self.conf.nodeuser)
```
MaximilianMeister: This actually broke the upgrade tests, as we assert these exact names there. Fix is in https://github.com/SUSE/skuba/pull/700

PR #693, opened 2019-09-09
Make `runshellcommand` return a result object

ereslibre:
## Why is this PR needed?

`runshellcommand` now returns a `CommandResult` object that contains
all attributes regarding the execution of the command: `retcode`,
`stdout` and `stderr`.

Tests are then free to retrieve the information they need to perform
their checks.

## Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/skuba/skuba.py
+++ b/ci/infra/testrunner/skuba/skuba.py
@@ -162,8 +162,8 @@ class Skuba:
 
         test_cluster = os.path.join(self.conf.workspace, "test-cluster")
         cmd = "cd " + test_cluster + "; " + self.binpath + " cluster status"
-        output = self.utils.runshellcommand(cmd)
-        return output.count(role)
+        res = self.utils.runshellcommand(cmd)
+        return res.stdout.count(role)
 
     @step
     def get_kubeconfig(self):
--- a/ci/infra/testrunner/skuba/skuba.py
+++ b/ci/infra/testrunner/skuba/skuba.py
@@ -162,8 +162,8 @@ class Skuba:
 
         test_cluster = os.path.join(self.conf.workspace, "test-cluster")
         cmd = "cd " + test_cluster + "; " + self.binpath + " cluster status"
-        res = self.utils.runshellcommand(cmd)
-        return res.stdout.count(role)
+        output = self.utils.runshellcommand(cmd)
+        return output.count(role)
 
     @step
     def get_kubeconfig(self):
--- a/ci/infra/testrunner/tests/test_cilium.py
+++ b/ci/infra/testrunner/tests/test_cilium.py
@@ -20,23 +20,23 @@ def test_cillium(deployment, kubectl):
 
     logger.info("Check with L3/L4 policy")
     kubectl.run_kubectl("create -f https://raw.githubusercontent.com/cilium/cilium/v1.5/examples/minikube/sw_l3_l4_policy.yaml")
-    tie_out = kubectl.run_kubectl("exec tiefighter -- {}".format(landing_req)).stdout
+    tie_out = kubectl.run_kubectl("exec tiefighter -- {}".format(landing_req))
     assert 'Ship landed' in tie_out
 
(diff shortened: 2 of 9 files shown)


ereslibre:
Added the `do not merge` label to check if I can make this change less intrusive.

pablochacin:
@ereslibre As a summary of what we discussed in the chat, one alternative I had considered was to add a `redirect_stderr` flag to the `runshellcommand` function which basically works like the shell redirect directive `2>&1` and will return `stderr` concatenated with `stdout`. 

I'm aware this is not the same as been able to check `stderr` separately from `stdout` but maybe it will cover the uses cases we have in testing and has no impact in the existing code. If more cases appear that need checking exclusively the `stderr`, we can consider other approaches.

ereslibre:
> As a summary of what we discussed in the chat, one alternative I had considered was to add a redirect_stderr flag to the runshellcommand function which basically works like the shell redirect directive 2>&1 and will return stderr concatenated with stdout.

I think this wouldn't be a good practice in terms of testing. It's perfectly fine for our users to redirect stderr to stdout, but at testing level we should make sure not only that messages exist in the output (or the lack thereof), but also that they get printed on the right stream.

Redirecting `stderr` to `stdout` on tests will remove a lot of important context when testing: not only if the output is present/absent; where it happened. Depending on what we write on `stdout` or `stderr`, or if we change those in the product lifecycle can break users pipes (if they didn't redirect).

ereslibre:
I'm working on a slightly different solution than the one proposed in this PR; by default it will refer to `stdout` (as the existing tests are doing this), but it also allows to inspect `stderr` and `retcode` if needed in an explicit way. I'm facing some python related problems though, the idea is along the lines:

```python
class CommandResult:
    """
    Embeds the result of running a command for easier testing of different
    output channels (stdout, stderr) and the return code of the execution.
    """
    def __init__(self, retcode, stdout, stderr):
        self.retcode = retcode
        self.stdout = stdout
        self.stderr = stderr

    def __eq__(self, other):
        if isinstance(other, str):
            return self.stdout == other
        elif isinstance(other, self.__class__):
            return self.retcode == other.retcode and self.stdout == other.stdout and self.stderr == other.stderr
        return False

    def __iter__(self):
        return iter(self.stdout)

    def __getattr__(self, name):
        print("calling to {}", name)
        def forward_method(*args, **kwargs):
            return getattr(self.stdout, name)(*args, **kwargs)
        return forward_method

def test_equality():
    res = CommandResult(0, "stdout", "stderr")
    assert res == "stdout"

def test_method_calling():
    res = CommandResult(0, "stdout", "stderr")
    assert res.count("stdout") == 1
```

However, `assert "stdout" in res` and `assert "stdout" not in res` is not working as expected, so I'm still investigating.

ereslibre:
@MaximilianMeister, @pablochacin: changed the implementation so existing tests don't require any change. `CommandResult` will delegate most actions to its `stdout` attribute, so all tests if not explicit will act against `stdout`. They can choose to read `retcode` or `stderr` if they want.

What do you think about this approach? Note that it's a separate commit, if you are fine with this change I'll squash.

ereslibre:
Also, as an aside, I have some tests for the `CommandResult` class itself, I couldn't find `testrunner` specific tests, but it would be great if I could add those as well.

ereslibre:
Closing, the class that delegates doesn't work as expected when casted directly to ints (as opposed to previous int(output)), and can introduce unexpected behavior in the future when relying on behavior not yet uncovered.

Returning an object with three attributes (`retcode`, `stdout` and `stderr`) is considered harmful and an anti-pattern because all tests need to be adapted to read `.stdout`. Even though in my opinion we should be testing `stdout` and `stderr` explicitly. This will be even more important if we adopt the `kubeadm`'s machine output (json/yaml) that will happen through `stdout`. A user can always unmarshal this contents, while human errors would go on `stderr`.

Doing on our tests the `2>&1` equivalent unifies all output and won't allow us to test specific messages on the different streams, that was the reason of this proposal.

I'm closing because I don't have a better solution, but I hope the statement of the problem is clear.

PR #694, opened 2019-09-10
runshellcommand return command result

ereslibre:
## Why is this PR needed?

Make `runshellcommand` return a result object
    
`runshellcommand` now returns a `CommandResult` object that contains
all attributes regarding the execution of the command: `retcode`,
`stdout` and `stderr`.
    
Tests are then free to retrieve the information they need to perform
their checks.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/utils/command.py
+++ b/ci/infra/testrunner/utils/command.py
@@ -1,0 +1,44 @@
+class CommandResult:
+    """
+    Embeds the result of running a command for easier testing of different
+    output channels (stdout, stderr) and the return code of the execution.
+
+    Delegates iterators, direct comparisons and unknown attribute retrieval
+    (including function calls) to the `stdout` attribute for better testing
+    ergonomics.
+    """
+    def __init__(self, retcode, stdout, stderr):
+        self.retcode = retcode
+        self.stdout = stdout
+        self.stderr = stderr
+
+    def __eq__(self, other):
+        if isinstance(other, str):
+            return self.stdout == other
+        elif isinstance(other, self.__class__):
+            return self.retcode == other.retcode and self.stdout == other.stdout and self.stderr == other.stderr
+        return False
+
+    def __contains__(self, other):
+        if isinstance(other, str):
+            return self.stdout.__contains__(other)
+        elif isinstance(other, self.__class__):
+            return self.retcode == other.retcode and self.stdout.__contains__(other.stdout) and self.stderr.__contains__(other.stderr)
+        return False
+
+    def __iter__(self):
+        return self.stdout.__iter__()
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}({self.retcode!r}, {self.stdout!r}, {self.stderr!r})'
+
+    def __str__(self):
+        return self.stdout
+
+    def __int__(self):
+        return int(self.__str__())
+
+    def __getattr__(self, name):
+        def forward_method(*args, **kwargs):
+            return getattr(self.stdout, name)(*args, **kwargs)
+        return forward_method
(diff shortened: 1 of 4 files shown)


ci/infra/testrunner/utils/command.py
```diff
@@ -0,0 +1,44 @@
...
+        self.retcode = retcode
+        self.stdout = stdout
+        self.stderr = stderr
+
+    def __eq__(self, other):
+        if isinstance(other, str):
+            return self.stdout == other
+        elif isinstance(other, self.__class__):
+            return self.retcode == other.retcode and self.stdout == other.stdout and self.stderr == other.stderr
+        return False
```
dannysauer: Should this also implement integer comparison, or do we assume that such a comparison will probably wrap this in `int()`?  Maybe so; anyone doing `>`, `<`, etc would also use `int()`.

ci/infra/testrunner/utils/command.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __iter__(self):
+        return self.stdout.__iter__()
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}({self.retcode!r}, {self.stdout!r}, {self.stderr!r})'
+
+    def __str__(self):
+        return self.stdout
```
dannysauer: The old code returns stderr in the case of an error, but I personally prefer the predictability of "the string representation is always stdout".  So, I'm on-board with this change.

ci/infra/testrunner/utils/command.py
ereslibre: For use cases that were doing `int(stdout)` in our test code, we have the cast `__int__` implemented, for comparisons I'd say comparing against an string or against another `CommandResult` object would be enough. Even only with string, the other was implemented for completion, but I doubt of its usefulness honestly.

ci/infra/testrunner/utils/command.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __int__(self):
+        return int(self.__str__())
```
ereslibre: I'm not yet sure about the intuition in this way though. I'm fine keeping it as is, but I see two different intuitions here:

- `int(aCommandResult)` returns `self.retcode`
- `int(aCommandResult)` returns `int(self.__str__())`, what is effectively `int(self.stdout)`

I would expect the first behavior I think, and users that want to parse the `stdout` to an int could do `int(aCommandResult.stdout)`.

dannysauer: I'm personally inclined to expect the behavior of `int` to be stdout as an integer, just because "object treated as scalar value" behaves as a string containing stdout. While the proposed idea of using the return code does make sense, it seems like it might be non-intuitive to have a string become a completely different value when wrapped in a certain method.

I view `bool` as a different case, since that'd usually not be explicitly wrapped by a method; it'd usually be `if shellcommand()... `, I'd think. I know it's the same argument, but I think the feeling people have about program execution succeeding or failing is strong enough that having a different value makes sense.

But in both cases, I might just be projecting my personal thoughts onto everyone - which could be completely wrong. :D

ereslibre:
I'm closing this PR because I agree there's no "ideal" solution to this issue. Moreover, after some thought I think modifying `runshellcommand` at large doesn't really make sense, because most of the times we are fine with the current approach.

Maybe, if desired, it would make more sense to only adapt `_run_skuba` and make this call return the said triplet, or instead of `_run_skuba` a explicit method that will return the triplet.

In any case, I think changing `runshellcommand` with any of the mentioned changes (triplet, explicit + changes everywhere, implicit + potential maintenance issues) is not optimal, thus, I'm closing this PR.

Thanks to everyone involved.

PR #706, opened 2019-09-19
Redirect pip3 install output to /dev/null

pablochacin:
## Why is this PR needed?

While running testrunner locally sending pip output to
stdout makes execution noisy.

This commit reverts 6a395d7e which was introduced to make
debugging in CI easier.

## What does this PR do?

Redirects pip output to /dev/null

## Info for QA

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/testrunner
+++ b/ci/infra/testrunner/testrunner
@@ -18,7 +18,7 @@ setup_python_env() {
   source ${VENVDIR}/bin/activate
   # Not calling to `pip3` directly because its shebang can exceed 128 characters
   # and this operation will fail. Check: https://github.com/pypa/pip/issues/1773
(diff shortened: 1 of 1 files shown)


kkaempf:
Hmm, does this just silence `pip3`s normal output ? Do pip3 error messages still get through ? (Does pip3 properly log errors via `stderr` ? Are these error message understandable without the `stdout` output ?)

stefsuse:
I agree with @kkaempf. if this was added to make CI debugging easier, what is the reason to revert now?

pablochacin:
@stefsuse  it was added to fix a particular issue, but in day-to-day use of testrunner it produces a significant noise. I think optimizing for exceptional cases at expense of usability is not a good tradeoff.

PR #710, opened 2019-09-20
Reference ssh key by path

pablochacin:
# Why is this PR needed?

Presently, the location for the ssh keys used by testrunner is defined indirectly by the `ssh_key_option` parameter which has two possible values:
* id_shared: the key at `<skuba.srcpath>/ci/infra/id_shared` is used
* id_rsa: the key at `<user home>/.ssh/id_rsa` is used

This has several inconvenientes and limitations:
* It is non intuitive
* Assumes the user uses `id_rsa`
* Assumes the skuba source is locally available and the skuba srcpath is specified
* Doesn't allow a team to share they own keys

Fixes https://github.com/SUSE/avant-garde/issues/688
Supersedes: https://github.com/SUSE/skuba/pull/593 

## What does this PR do?

Replace the `ssh_key_option` configuration parameter by the `ssh_key` parameter, which
points to to ssh key to be used. It defaults to the `id_rsa` key at the `~/.ssh` for the current user.

Additionally, make `vars.yaml` to point to the `shared_id` keys in the  `skuba/ci/
directory` to make it compatible with the CI setup.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/README.md
+++ b/ci/infra/testrunner/README.md
@@ -109,9 +109,7 @@ There are some arguments that are currently at the top level of the configuratio
 
 - log_dir: path to the directory where platform logs are collected. Defaults to `<workspace>/testrunner_logs` 
 - nodeuser: the user name used to login into the platform nodes. Optional. 
-- ssh_option: specifies the location of the key used to access nodes. Can have two possible values:
-  - "id_shared": uses the key located at `<skuba src directory>/ci/infra/id_shared` (default)
-  - "id_rsa": uses the user's key located at `$HOME/.ssh/id_rsa`
+- ssh_key: specifies the location of the key used to access nodes. The default is to use the user's key located at `$HOME/.ssh/id_rsa`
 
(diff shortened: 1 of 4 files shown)


ci/infra/testrunner/README.md
```diff
@@ -234,7 +232,7 @@ Be sure you don't specify the `terraform.tfdir` directory, so terraform configur
 You use your `id_rsa` keys to connect to the cluster nodes, as the `shared_id` is not available.
 
 ```
-ssh_option: "id_rsa"
+ssh_key: "path/to/id_rsa"
```
dannysauer: I missed this earlier, but it's not a huge deal - typically you'd want to start this with a `/` to indicate that it needs to be an absolute path.  But that can be fixed later.
```suggestion
ssh_key: "/path/to/id_rsa"
```

PR #717, opened 2019-09-24
Dex/gangway HA deployments (#bsc1143232)

jenting:
Align with v3, replicas set to 3.

## Why is this PR needed?

At v3, the number of dex replicas is 3 and also for HA requirement.

Fixes #bsc1143232

## What does this PR do?

Change deployment replicas from 1 to 3.

N/A

## Info for QA

### Related info

N/A

After cluster deployed, run
```
kubectl get pods -n kube-system
```
1 pod for dex and gangway separately.

After cluster deployed, run
```
kubectl get pods -n kube-system
```
3 pods for dex and gangway separately.

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/addons/dex.go
+++ b/internal/pkg/skuba/addons/dex.go
@@ -159,7 +159,7 @@ metadata:
   name: oidc-dex
   namespace: kube-system
 spec:
-  replicas: 1
+  replicas: 3
   selector:
     matchLabels:
       app: oidc-dex
--- a/internal/pkg/skuba/addons/gangway.go
+++ b/internal/pkg/skuba/addons/gangway.go
@@ -122,7 +122,7 @@ metadata:
   labels:
     app: oidc-gangway
 spec:
-  replicas: 1
+  replicas: 3
   selector:
     matchLabels:
       app: oidc-gangway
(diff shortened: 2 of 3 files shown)


innobead:
@ereslibre not sure if we should manually bump add on version? When bumping version, it means it gets ready to release. Should we think of bumping updated addons versions by automatic way when it’s ready to release to avoid any manual intervention. (Maybe someone needs to update the same code, but need to bump the version again even not released yet) if missing something, please correct. Thanks.

@jordimassaguerpla for current v4, it doesn’t have addon mechanism, so which branch will be cherrypicked to then release this fix to update channel. (this will be the question for following fixes). Thanks.

ereslibre:
> @ereslibre not sure if we should manually bump add on version? When bumping version, it means it gets ready to release. Should we think of bumping updated addons versions by automatic way when it’s ready to release to avoid any manual intervention. (Maybe someone needs to update the same code, but need to bump the version again even not released yet) if missing something, please correct. Thanks.

Note that the version bumped is the "metadata" one, it's only for skuba to know that it should reapply that addon when running `skuba addon upgrade apply`. This happens if we changed something on the manifests, but the version of the component is the same as previously. Not sure if I understood your question.

innobead:
@ereslibre answer indirectly, thanks

ereslibre:
> This should be refactored.

Hmm, it was changed for every Kubernetes version we allow to deploy (`1.14.1`, `1.15.0` and `1.15.2`). In this case, I think explicit is better than optimizing for avoiding repetition, because a certain addon could be not present or at an older version on a previous Kubernetes version, for example.

We can give it a thought though, could be that there's a better pattern while still keeping the explicit version benefits.

PR #719, opened 2019-09-24
CODEOWNERS: Clean up file before turning mandatory reviews on

hwoarang:
The CODEOWNERS file was being used as a notification method for those
who were interested in specific domains. However, it was decided to
repurpose it and make reviews from CODEOWNERS mandatory. As a result of
which, we need to clean it up in order to prevent dragging people into
reviews before they sign up for it.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,16 +12,8 @@
 # Order is important; the last matching pattern takes the most
 # precedence.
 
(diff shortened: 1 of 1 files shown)


hwoarang:
Yes this PR removes all reviewers except for the file itself. Mandatory reviews only become applicable for patterns in this file which are none right now.

PR #721, opened 2019-09-24
Added timeout to Jenkins user input

tdaines42:
Otherwise it will hang until the pipeline timeout

## Why is this PR needed?

If someone who's not a collaborator submits a pull request the pipeline will be paused until it's approved in Jenkins. This will eventually block other runs until the job times out.

Fixes #

## What does this PR do?

Adds a timeout to the input

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/jenkins/pipelines/prs/skuba-code-lint.Jenkinsfile
+++ b/ci/jenkins/pipelines/prs/skuba-code-lint.Jenkinsfile
@@ -23,9 +23,24 @@ pipeline {
                 echo "Test execution for collaborator ${CHANGE_AUTHOR} allowed"
 
             } else {
-                def allowExecution = input(id: 'userInput', message: "Change author is not a SUSE member: ${CHANGE_AUTHOR}", parameters: [
-                    booleanParam(name: 'allowExecution', defaultValue: false, description: 'Run tests anyway?')
-                ])
+                def allowExecution = false
+
+                try {
+                    timeout(time: 5, unit: 'MINUTES') {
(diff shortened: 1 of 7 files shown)


pablochacin:
I'm not sure how useful is to have a timeout of 5 minutes. I was looking around and found that an alternative approach using "agent none" for input stage as described in [this post](https://stackoverflow.com/questions/46276140/jenkins-declarative-pipeline-user-input-parameters) However, depending on where in the pipeline we put this input, ensuring all stages run in the same node may result [more complex](https://stackoverflow.com/questions/42159221/use-a-lightweight-executor-for-a-declarative-pipeline-stage-agent-none).

stefsuse:
How about a timeout of 1 minute for the short term, and if blocking becomes an issue later look into "agent none"?

pablochacin:
@stefsuse I would actually propose to increase the timeout to 15 o 30 minutes to give time to actually go to the ci and approve the PR. If this becomes a problem, then look into the alternative as you propose.

dannysauer:
I don't see the benefit of a long timeout holding resources up.  If it fails fast, someone will eventually notice that it failed and come back to restart the job - which is really the only way this will be encountered 90% of the time.  They'll see the prompt essentially interactively at that time, and approve.  Given that workflow, I think the wiggle room left by a 5-minute timeout is probably about right.  One minute seems a little short to me, but either way.

I'd approve of this as-is, but it'd be nice to see a decision on the suggestion raised in @pablochacin 's `agent none` link before that. :)

PR #722, opened 2019-09-26
Provide function for waiting actions in tests

pablochacin:
## Why is this PR needed?

Sometimes tests include actions such as executing commands in the nodes which must wait for a bounded time, and retried in case of failure. It would be convenient to implement this logic as part of the test library to be reused by tests.

Fixes https://github.com/SUSE/avant-garde/issues/871

## What does this PR do?

Implements the `wait` function which allows waiting actions in tests
to complete and retry in case of timeout or selected exceptions occurs.

N/A

## Info for QA

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/tests/utils.py
+++ b/ci/infra/testrunner/tests/utils.py
@@ -1,0 +1,34 @@
+import signal
+import time
+
+def wait(func, *args, **kwargs):
+
+    class TimeoutError(Exception):
+        pass
+
+    timeout = kwargs.pop("wait_timeout", 5)
+    delay   = kwargs.pop("wait_delay", 0)
+    backoff = kwargs.pop("wait_backoff", 1)
+    retries = kwargs.pop("wait_retries", 3)
+    allow   = kwargs.pop("wait_allow", ())
+    def _handle_timeout(signum, frame):
+        raise TimeoutError()
+
+    time.sleep(delay)
+    reason=""
+    for i in range(0, retries):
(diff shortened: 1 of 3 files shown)


pablochacin:
@Klaven I was undecided about to rewrite one of the existing tests using the function but maybe makes sense.

ci/infra/testrunner/tests/utils.py
```diff
@@ -0,0 +1,34 @@
...
+    retries = kwargs.pop("wait_retries", 3)
+    allow   = kwargs.pop("wait_allow", ())
+    def _handle_timeout(signum, frame):
+        raise TimeoutError()
+
+    time.sleep(delay)
+    reason=""
+    for i in range(0, retries):
+        signal.signal(signal.SIGALRM, _handle_timeout)
+        signal.alarm(timeout)
```
ereslibre: This solution won't allow for nesting wait calls (e.g. calling `wait` when the callback `func` calls `wait` again on its own). Creating a thread and using `thread.wait(timeout=<timeout>)` would be safer if we expect this scenario.

pablochacin: Good point. Can we document this limitation of nested waits and implement the thread in a follow up PR?

ereslibre: Absolutely, resolving.

PR #725, opened 2019-09-26
[VMware] Fix HAProxy conn err (bsc#1145907)

lcavajani:
# Why is this PR needed?

HAProxy was not closing the connection with
the apiservers so after 2000 conn (default) over 24h,
it was not accepting new connections. The connections
were stuck in CLOSE_WAIT state.

Here were the issues:

* We were using haproxy v1.8.1 from HA Product and not from HA Updates,
I've tested with the one from HA Updates (1.8.7) and the issue is gone.
This is fixed here by adding the Updates repo to the LB. FWIW, it's
working fine as well with HAProxy 2.0.5

* In version 1.8.1, lowering the timeouts was fixing the issue since the
connection were closed automatically after a few seconds instead
of the 24h that was configured. However, this 24h timeout were set
here to kubectl logs/forward can work but there is way to use
proper timeouts here:

  timeout connect 5s
  timeout client 5s
  timeout server 5s
  timeout tunnel 86400s

This way, we allow only long connection when tunneling. In HTTP mode we
would have a finer way to deal with this but this is conf if good enough for us.

Fixes bsc#1145907

## Info for QA

When running this script, it was always stuck around 1990 which is close the max default (2000).

```bash
#!/usr/bin/env bash

i=0
while true; do
  kubectl --kubeconfig "$KUBECONFIG" get nodes;
  echo ">>> $i"
  ((++i))
done
```

And the following on the LB shows the same amount of connections in `CLOSE_WAIT` state.

> ss|grep sun-sr-https| grep '$CLIENT_IP'|wc -l

With the current fix, the connections are correctly closed and it's possible to go above 2000 connections without issues.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/vmware/cloud-init/lb.tpl
+++ b/ci/infra/vmware/cloud-init/lb.tpl
@@ -32,54 +32,10 @@ ${repositories}
 packages:
   - haproxy
 
-write_files:
-- path: /etc/haproxy/haproxy.cfg
-  content: |
-    global 
-      log /dev/log local0 debug
-      user haproxy
-      group haproxy
-      daemon
-    
-    defaults
-      mode      tcp
-      log       global
-      option    redispatch
-      option    tcpka
-      retries   2
-      http-check     expect status 200
-      default-server check check-ssl verify none
-      timeout connect 5s
-      timeout client 5s
-      timeout server 5s
-      timeout tunnel 86400s
-    
-    listen stats
-      bind    *:9000
-      mode    http
-      stats   hide-version
-      stats   uri       /stats
-
-    listen apiserver
-      bind    *:6443
-      option  httpchk GET /healthz
-      ${apiserver_backends}
-
-    listen gangway
-      bind    *:32001
-      option  httpchk GET /
-      ${gangway_backends}
-
-    listen dex
-      bind    *:32000
-      option  httpchk GET /healthz
-      ${dex_backends}
-
 runcmd:
   # Since we are currently inside of the cloud-init systemd unit, trying to
   # start another service by either `enable --now` or `start` will create a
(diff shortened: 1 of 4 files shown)


lcavajani:
I also configured the login so we get info in journalctl

ci/infra/vmware/cloud-init/lb.tpl
```diff
@@ -35,42 +35,43 @@ packages:
...
+    listen apiserver
+      bind    *:6443
+      option  httpchk GET /healthz
       ${apiserver_backends}
 
-    backend gangway-backend
-      option httpchk GET /
-      http-check expect status 200
+    listen gangway
+      bind    *:32001
```
jenting: any reason for removes gangway httpchk?

lcavajani: It got lost on the way, it's back. Thanks for the catch!

lcavajani:
Do not merge, I will add the fix to allow removing/adding masters

lcavajani:
let's merge this first, I will send an other PR ;)

lcavajani:
I've added the fix to avoid the recreation of the load-balancer when a master is added or removed. From now on, only the configuration will be updated.

thehejik:
* ha_updates repo is added into lb instance
* latest haproxy is installed `haproxy-1.8.17~git0.e89d25b2-6.23.x86_64`
* after sending ~4000 requests for kubeapi, dex and gangway for each there is still only 62 connections and none of them is in CLOSE_WAIT state on loadbalancer node

I need to verify the last part with changing haproxy.conf instead of redeploying lb instance again.

PR #726, opened 2019-09-26
Bump gangway version and change dex image (bsc#1148412)

jenting:
## Why is this PR needed?

For better package naming and bump gangway to a new version

Fix #bsc1148412

## What does this PR do?

Package renaming from caasp-dex to dex and bump gangway version to v3.2.0 in order to use the updated package `caasp-dex` (the HTML file without using the external resource)

N/A

## Info for QA

### Related info

  - package: caasp-branding, dex, gangway
  - container image: caasp-dex-image, caasp-gangway-image

Open browser's Developer's tool and do user login via dex/gangway GUI, it _contains_ using external Javascript and CSS file which would fail in Airgap env.

Open browser's Developer's tool and do user login via dex/gangway GUI, it _should not contains_ using external Javascript and CSS file which would pass in Airgap env.

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -92,7 +92,7 @@ var (
 				Cilium:  &AddonVersion{"1.5.3", 0},
 				Kured:   &AddonVersion{"1.2.0", 0},
 				Dex:     &AddonVersion{"2.16.0", 1},
-				Gangway: &AddonVersion{"3.1.0", 1},
+				Gangway: &AddonVersion{"3.2.0", 1},
```
ereslibre: Please, bump the manifest version as well here... this number should be always bumped in any case.

jenting: Hmmm... I thought we need to bump version when skuba has a new version release. Am I right?

ereslibre: Yes, in that case it's better if we both bump the version and the manifest version. As per [the RFC](https://github.com/SUSE/caasp-rfc/blob/master/accepted/0043-addon-upgrade.md), the logic could only check the manifest version (despite the current addon logic is actually taking into account both version and manifest version)

PR #728, opened 2019-09-27
Crio configuration airgap

bergmannf:
## Why is this PR needed?

It allows the customer to specify a registry mirror during cluster initialization.

This way it is possible to redirect images that would be pulled from one registry to pull the from another - by redirecting `registry.suse.com` to an internally hosted registry this allows airgapped installation.

The PR will generate a `registries.conf` file with the `v2` format for crio (https://github.com/containers/image/blob/master/docs/containers-registries.conf.5.md).

Right now the included test cases onl

### Related info

Part of the epic: https://github.com/SUSE/avant-garde/issues/567

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/cmd/skuba/cluster/init.go
+++ b/cmd/skuba/cluster/init.go
@@ -41,17 +41,17 @@ type initOptions struct {
 	RegistryMirror    string
 }
 
-func ParseCrioRegistryConfiguration(config string) cluster.CrioMirrorConfiguration {
+func ParseCrioRegistryConfiguration(config string) *cluster.CrioMirrorConfiguration {
 	values := strings.Split(config, ":")
 	if len(values) != 3 {
-		return cluster.CrioMirrorConfiguration{}
+		return nil
 	}
 	insecure, err := strconv.ParseBool(values[2])
 	// Default to secure registry
 	if err != nil {
 		insecure = false
 	}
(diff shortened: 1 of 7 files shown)


bergmannf:
I also don't like the CLI syntax right now - I tried to base it on the `docker` style with separating it by `:`, but it is ugly and easy to mess up for the user.

I just think it is nice if the customer doesn't have to change the `registries.conf` on every single cluster node, which is not that scalable if the cluster gets big.

Maybe it would be feasible to just have the customer create the `addons/containers/registries.conf` file by hand and if that file exists then `skuba` would move the file to the machine?

flavio:
I think at this point we just need to document which files should be changed and how. I don't want skuba to end up with lots of options.

Customers have already their own tools and procedures to handle configuration management on their linux boxes. We are now a regular SLES node, hereby they can use salt, ansible, puppet, SUMA, AutoYaST, bash scripts written by their operators, ... to handle that.

This is similar to the discussion we had [here](https://github.com/SUSE/avant-garde/issues/846)

bergmannf:
Closing in favour of https://github.com/SUSE/doc-caasp/pull/545

PR #729, opened 2019-09-27
Addon upgrade plan command

MaximilianMeister:
## Why is this PR needed?

Check if there are new addon updates available

## What does this PR do?

Adds a new command `skuba addon upgrade plan`

Can be tested by manipulating [these numbers](https://github.com/SUSE/skuba/blob/master/internal/pkg/skuba/kubernetes/versions.go#L92-L96)

## Docs

will follow

### DIFF
--- a/internal/pkg/skuba/upgrade/addon/version.go
+++ b/internal/pkg/skuba/upgrade/addon/version.go
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2019 SUSE LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package addon
+
+import (
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubeadm"
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
+	skubaconfig "github.com/SUSE/skuba/internal/pkg/skuba/skuba"
+)
+
+type AddonVersionInfoUpdate struct {
+	Current kubernetes.AddonsVersion
+	Updated kubernetes.AddonsVersion
+}
+
+func UpdatedAddons() (AddonVersionInfoUpdate, error) {
+	client, err := kubernetes.GetAdminClientSet()
+	if err != nil {
+		return AddonVersionInfoUpdate{}, err
+	}
+	currentClusterVersion, err := kubeadm.GetCurrentClusterVersion(client)
+	if err != nil {
+		return AddonVersionInfoUpdate{}, err
+	}
+	currentVersion := currentClusterVersion.String()
+	latestAddonVersions := kubernetes.Versions[currentVersion].AddonsVersion
+
+	skubaConfig, err := skubaconfig.GetSkubaConfiguration()
+	if err != nil {
+		return AddonVersionInfoUpdate{}, err
+	}
+	aviu := AddonVersionInfoUpdate{
+		Current: kubernetes.AddonsVersion{},
+		Updated: kubernetes.AddonsVersion{},
+	}
+
+	for addonName, version := range latestAddonVersions {
+		skubaConfigVersion := skubaConfig.AddonsVersion[addonName]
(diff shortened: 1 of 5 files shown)


MaximilianMeister:
## Example outputs

Has updates:
```
% skuba addon upgrade plan                                                                                                                                                               
skuba version: untagged (v1.0.0-129-g764f1b68) (development) 20190927 go1.12.9
Current Kubernetes cluster version: 1.15.2

Addon upgrades:
  - cilium: 1.5.3 -> 1.5.3 (manifest version from 0 to 4)
  - kured: 1.2.0 -> 1.3.0
  - dex: 2.16.0 -> 2.16.1
  - gangway: 3.1.0 -> 3.1.0 (manifest version from 1 to 2)
  - psp: 1.0.0 -> 1.1.0
```

Has no updates:
```
% skuba addon upgrade plan                                                                                                                                                                 
skuba version: untagged (v1.0.0-129-g5329e25c) (development) 20190927 go1.12.9
Current Kubernetes cluster version: 1.15.2

Congratulations! Addons are already at the latest version available
```

internal/pkg/skuba/upgrade/addon/version.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+	return false
+}
+
+func HasAddonManifestUpdateWithAddon(aviu AddonVersionInfoUpdate, addon kubernetes.Addon) bool {
+	if aviu.Updated[addon].ManifestVersion > aviu.Current[addon].ManifestVersion {
+		return true
+	}
+	return false
```
ereslibre: Minor suggestion:

`return aviu.Updated[addon].ManifestVersion > aviu.Current[addon].ManifestVersion`

internal/pkg/skuba/upgrade/addon/version.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func HasAddonVersionUpdateWithAddon(aviu AddonVersionInfoUpdate, addon kubernetes.Addon) bool {
+	if aviu.Updated[addon].Version > aviu.Current[addon].Version {
+		return true
+	}
+	return false
```
ereslibre: Minor suggestion:

`return aviu.Updated[addon].Version > aviu.Current[addon].Version`

pkg/skuba/actions/addon/upgrade/plan.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return err
+	}
+	currentVersion := currentClusterVersion.String()
+	latestVersion := kubernetes.LatestVersion().String()
+	allNodesVersioningInfo, err := kubernetes.AllNodesVersioningInfo()
+	if err != nil {
+		return err
+	}
+	allNodesMatchClusterVersion := kubernetes.AllNodesMatchClusterVersionWithVersioningInfo(allNodesVersioningInfo, currentClusterVersion)
```
ereslibre: I think we could extract this logic because `skuba addon upgrade apply` will also need to perform this check, and this logic can be reused there. Can also be refactored on the `skuba addon upgrade apply` PR though.

MaximilianMeister: As it is just one function call, there's not so much to extract I'm afraid. Also we need `currentClusterVersion` in the plan output, so extracting this as well would mean we have to fetch it twice. I can only think of this but I think it provides not much added value:

```go
package addon

func SanityCheck(allNodesVersioningInfo kubernetes.NodeVersionInfoMap, currentClusterVersion *version.Version) bool {
	return kubernetes.AllNodesMatchClusterVersionWithVersioningInfo(allNodesVersioningInfo, currentClusterVersion)
}
```

Maybe you have a better idea there

ereslibre: What I meant was to extract the whole piece:

```go
        currentClusterVersion, err := kubeadm.GetCurrentClusterVersion(client)
	if err != nil {
		return err
	}
	currentVersion := currentClusterVersion.String()
	allNodesVersioningInfo, err := kubernetes.AllNodesVersioningInfo()
	if err != nil {
		return err
	}
	allNodesMatchClusterVersion := 
        kubernetes.AllNodesMatchClusterVersionWithVersioningInfo(allNodesVersioningInfo, currentClusterVersion)
```

In a single function helper call, so `skuba addons upgrade apply` can use it as well.

MaximilianMeister: yes I understood, but we'd have to leave half of it

```go
         currentClusterVersion, err := kubeadm.GetCurrentClusterVersion(client)
	if err != nil {
		return err
	}
	currentVersion := currentClusterVersion.String()
```

because we need it in the `plan` output. so we'd have to fetch the cluster version twice, and extracting would not be beneficial at all. same for `apply` i guess

ereslibre: I see, thanks for the explanation :)

PR #731, opened 2019-09-30
Add man-page for addon upgrade plan

MaximilianMeister:
## What does this PR do?

Add a man page for `skuba addon upgrade plan`

### Related info

Follow up of https://github.com/SUSE/skuba/pull/729


MaximilianMeister:
> Could you maybe add an `EXAMPLE` section? As a user, I always like when that is done in man pages.

yes, me too! should we do it for all commands as a separate task?

PR #733, opened 2019-09-30
Print Addon updates also on cluster upgrade plan

MaximilianMeister:
## Why is this PR needed?

according to the RFC we also need the addon update output on `skuba cluster upgrade plan`

## What does this PR do?

* puts some logic into the addon library
* adds the output also to `cluster upgrade plan`

### DIFF
--- a/pkg/skuba/actions/cluster/upgrade/plan.go
+++ b/pkg/skuba/actions/cluster/upgrade/plan.go
@@ -26,12 +26,9 @@ import (
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
 	"github.com/SUSE/skuba/internal/pkg/skuba/upgrade/addon"
 	upgradecluster "github.com/SUSE/skuba/internal/pkg/skuba/upgrade/cluster"
-	"github.com/SUSE/skuba/pkg/skuba"
 )
 
 func Plan() error {
-	fmt.Printf("%s\n", skuba.CurrentVersion().String())
-
 	client, err := kubernetes.GetAdminClientSet()
 	if err != nil {
 		return err
--- a/pkg/skuba/actions/cluster/upgrade/plan.go
+++ b/pkg/skuba/actions/cluster/upgrade/plan.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubeadm"
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
+	"github.com/SUSE/skuba/internal/pkg/skuba/upgrade/addon"
 	upgradecluster "github.com/SUSE/skuba/internal/pkg/skuba/upgrade/cluster"
 	"github.com/SUSE/skuba/pkg/skuba"
 )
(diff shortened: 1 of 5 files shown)


pkg/skuba/actions/cluster/upgrade/plan.go
```diff
@@ -78,5 +79,12 @@ func Plan() error {
 		}
 	}
 
+	updatedAddons, err := addon.UpdatedAddons(kubernetes.LatestVersion())
```
ereslibre: I think this one should use the immediate next version to the current one instead of the latest one (so they upgrade the platform, then addons, then the platform, then addons... following the QA'ed process).

MaximilianMeister: My suggestion is we should then make a list view for the addon upgrade paths as we do on the cluster version upgrade paths.

We show the latest cluster version in `cluster upgrade plan` as well, but then if we only show the immediate next, we'd hide the latest addon versions.

I think overall a user would expect to see which addon versions will be on the next cluster version update, and also on the latest cluster version. Please have a look if you like this:

```
Upgrade path to update from 1.15.2 to 1.16.4:
 - 1.15.2 -> 1.15.3
  - gangway: 3.1.0 -> 3.2.0 (manifest version from 1 to 3)
 - 1.15.3 -> 1.16.4
  - dex: 2.16.0 -> 2.18.0
  - gangway: 3.1.0 -> 4.1.0 (manifest version from 1 to 5)
```

Only problem (in this example) is that currently we can't infer how to get e.g. from `1_15_3["gangway_version"]` to `1_16_4["gangway_version"]` so we'd always show `currentAddonVersion -> latestAddonVersion` but I think it's still useful

ereslibre: > Only problem (in this example) is that currently we can't infer how to get e.g. from 1_15_3["gangway_version"] to 1_16_4["gangway_version"] so we'd always show currentAddonVersion -> latestAddonVersion but I think it's still useful

I don't think I get this last sentence here, why we cannot infer this information?

MaximilianMeister: > I don't think I get this last sentence here, why we cannot infer this information?

wrong wording, it can be infered but would have required more code changes and added complexity. I've gone back to the suggestion we have talked about now, see latest commit

pkg/skuba/actions/cluster/upgrade/plan.go
```diff
@@ -78,5 +76,19 @@ func Plan() error {
 		}
 	}
 
+	// fetch addon upgrades for the next available cluster version
+	nextClusterVersion := upgradePath[0]
```
chentex: Does it make sense to return an array on line#51 if you are defaulting to just the first option?

MaximilianMeister: yes it's fine, we need the cluster upgrade paths in an array format, to be able to show 2 step upgrades as well (e.g. 1.14.3 -> 1.15.0 -> 1.16.0)

PR #735, opened 2019-10-01
Add verbosity level in man page and skuba help

suseclee:
Detail description about log level in skuba man page
Shows the level of log values in skuba help

## Why is this PR needed?

https://github.com/SUSE/avant-garde/issues/841

`local.PrintDefaults()` does not show default values with `skuba -h`.
```
skuba -h
Usage:
  skuba [command]

Available Commands:
  addon       Commands to handle addons
  auth        Commands to handle authentication
  cluster     Commands to handle a cluster
  help        Help about any command
  node        Commands to handle a specific node
  version     Print version information

Flags:
  -h, --help              help for skuba
  -v, --verbosity Level   number for the log level verbosity [0-10]
```

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/cmd/skuba/main.go
+++ b/cmd/skuba/main.go
@@ -79,6 +79,7 @@ func register(local *pflag.FlagSet, globalName string) {
 	if f := flag.CommandLine.Lookup(globalName); f != nil {
 		pflagFlag := pflag.PFlagFromGoFlag(f)
 		pflagFlag.Name = "verbosity"
+		pflagFlag.Usage = "number for the log level verbosity [0-10]"
 		local.AddFlag(pflagFlag)
 	} else {
 		klog.Fatalf("failed to find flag in global flagset (flag): %s", globalName)
--- a/docs/man/skuba.1.md
+++ b/docs/man/skuba.1.md
@@ -17,7 +17,7 @@ reconfiguration in an easy way.
   Print usage statement.
 
 **-v**
-  set verbosity level.
+  number for the log level verbosity [0-10].
 
 # COMMANDS
 


cmd/skuba/main.go
```diff
@@ -79,7 +79,9 @@ func register(local *pflag.FlagSet, globalName string) {
 	if f := flag.CommandLine.Lookup(globalName); f != nil {
 		pflagFlag := pflag.PFlagFromGoFlag(f)
 		pflagFlag.Name = "verbosity"
-		local.AddFlag(pflagFlag)
+        pflagFlag.Usage = "number for the log level verbosity [0-10]"
```
mjura: Wrong  indentation, can you please fix it

cmd/skuba/main.go
```diff
@@ -79,6 +79,8 @@ func register(local *pflag.FlagSet, globalName string) {
 	if f := flag.CommandLine.Lookup(globalName); f != nil {
 		pflagFlag := pflag.PFlagFromGoFlag(f)
 		pflagFlag.Name = "verbosity"
+		pflagFlag.Usage = "number for the log level verbosity [0-10]"
+		local.PrintDefaults()
```
ereslibre: What is the intent of this line?

suseclee: local.PrintDefaults() shows default value of verbosity level in specific cases when I tested it.
when I  used "skuba -v 1" or 'skuba -h v 2', it showed the default values. 
'skuba -h' does not show the default value.
If you think this is not useful, I can delete it.
Let me know.

docs/man/skuba.1.md
```diff
@@ -17,7 +17,9 @@ reconfiguration in an easy way.
   Print usage statement.
 
 **-v**
-  set verbosity level.
+  set verbosity level [0-10] (default 0).
+  Most of Skuba INFO message appears in the level of 1
+  Regardless of INFO levels, ERROR and WARN messages will appear
```
Klaven: I might make this match the `-h` `number for the log level verbosity [0-10] (default 0)` to be consistent. I think the end goal is to have these man pages generated by the code anyhow and where most of the messages appears would be hard to track and could change easily without updating the man pages.

cmd/skuba/main.go
ereslibre: This call should not be here, because it's conflating the building of the arguments with printing help information about it. I don't see any difference on the output of `skuba -h` and `skuba -v1 -h` with and without this patch in this specific regard (showing the `default` value). On current master:

```
~/p/g/s/g/S/skuba (master) > skuba -h
(snip)
  -v, --verbosity Level   number for the log level verbosity
(snip)
```

```
~/p/g/s/g/S/skuba (master) > skuba -v1 -h
(snip)
  -v, --verbosity Level   number for the log level verbosity (default 0)
(snip)
```

docs/man/skuba.1.md
jenting: How do you know the range is [0-10]?

suseclee: I searched all Skuba code.
These are a few examples for levels.
```
klog.V(0)
No
klog.V(1): Every Skuba
klog.V(1).Infof("=== applying state %s ===", stateName)
klog.V(1).Infof("=== state %s applied successfully ===", stateName)
klog.V(1).Infof("applying %q addon", addon.addon)
klog.V(2) : Vendor
klog.V(2).Infof("fetching Kubernetes version from URL: %s", url)


klog.V(3): vendor
klog.V(3).Infof("Couldn't find the LC_ALL, LC_MESSAGES or LANG environment variables, defaulting to en_US")
klog.V(3).Infof("Skipping service %s due to clusterIP = %q", svcName, service.Spec.ClusterIP)
klog.V(4): vendor
klog.V(4).Infof("Skipping: no address family match for %q on interface %q.", ip, intf.Name)
log.V(4).Infof("Got a Retry-After %ds response for attempt %d to %v", seconds, retries, url)
klog.V(5): vendor
klog.V(5).Infof("Unable to decode an event from the watch stream: %v", err)
klog.V(5).Infof("body was not decodable (unable to check for Status): %v", err)
klog.V(6): Vendor
klog.V(6).Infof("Copying %s", s)
klog.V(6).Infoln("Config loaded from file: ", filename)
klog.V(7): vendor
case bool(klog.V(7)):
   rt = newDebuggingRoundTripper(rt, debugJustURL, debugRequestHeaders, debugResponseStatus)
klog.V(8): vendor
case bool(klog.V(8)):
   max = 1024
}
klog.V(9): vendor
case bool(klog.V(9)):
   rt = newDebuggingRoundTripper(rt, debugCurlCommand, debugURLTiming, debugResponseHeaders)
klog.V(10):vendor
klog.V(10).Infof("ParseInt failed for value %+v in label %+v, %+v", ls.Get(r.key), ls, err)
klog.V(10).Infof("Invalid values count %+v of requirement %#v, for 'Gt', 'Lt' operators, exactly one value is required", len(r.strValues), r)
klog.V(10).Infof("returning cached discovery info from %v", filename)
```

ereslibre: Just to add to the conversation, there is some documentation upstream regarding this:

- https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-output-verbosity-and-debugging
- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

We have been trying to follow upstream approach here to be consistent.

docs/man/skuba.1.md
```diff
@@ -17,7 +17,7 @@ reconfiguration in an easy way.
   Print usage statement.
 
 **-v**
-  set verbosity level.
+  number for the log level verbosity [0-10].
```
dannysauer: Might be nice to display the default level here, maybe in a future PR.

PR #637, opened 2019-08-23
AWS fix deployment

flavio:
## Why is this PR needed?

This PR includes several fixes that are needed to be able to deploy v4 clusters on top of AWS.

The most notable fix is the one for [bsc#1146774](https://bugzilla.suse.com/show_bug.cgi?id=1146774).

## What does this PR do?

Critical fixes:
  * Do not use hard coded AMI IDs: these IDs change over the time and are dependent on the AWS region being used. To solve the problem the terraform files take advantage of a SUSE-dedicated [data source](https://github.com/SUSE/terraform-provider-susepubliccloud). This data source fetches the AMI in a dynamic fashion from a REST API that is owned and maintained by the SUSE Public Cloud team. This is recommended way to proceed
  * Ensure the right AMI is being used:
     - We have to pick the *"container host AMI"* to avoid issues like [bsc#1146774](https://bugzilla.suse.com/show_bug.cgi?id=1146774). This is an AMI built by the SUSE Public Cloud team to act as a container host, due to that it has some optimizations (like the network related one).
     - We have to use a *"BYOS"* (Bring Your Own Subscription) AMI: currently there's no way to register a SLE system against a "Public Cloud team - maintained" RMT instance and then add product channels using a personal registration key. Previous to this PR the code used a "pre-registered" AMI, de-registred it and registered it against SCC/a customer owned RMT server using user credentials. With a BYOS AMI this is no longer needed.
  * Cleanup docker iptable leftovers: **all** the cloud AMIs come with docker running. In addition to removing the docker package we should also cleanup all the iptable rules and table it creates. This is done to avoid issues with kube-proxy later.
  * Fix multi master clusters: the previous code didn't pass terraform planning phase when more than a master was used
  * Remove some cloud-init commands designed for old AWS images: the cri-o package doesn't come with BTRFS hard coded anymore
  * Enable CPI integration
  * Allow cluster VPC to be connected to existing VPCs: this allows the cluster to consume resources hosted by the same organization inside of other VPC (eg: SUSE Manager)
  * Allow registration against SUSE Manager

Minor changes:
  * Move aws keypair definition out of the file defining security group rules.

## Anything else a reviewer needs to know?

* The PR is relatively big, you can have a better time reviewing it at a commit level: you will know exactly why something has been changed
* This is an **unsupported** deployment platform for CaaSP v4 at **this** point in time. The final package we ship doesnt' include any of these files.
[...]

PR #737, opened 2019-10-01
AWS feedback

flavio:
## Why is this PR needed?

This PR addresses the feedback raised on this PR https://github.com/SUSE/skuba/pull/637#issuecomment-535254583 by @rssfed23.

### Be more secure

All the worker nodes are now moved inside of a private subnet. The subnet was already there, we weren't simply using it (dunno why).
This design is more secure and reflects a [common pattern on AWS](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Scenario2.html).

TBH, we could move also the master nodes into the secure network, but then we would need to have a bastion host. This is definitely something for later, once we collect more feedback from users.

### Fix IAM role creation for worker node

@rssfed23 faced the issue. Terraform wasn't creating a IAM role for the worker nodes due to a leftover inside of the state file.

This is not yet going to be exposed to our customers, right now our packages don't have any AWS terraform stuff in them.

## Info for QA

Nothing special to do: just ensure all the worker nodes are inside of the private subnet and the cluster can still be provisioned as expected via skuba.

### Related info

Nothing.

All the worker nodes had a public IP.

Only master nodes have a public IP. Worker nodes are reachable from nodes located inside of the VPC created by terraform.

## Docs

Docs need to be updated, I'm going to file a PR. The contents about AWS are already dated.

### DIFF
--- a/ci/infra/aws/README.md
+++ b/ci/infra/aws/README.md
@@ -11,6 +11,12 @@ this is done by `cloud-init` through a terraform variable called `authorized_key
 All the instances have a `root` and a `ec2-user` user. The `ec2-user` user can
 perform `sudo` without specifying a password.
 
+Only the master nodes have a public IP associated with. All the worker nodes
+are located on a private subnet.
+
+The network structure resembles the one describe inside of this
+[AWS document](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Scenario2.html).
+
 ## Load balancer
 
 The deployment will also create a AWS load balancer sitting in front of the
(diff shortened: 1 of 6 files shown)


ci/infra/aws/README.md
```diff
@@ -11,6 +11,12 @@ this is done by `cloud-init` through a terraform variable called `authorized_key
 All the instances have a `root` and a `ec2-user` user. The `ec2-user` user can
 perform `sudo` without specifying a password.
 
+Only the master nodes have a public IP associated with. All the worker nodes
+are located on a private subnet.
```
ereslibre: Would it make sense to have the master nodes exclusively on the private subnet as well? So the only public IP address is the LB.

flavio: Yes, that would make sense. But it would require the user to create a bastion host, or to have our own terraform stuff handle that...

I would wait a first release to get this change done.

flavio:
I've fixed the linting errors. @ereslibre could you approve it again?

rssfed23:
Just tried it out again on my machine and can confirm all the feedback I raised has been addressed and it's working awesomely and more secure than it was before - LGTM :)

PR #738, opened 2019-10-01
Fix build: `UpdatedAddons` now takes the version of the cluster

ereslibre:
## Why is this PR needed?

`UpdatedAddons` now takes the version of the cluster, pass this
information when calling to `skuba addon upgrade apply`.


ereslibre:
Merging because it's fixing a broken build.

PR #739, opened 2019-10-01
Dynamic node management

tdaines42:
## Why is this PR needed?

Added functions for provisioning and deprovisioning extra nodes.

Fixes https://github.com/SUSE/avant-garde/issues/872

## What does this PR do?

Makes it so we can provision and deprovision nodes without changing the rest of the cluster.

I added the cli commands mostly for testing but also because they could be useful.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/platforms/platform.py
+++ b/ci/infra/testrunner/platforms/platform.py
@@ -25,6 +25,10 @@ class Platform:
         # Files that will be deleted during the cleanup stage
         self.tmp_files = []
 
+    def add_nodes(self, role, count):
+        """Add nodes of the given type"""
+        self._add_nodes(role, abs(count))
+
     @step
     def cleanup(self):
         """Clean up"""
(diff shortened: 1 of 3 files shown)


pablochacin:
I fail to understand why this PR is needed. It introduces a significant complexity to the process of provisioning the platform and seams to be easier ways to accomplish the same. 

If the objective is to allow provisioning nodes which are not part of the initial cluster, there are easier ways to do this by means of the fixtures. or introducing some additional configuration parameters which better control the process of deploying the initial cluster.

Regarding removing nodes from the platform, that can be easily simulated by shooting down the node with `ssh`

I think we should have some very concrete use cases (tests) which are not possible now in order to justify this change.

tdaines42:
> I fail to understand why this PR is needed. It introduces a significant complexity to the process of provisioning the platform and seams to be easier ways to accomplish the same.
> 
> If the objective is to allow provisioning nodes which are not part of the initial cluster, there are easier ways to do this by means of the fixtures. or introducing some additional configuration parameters which better control the process of deploying the initial cluster.
> 
> Regarding removing nodes from the platform, that can be easily simulated by shooting down the node with `ssh`
> 
> I think we should have some very concrete use cases (tests) which are not possible now in order to justify this change.

We don't have unlimited resources for testing. So while we could simulate by shutting down why when we can actually just destroy the node. Also as far as I know once a node has been added and then removed from a cluster it's tainted and not reusable. Unless this has changed I think the better way is to just destroy the no longer usable resource. We may not have other tests aside from basic node removal right now but I'm sure we will eventually.

PR #740, opened 2019-10-02
Init commit unit-test for etcd

c3y1huang:
## Why is this PR needed?
Currently there is no unit-test file for etcd. Add unit-test helps enhance feature hardening.

## What does this PR do?
Added new unit test for etcd.

Test include:
* should remove etcd member from etcd cluster.

* Minor changes to suit unit-test.
* Fixed wrong job name in remove unit-tests which causes job to wait for status until timed out.

## Info for QA

This adds unit-test for etcd.

### Related info

https://github.com/SUSE/avant-garde/issues/900

Unit test code coverage: no test file.

Unit test code coverage: 81.2%

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/etcd/member_test.go
+++ b/internal/pkg/skuba/etcd/member_test.go
@@ -1,0 +1,176 @@
+/*
+ * Copyright (c) 2019 SUSE LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package etcd_test
+
+import (
+	"crypto/sha1"
+	"fmt"
+	"testing"
+
+	"github.com/SUSE/skuba/internal/pkg/skuba/etcd"
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubeadm"
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
+
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubeadmconstants "k8s.io/kubernetes/cmd/kubeadm/app/constants"
+
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func Test_RemoveMember(t *testing.T) {
+	fakeCfgFiles := map[string][]byte{
+		"KubeletComponentConfig": []byte(`
+apiVersion: kubelet.config.k8s.io/v1beta1
+kind: KubeletConfiguration
+`),
+		"ClusterConfiguration": []byte(`
+apiVersion: kubeadm.k8s.io/v1beta1
+kind: ClusterConfiguration
+kubernetesVersion: v1.13.0
+apiServer:
+  extraArgs:
+    advertiseAddress: 1.2.3.4
+`),
+	}
+
+	fakeConfigMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kubeadm-config",
+			Namespace: metav1.NamespaceSystem,
+		},
+		Data: map[string]string{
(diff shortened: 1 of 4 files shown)


pkg/skuba/actions/node/remove/remove_test.go
```diff
@@ -136,9 +137,24 @@ apiServer:
...
+			for _, executer := range tt.executer {
+				shaExecuter = fmt.Sprintf("%x", sha1.Sum([]byte(executer)))
+				tt.clientset.BatchV1().Jobs(metav1.NamespaceSystem).Create(&batchv1.Job{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      fmt.Sprintf("caasp-remove-etcd-member-%.10s-from-%.10s", shaTarget, shaExecuter),
+						Namespace: metav1.NamespaceSystem,
+					},
+					Spec:   jobSpec,
+					Status: batchv1.JobStatus{Active: 1},
+				})
```
innobead: Job mocking here is a indirect way to bypass jobs processing logic and job verification when calling `kubernetes.CreateAndWaitForJob`. 

For now, it is OK at least to make tests not blocking to wait job finished.

internal/pkg/skuba/etcd/member_test.go
```diff
@@ -0,0 +1,176 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      job,
+					Namespace: metav1.NamespaceSystem,
+				},
+				Spec: fakeJobSpec,
+			})
+
+			controlPlaneComponentsVersion, _ := kubeadm.GetCurrentClusterVersion(clientset)
+
+			err := etcd.RemoveMember(clientset, &fakeWorker, controlPlaneComponentsVersion)
```
innobead: This function call is already covered in remove_test.go but it is a public interface, so deserves a unit test.

c3y1huang: Thanks, updated this PR description.

internal/pkg/skuba/etcd/member_test.go
```diff
@@ -0,0 +1,176 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			clientset := fake.NewSimpleClientset(&corev1.NodeList{Items: []corev1.Node{fakeMaster, fakeWorker}})
+
+			clientset.CoreV1().ConfigMaps(metav1.NamespaceSystem).Create(fakeConfigMap)
+			clientset.AppsV1().DaemonSets(metav1.NamespaceSystem).Create(fakeDaemonSet)
+
+			controlPlaneNodes, _ := kubernetes.GetControlPlaneNodes(clientset)
+			hashTarget := fmt.Sprintf("%x", sha1.Sum([]byte(fakeWorker.ObjectMeta.Name)))
+			hashExecutor := fmt.Sprintf("%x", sha1.Sum([]byte(controlPlaneNodes.Items[0].ObjectMeta.Name)))
+			job := fmt.Sprintf("caasp-remove-etcd-member-%.10s-from-%.10s", hashTarget, hashExecutor)
```
ereslibre: We could make `removeMemberFromJobName` public instead of duplicating this logic here.

internal/pkg/skuba/etcd/member_test.go
```diff
@@ -0,0 +1,176 @@
...
+			clientset.BatchV1().Jobs(metav1.NamespaceSystem).Create(&batchv1.Job{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      job,
+					Namespace: metav1.NamespaceSystem,
+				},
+				Spec: fakeJobSpec,
+			})
```
ereslibre: Why are we creating this job beforehand?

c3y1huang: This is mocking the job to bypass processing logic and job verification. Without this the unit-test will wait until timed out when reached `kubernetes.CreateAndWaitForJob` with the fake clientset.

innobead: Like we discussed, this just avoid unit tests running without issues but not really test job verification behavior like `kubernetes.CreateAndWaitForJob`. 

It deserves to enhance later.

innobead: Like we discussed, it can be enhanced later to really test job verification behavior somehow in the future.

c3y1huang: I think maybe just to have a separate unit-test to handle the job processing logic.

internal/pkg/skuba/etcd/member_test.go
c3y1huang: Thanks, I'll enhance in another PR.

ereslibre: Actually, scratch my comment, the function is on the same package (missed this) so we can call it as it is and we don't need to make it public.

internal/pkg/skuba/etcd/member_test.go
ereslibre: +1, thanks!

PR #741, opened 2019-10-02
Support multiple package installation scenarios

pablochacin:
## Why is this PR needed?

There are different test scenarios which are not easily covered without manual setup or extensive changes in the terrarform configuration used by testrunner. Concretely:
* Installing maintenance packages in the nodes
* Installing packages in the nodes from release repositories using registry codes

Fixes https://github.com/SUSE/avant-garde/issues/880

## What does this PR do?
* Introduces a new  section in testrunner configuration file that consolidates all options related to package installation
* Add an option for using a registration code and ignore repositories in the terraform file
* add option for specifying maintenance repositories
* Add option for adding certificates packages to installation, needed when testing self-signed packages
* Improves the logic used for setting repository mirrors

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/README.md
+++ b/ci/infra/testrunner/README.md
@@ -117,15 +117,16 @@ There are some arguments that are currently at the top level of the configuratio
 #### Packages
 The `packages` section configures the source of the packages to be installed in the nodes:
 
-* certificates: package with the certificates to be installed in the nodes in order to install other packages and images. For example, for installing SUSE self-signed packages:
+* additional_pkgs: list with additional packages to be installed in the nodes. For example, for installing SUSE certificates for self-signed packages in development environments:
 ```
 packages:
-  certificates: "ca-certificates-suse"
+  additional_pkgs:
+  - "ca-certificates-suse"
 ```
-* maintenance: repositories to be added to the installation repositories for installing maintenance updates. It takes the form of a map:
+* additional_repos: repositories to be added to the nodes. For example, for installing maintenance updates. It takes the form of a map:
 ```
 packages:
-  maintenance:
+  additional_repos:
     repo1: url/to/repo1
     repo2: url/to/repo2
 ```
--- a/ci/infra/testrunner/README.md
+++ b/ci/infra/testrunner/README.md
@@ -6,6 +6,7 @@
 - [Design](#design)
 - [Configuration](#configuration-parameters)
   - [Work Environment](#work-environment)
+  - [Packages](#packages)
   - [Platform](#platform)
     - [Terraform](#terraform)
     - [Openstack](#openstack)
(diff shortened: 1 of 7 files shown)


ci/infra/testrunner/vars.yaml
```diff
@@ -13,13 +13,16 @@ skuba:
...
-# disable mirrors
-# mirror: "ibs-mirror.prv.suse.net" # mirror url for downloading packages
   plugin_dir: "" # path to the terraform plugin dir
   stack_name: "" # name of the stack
   tfdir: ""  # path to terraform templates
   tfvars: "" # terraform vars
 
+packages:
+# disable mirrors
+# mirror: "ibs-mirror.prv.suse.net" # mirror url for downloading packages
```
tdaines42: Not for this PR but maybe we should look at enabling the provo mirror again.

ci/infra/testrunner/platforms/terraform.py
```diff
@@ -129,11 +130,25 @@ def _update_tfvars(self, tfvars):
                 else:
                     tfvars[k] = v
 
+        # if registry code specified, repositorios are not needed
```
tdaines42: "repositories are not needed"

ci/infra/testrunner/vars.yaml
pablochacin: I will prefer to make it in another PR to avoid this "side effect" not related to the objective of this PR.

ci/infra/testrunner/README.md
```diff
@@ -111,12 +114,29 @@ There are some arguments that are currently at the top level of the configuratio
 - nodeuser: the user name used to login into the platform nodes. Optional. 
 - ssh_key: specifies the location of the key used to access nodes. The default is to use the user's key located at `$HOME/.ssh/id_rsa`
 
+#### Packages
+The `packages` section configures the source of the packages to be installed in the nodes:
+
+* certificates: package with the certificates to be installed in the nodes in order to install other packages and images. For example, for installing SUSE self-signed packages:
+```
+packages:
+  certificates: "ca-certificates-suse"
```
davidcassany: I don't understand why are we using certificates variable instead of something as generic as _package_ or _packages_. If I understood the code properly this `certificates` variable holds a package that will be provisioned by terraform at provisioning time. If so, a user could write something like `certificates: patterns-caasp-Node-1.15` and this would request  `patterns-caasp-Node-1.15` installation. Am I missing something here?

In fact, I'd love to have a way to include a custom set of packages in terraform, this way we could run the installation of several packages in parallel instead of installing them sequentially on each node (during node join) when deploying a cluster.

pablochacin: Is a good point. Nothing prevents that. I can rename it ot something more general. Not to repeat `packages` inside`packages`, can we use `addons`? `extra`?

The same logic applies to the `maintenance` these are just registries. Should we follow the same logic?

davidcassany: `extras` or `additional_packages` (`additional_pkgs` for shorter) sounds good to me :) 

About the `maintenance`, this is related to repositories right? Probably `additional_repos`? I am not that good naming stuff :sweat_smile:

In any case I believe that a more generic name is better if the logic is already generic. Then we could add specific examples to show the intended use case.

That's my view but I am not frontally against the current implementation.

ci/infra/testrunner/README.md
```diff
@@ -111,12 +114,29 @@ There are some arguments that are currently at the top level of the configuratio
...
+```
+* maintenance: repositories to be added to the installation repositories for installing maintenance updates. It takes the form of a map:
+```
+packages:
```
davidcassany: What about 

```
packages:
    extra_repos:
        repo1: url/to/maintenance/repo1
        repo2: url/to/maintenance/repo2
```

I think the intended use case can be just exemplified like that and it is obvious to the user that the underlaying logic is just appending repositories at runtime time. Dos it make sense?

pablochacin: It makes sense. I think this applies for both  `extra_repos` and `extra_packages`.

PR #742, opened 2019-10-02
Add e2e test for skuba addon upgrades

davidcassany:
## Why is this PR needed?

This PR is required to fulfill SUSE/avant-garde#781 and SUSE/avant-garde#782 acceptance criteria of having an end to end test.

## What does this PR do?

This PR adds an e2e test that:

1 deploys a cluster
2 verifies it is up to date
3 manually decreases addons version of the deployed cluster
4 verifies skuba detects available updates
5 upgrades the addons
6 verifies the cluster is up to date again

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


ci/infra/testrunner/tests/test_skuba_addon_upgrade.py
```diff
@@ -0,0 +1,78 @@
...
+        v.major = 0
+        v.minor = 1
+    return str(v)
+
+def test_addons_upgrade(deployment, kubectl, skuba):
+    all_fine = re.compile(
+        r'Congratulations! Addons for \d+\.\d+\.\d+ are already at the latest version available'
+    )
+
+    out = skuba.addon_upgrade('plan')
```
MaximilianMeister: can we have this separated into it's own test? similar to `node upgrade plan` and `node upgrade apply`

davidcassany:
>     1. I'd only separate the `apply` and `plan` each into it's own test, it might be that more plan/apply scenarios will follow so I think we want separation anyways.

I was thinking about it too, but then realized that most of the logic would be equivalent. Also I am not sure what we gain having them separate if we reuse the deployment or would you expect to run a deployment for each test? I thought this would be a waste of cycles and make the test take way longer than it does now.
 
>     2. And please also add a Make target in `ci/Makefile` for the jenkins caller and also add it to the jenkins nightly job list.

Yep, this escaped me, thanks, gonna include it.
 
>     3. as the test currently has lots of business logic, it might be worth to create functions outside the test for some procedures

I was also thinking about it, but the operations that I'd target for functions are likely to be two lines methods. Anyway gonna try to better split it.

MaximilianMeister:
> I was thinking about it too, but then realized that most of the logic would be equivalent. Also I am not sure what we gain having them separate if we reuse the deployment or would you expect to run a deployment for each test? I thought this would be a waste of cycles and make the test take way longer than it does now.

One thing that might be annoying is if something on the `plan` is broken and we run it nightly, we'll never test the `apply` until this is fixed.

I would not mind having another cluster just to test the `plan`, we do the same now in the upgrade tests, and the separation gives a clear summary of what exactly failed. so another benefit is the isolation of potential bugs. besides that we have enough resources to run the tests and time is not a factor if it's a nightly run.

ci/infra/testrunner/kubectl/kubectl.py
```diff
@@ -12,11 +12,11 @@ def __init__(self, conf, platform):
         self.platform = platforms.get_platform(conf, platform)
         self.skuba = Skuba(conf, platform)
 
-    def run_kubectl(self, command):
+    def run_kubectl(self, command, stdin=None):
```
pablochacin: I love this change. will make tests easier by allowing embedding the manifests in the test as string constant. Thanks

pablochacin:
I have nothing to add to @MaximilianMeister commens. Besides that, LGTM. I'll approve once you address that points.

ci/infra/testrunner/tests/test_skuba_addon_upgrade.py
```diff
@@ -0,0 +1,107 @@
...
+    if v.patch > 0:
+        v.patch -= 1
+    elif v.minor > 0:
+        v.minor -= 1
+    elif v.major > 1:
+        v.major -= 1
+    else:
+        v.major = 0
+        v.minor = 1
+    return str(v)
```
ereslibre: An idea, for testing purposes of these commands we could decrease to something like `0.0.1` so we don't need to really perform this decrease logic, after all we can perform the same checks regardless of what previous version we are mimicking, or is there an issue with this approach?

davidcassany: You are right, this can be simplified, thanks!

davidcassany:
@MaximilianMeister I think I addressed all your concerns, thanks for the feedback.

@MaximilianMeister @pablochacin I have not verified the nightly jobs on the CI, is it doable at PR's time? I assume I can't see the pipe changes being applied anywhere until merged right? Sorry I am not that familiar with the Jenkins pipes yet.

pablochacin:
@davidcassany  effectively the jenkins jobs are updated from master, once the PR is merged. 

This PR LGTM. 

I only have one observation which can be addressed in another PR: do we need two jobs? Testrunner (well actually, pytest) can run the two tests independently just pointing to the test suit without specifying a test. As each one does its own cluster allocation they can run without interfering each one. What do you think @tdaines42?

davidcassany:
> I only have one observation which can be addressed in another PR: do we need two jobs? Testrunner (well actually, pytest) can run the two tests independently just pointing to the test suit without specifying a test. As each one does its own cluster allocation they can run without interfering each one.

Interesting I am still learning how the pipe organized. I assume you suggest that I create a single target in the make file pointing to the suite (not specific tests) and then just use that a as job in jenkins. I am fine with it, but I am not sure I see the difference if they run in parallel they should not use the same cluster and if they run one after the other it is basically the same. The tests on Jenkins for this PR revealed that they both use the same cluster, probably because of the job definition I did or probably because I am not using the @pytest.mark.disruptive decorator, I am not sure.

PR #746, opened 2019-10-04
Don't run the collab check on master

tdaines42:
## Why is this PR needed?

The collab check fails on master because there is no change author. 

## What does this PR do?

This skips the collab check if the job is running on master

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/jenkins/pipelines/prs/skuba-code-lint.Jenkinsfile
+++ b/ci/jenkins/pipelines/prs/skuba-code-lint.Jenkinsfile
@@ -14,38 +14,40 @@ pipeline {
 
     stages {
         stage('Collaborator Check') { steps { script {
-            def membersResponse = httpRequest(
-                url: "https://api.github.com/repos/SUSE/skuba/collaborators/${CHANGE_AUTHOR}",
-                authentication: 'github-token',
-                validResponseCodes: "204:404")
+            if (env.BRANCH_NAME.startsWith('PR')) {
+                def membersResponse = httpRequest(
+                    url: "https://api.github.com/repos/SUSE/skuba/collaborators/${CHANGE_AUTHOR}",
+                    authentication: 'github-token',
+                    validResponseCodes: "204:404")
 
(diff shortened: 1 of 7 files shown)


jenting:
What about other branches? Like `release-caasp-4.0-gmc`, it's a hotfix branch, developers would file PR and merged/cherry-pick onto it.

tdaines42:
We aren't running that in CI currently as far as I know. If we decide to then it can be added later.

jenting:
Is it possible to check if it's a PR then run a collaborator check?

jenting:
Two questions to clarify:
1. For the master branch, I saw only 3 pipelines run. Where to set up which pipelines to run? I cannot find in Jenkinsfile
2. Right now all PR needs to rebase onto the target (master) branch, so I don't know why we need to run CI on evey merged commit? IMO, We could move to test the target branch on a nightly test. This will reduce the CI server jobs.

tdaines42:
> Two questions to clarify:
> 
>     1. For the master branch, I saw only 3 pipelines run. Where to set up which pipelines to run? I cannot find in Jenkinsfile
> 
>     2. Right now all PR needs to rebase onto the target (master) branch, so I don't know why we need to run CI on evey merged commit? IMO, We could move to test the target branch on a nightly test. This will reduce the CI server jobs.

Everything is setup using Jenkins Job Builder. 

I think it was just initially set up that way and never changed. It wasn't really hurting anything at the time and runs used to only take ~20 min. We have nightly pipelines running already so we could disable the runs on master if someone wants to do that.

jenting:
I prefer running full pipeline tests (including deployment, bootstrap, join/e2e tests) on every merged commit or running on nightly pipeline. Either way is fine.

innobead:
Following JenTing suggestion, there are 2 things as below. Let’s follow up in other places.

1. Duplicated
2. What checks or tests should be run in PR

PR #756, opened 2019-10-08
Fix skuba addon crash if skuba-config not exist

jenting:
## Why is this PR needed?

When `skuba-config` configmap not exist, run `skuba addon upgrade [plan|apply]` will crash

## What does this PR do?

Add new command output for new addon `addon: X.Y.Z (new addon)`

For example:
Deploying a cluster with skuba version `1.0.2` and built new skuba command on the master branch. Run `skuba addon upgrade plan`, the console output is:
```
Current Kubernetes cluster version: 1.15.2

Addon upgrades for 1.15.2:
  - cilium: 1.5.3 (new addon)
  - kured: 1.2.0 (new addon)
  - dex: 2.16.0 (new addon)
  - gangway: 3.1.0-rev4 (new addon)
  - psp: 1.0.0 (new addon)
```

N/A

## Info for QA

### Related info

N/A

1. Deploy a cluster with skuba version `1.0.2`
2. Built skuba from the master branch
3. Executes `skuba addon upgrade [plan|apply]`, skuba command crash

1. Deploy a cluster with skuba version `1.0.2`
2. Built skuba from the master branch
3. Executes `skuba addon upgrade [plan|apply]`, skuba command won't crash

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/upgrade/addon/version.go
+++ b/internal/pkg/skuba/upgrade/addon/version.go
@@ -46,7 +46,8 @@ func UpdatedAddons(clusterVersion *version.Version) (AddonVersionInfoUpdate, err
 	for addonName, version := range latestAddonVersions {
 		skubaConfigVersion := skubaConfig.AddonsVersion[addonName]
 		aviu.Current[addonName] = skubaConfigVersion
-		if version.Version > skubaConfigVersion.Version || version.ManifestVersion > skubaConfigVersion.ManifestVersion {
+		if skubaConfigVersion == nil ||
+			skubaConfigVersion != nil && (version.Version > skubaConfigVersion.Version || version.ManifestVersion > skubaConfigVersion.ManifestVersion) {
 			aviu.Updated[addonName] = version
 		}
 	}
@@ -55,16 +56,17 @@ func UpdatedAddons(clusterVersion *version.Version) (AddonVersionInfoUpdate, err
 
 func PrintAddonUpdates(updatedAddons AddonVersionInfoUpdate) {
 	for addonName, versions := range updatedAddons.Updated {
-		currentVersion := updatedAddons.Current[addonName].Version
-		currentManifest := updatedAddons.Current[addonName].ManifestVersion
-		updatedVersion := versions.Version
-		updatedManifest := versions.ManifestVersion
+		if updatedAddons.Current[addonName] == nil && updatedAddons.Updated[addonName] != nil {
+			fmt.Printf("  - %s: %s (new addon)\n", addonName, versions.Version)
+			continue
+		}
+
 		hasVersionUpdate := hasAddonVersionUpdateWithAddon(updatedAddons, addonName)
 		hasManifestUpdate := hasAddonManifestUpdateWithAddon(updatedAddons, addonName)
 		if hasVersionUpdate && !hasManifestUpdate {
-			fmt.Printf("  - %s: %s -> %s\n", addonName, currentVersion, updatedVersion)
+			fmt.Printf("  - %s: %s -> %s\n", addonName, updatedAddons.Current[addonName].Version, versions.Version)
 		} else if hasVersionUpdate || hasManifestUpdate {
-			fmt.Printf("  - %s: %s -> %s (manifest version from %d to %d)\n", addonName, currentVersion, updatedVersion, currentManifest, updatedManifest)
+			fmt.Printf("  - %s: %s -> %s (manifest version from %d to %d)\n", addonName, updatedAddons.Current[addonName].Version, versions.Version, updatedAddons.Current[addonName].ManifestVersion, versions.ManifestVersion)
 		}
 	}
 }
@@ -79,9 +81,15 @@ func HasAddonUpdate(aviu AddonVersionInfoUpdate) bool {
 }
 
 func hasAddonManifestUpdateWithAddon(aviu AddonVersionInfoUpdate, addon kubernetes.Addon) bool {
+	if aviu.Current[addon] == nil {
+		return true
+	}
 	return aviu.Updated[addon].ManifestVersion > aviu.Current[addon].ManifestVersion
 }
 
 func hasAddonVersionUpdateWithAddon(aviu AddonVersionInfoUpdate, addon kubernetes.Addon) bool {
+	if aviu.Current[addon] == nil {
+		return true
+	}
 	return aviu.Updated[addon].Version > aviu.Current[addon].Version
 }


internal/pkg/skuba/upgrade/addon/version.go
```diff
@@ -45,6 +45,19 @@ func UpdatedAddons(clusterVersion *version.Version) (AddonVersionInfoUpdate, err
...
 	for addonName, version := range latestAddonVersions {
 		skubaConfigVersion := skubaConfig.AddonsVersion[addonName]
+		if skubaConfigVersion == nil {
+			skubaConfigVersion = &kubernetes.AddonVersion{
+				// Assign current version as updated version and manifest-version to 0 if skuba-config not exist since:
+				//
+				// 1) For better user experience on `skuba addon upgrade plan` if this addon does not updated
+				//    `skuba addon upgrade plan` won't show need upgrade for this addon.
+				//
+				// 2) If addon image version bumped, developer needs to bump manifest version also.
```
davidcassany: This should not be needed if using appropriate schema for tags, see https://github.com/SUSE/avant-garde/issues/905. Am I missing something? If not I would not include this comment here.

jenting: No, this PR is for fixing for the customers who had installed CaaSP v4 with skuba `1.0.2` already.

For the next coming release, we will release a new skuba from the master branch with the tag `X.Y.Z`. The customers updated his/her `skuba` with version `X.Y.Z` and he/she wanna use new skuba command to upgrade cluster or upgrade addon. Then, the customer will bump into this issue by running `skuba addon upgrade [plan|apply]`.

davidcassany: But this won't happen if we use the `x.y.z-revk` tags. I am not concerned about the code logic, just about the comment `2) If addon image version bumped, ...`. I believe it is not needed to update the manifest if the image version bumps. It is just a matter of using the correct tag, isn't it?

jenting: Yes, take a look at Rafa's [comment](https://github.com/SUSE/skuba/pull/726#pullrequestreview-293964654)

davidcassany: Ok I see, then I am fine, thanks for the feedback and details :+1:

internal/pkg/skuba/upgrade/addon/version.go
```diff
@@ -79,9 +81,15 @@ func HasAddonUpdate(aviu AddonVersionInfoUpdate) bool {
...
+		return true
+	}
 	return aviu.Updated[addon].ManifestVersion > aviu.Current[addon].ManifestVersion
 }
 
 func hasAddonVersionUpdateWithAddon(aviu AddonVersionInfoUpdate, addon kubernetes.Addon) bool {
+	if aviu.Current[addon] == nil {
+		return true
+	}
 	return aviu.Updated[addon].Version > aviu.Current[addon].Version
```
ereslibre: I seem to have missed this one on a previous review (!!), this comparison is not right. We are depending on string comparisons here (https://play.golang.org/p/mP5fBhm2_kb), but we should parse the semantic version to a `*version.Version`, and use `Compare` or `LessThan`. We can follow up on this one on another PR if you prefer.

jenting: Already found this issue and I prefer file another PR to address it.

internal/pkg/skuba/upgrade/addon/version.go
```diff
@@ -46,7 +46,8 @@ func UpdatedAddons(clusterVersion *version.Version) (AddonVersionInfoUpdate, err
 	for addonName, version := range latestAddonVersions {
 		skubaConfigVersion := skubaConfig.AddonsVersion[addonName]
 		aviu.Current[addonName] = skubaConfigVersion
-		if version.Version > skubaConfigVersion.Version || version.ManifestVersion > skubaConfigVersion.ManifestVersion {
+		if skubaConfigVersion == nil ||
+			skubaConfigVersion != nil && (version.Version > skubaConfigVersion.Version || version.ManifestVersion > skubaConfigVersion.ManifestVersion) {
```
innobead: As per `(version.Version > skubaConfigVersion.Version || version.ManifestVersion > skubaConfigVersion.ManifestVersion)`, it seems we don't need to update manifest version if only image version updated.

Ref: @ereslibre https://github.com/SUSE/skuba/pull/726#pullrequestreview-293964654

ereslibre: It's more tied to the RFC reasoning than to the actual implementation. You are right, but I prefer if we always bump the manifest version as well (we can trace every possible change to a univocally defined manifest version).

innobead: Understood and make sense to me. Thanks!

jenting: We bump into an issue with version `3.1.0-rev4` is **LessThan** version `3.1.0`, so I think only bump the manifest version could solve this problem.

innobead: This should be resolved by semantic version PR later.

jordimassaguerpla:
@pablochacin : IIUC you are looking into the CI failures. Please sync with @jenting and myself on this PR cause this is blocking us the release of a bug. Thanks.

jenting:
> @pablochacin : IIUC you are looking into the CI failures. Please sync with @jenting and myself on this PR cause this is blocking us the release of a bug. Thanks.

Guess affected by this [PR](https://github.com/SUSE/skuba/commit/7cb1f9557e44e1a79bf58a0f30e26fc39e2ddf8a)

innobead:
For current CI, I don't think it can be rerun without merging/rebasing the fix #761 from master, so I will merge this fix to avoid blocking the release.

internal/pkg/skuba/upgrade/addon/version.go
davidcassany: This looks like a blocker... Damn probably I was just lucky not to hit this issue, I did some test using `3.1.0-rev3` vs `3.1.0-rev4` vs `3.1.0` in the addons e2e tets and they all behaved as expected. I guess this happened because in those tests I did they all sort the same following alphabetical order.

I had the impression the versions were parsed as semantic versions somehow, I just verified that there was a library for such a thing in skuba...

MaximilianMeister: @ereslibre is the *version.Version object converted to a String once you use `>` ?

jenting: Before filing the PR to fix version compare, I would like to discuss our image version (image tag + revision).

Currently, gangway bump version from `3.1.0-rev4` to `3.1.0` and I make some example tests at [here](https://github.com/jenting/semver/blob/master/example_test.go). It seems like the version comparison is not what I thought `3.1.0-rev4` is greater than `3.1.0`. Here is the reference to [semantic versioning comparision](https://semver.org/#spec-item-11)

I come to a workaround on fixing this issue is to compare the manifest version first, and if equals, compare image version.

Any ideas are welcome to come out. @ereslibre @innobead @MaximilianMeister @jordimassaguerpla @davidcassany

davidcassany: Ouch! :disappointed: Yes I fact I remember I was afraid of this behavior and so that I run several tests last week of `skuba addon upgrade plan` with different combinations, however I did not realize that semantic versionig parser was not in use...

davidcassany: How is the new addon thing working? Probably we could force an update of all images and use the revision number for all of them. Then if we always use the schema as `\d+(\.\d+){0,2}-rev\d+` we are safe. I am not a fan of this solution, but is the first that comes to my mind. Still thinking

ereslibre: `semver` has some woes when it comes to this way of [comparing versions](https://semver.org/#spec-item-11), for our use case. Quoting:

> When major, minor, and patch are equal, a pre-release version has lower precedence than a normal version.

> Numeric identifiers always have lower precedence than non-numeric identifiers.

So I propose to do the following, what should always work:

1. Compare `manifestVersion`; if `new.manifestVersion <= old.manifestVersion`, we are done. Either the addon wasn't updated, or we are calling this from an older skuba.

2. If `old.manifestVersion == nil` and `new.manifestVersion != nil`, then it's a new addon.

3. If `new.manifestVersion > old.manifestVersion`, then this addon has an upgrade.

3.1 Compare (string comparison) of versions, no need to parse anything: If `new.AddonVersion != old.AddonVersion` show on the `skuba addon upgrade plan` output the old and new versions, like `%s => %s` with old.AddonVersion and new.AddonVeresion.

3.2 If `new.AddonVersion == old.AddonVersion`, then just show something like `%s (manifest version from %s to %s)` with `new.AddonVersion`, `old.manifestVersion`, `new.manifestVersion`.

So we always depend on the comparison of the int `manifestVersion` to determine whether we have a new version of the addon overall or not, and then just equality string comparisons to show old and new (since we are comparing with our `manifestVersion` we don't have to parse semvers at all, a version on an older manifestVersion **has** to be smaller).

ereslibre: Let me open a PR.

PR #757, opened 2019-10-08
Retry upgraded check more often

MaximilianMeister:
## Why is this PR needed?

The e2e tests are failing sometimes after a successful upgrade of
a node, due to kubernetes being slow updating the NodeVersionInfo.

## What does this PR do?

Double the amount of checks if a node is upgraded.

example of a failed test:
https://ci.suse.de/view/CaaSP/job/caasp-jobs/job/e2e/job/caasp-v4-openstack-test_upgrade_apply_user_lock-nightly/62/console

### DIFF
--- a/ci/infra/testrunner/tests/test_skuba_upgrade.py
+++ b/ci/infra/testrunner/tests/test_skuba_upgrade.py
@@ -31,7 +31,7 @@ def setup_kubernetes_version(platform, skuba, kubectl, kubernetes_version=None):
 
 
 def node_is_upgraded(kubectl, platform, node_name):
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
> The question is _how slow_ is it? We need to define the maximum time we are willing to wait for kubernetes to update the kubelet version info. Try 20 times, doesn't tell me much, unless you know that each time takes 1 seconds, you basically wait for 20 seconds. So, I would prefer a timeout based on time, not re-tries.

usually the last worker node fails on this check, so all nodes seem to be changing within 10 seconds, so I assume 20 should be good enough

PR #759, opened 2019-10-08
Marked addon upgrade tests disruptive

tdaines42:
## Why is this PR needed?

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/tests/test_skuba_addon_upgrade.py
+++ b/ci/infra/testrunner/tests/test_skuba_addon_upgrade.py
@@ -48,6 +48,7 @@ def addons_up_to_date(skuba):
     out = skuba.addon_upgrade('plan')
     return bool(all_fine.findall(out))
 
+@pytest.mark.disruptive
 def test_addon_upgrade_plan(deployment, kubectl, skuba):
     assert addons_up_to_date(skuba)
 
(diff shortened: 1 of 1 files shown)


tdaines42:
@davidcassany If these tests can be fixed then I'd prefer them to keep them enabled for PRs. Maybe it's some other issue that these tests are finding though.

davidcassany:
@tdaines42 I have no clue on what I should fix in them... Let me explain you what they do, since you have a way better understanding of the Jenkins configuration you might see if there is something certainly wrong.

This testsuite includes 2 tests:

1. It hacks the skuba-config configMap of the deployed cluster (that's nasty) to downgrade some addons version, this way when executing `skuba addons upgrade plan` the `versions.go` has a newer version for some addons and assumes there is something to upgrade.

    The hack is not reverted! If some operation is expected to happen within this cluster after this test or during this test (e.g. concurrent tests), this is likely to cause unexpected issues.

2. It hacks the skuba-config configmap of the deployed cluster (that's nasty) to downgrade some addons version, this way when executing `skuba addons upgrade apply` it attemps to redeploy any addons that was previously hacked in skuba-config configmap. After this operation it is expected that the cluster is in a sane position again.

I had assumed that these tests are running sequentially and over an idle cluster. If that is not the case they should not be part of the tests executed within a PR.

Also another side effect of this test is that any other property beyond `skubaConfiguration` inside `skuba-config` configmap will be erased. At the time of writing these tests `skuba-config` had only a single yaml (`skubaConfiguration`) and I haven't seen any change to that so far and I consider it unlikely to change at mid term.

Anyway if it all doesn't ring any bell on your side on what could be fixed/improved I'd say that it is fine to make this test disruptive and then in SUSE/avant-garde#936 scope discuss again if this should be part of the PR tests or not.

I am accepting the PR and leave it up to you and @pablochacin to merge it.

tdaines42:
Looks like the errors from the test were resolved so closing this

PR #760, opened 2019-10-09
Update cilium-agent flags for ipv4/6

Itxaka:
Update to use the current flags to provide the ip version
supported to the cilium agent. the --disable-ipv4 option
is deprecated and hidden since 1.5.x and this other 2
options are now in place instead[0]

Just switch to using those instead

[0] https://github.com/cilium/cilium/commit/6f2218f42513a55821aa6da5dd86157670094bbc

## Why is this PR needed?

Update to current flags and allow more control over ipv6 in cilium

## What does this PR do?

Just updates the cilium-agent flags to the current used ones.
Also removes the check for existing configmap and lets CreateOrUpdateCiliumConfigMap do the proper thing

**I dont know how to test upgrading an existing cluster so I havent done that yet. Looking for advice here**

## Info for QA

Im not sure if this is applicable here, as to trigger the bug related there is a step missing (updating the cri-o version)

### Related info

https://github.com/SUSE/avant-garde/issues/858 led me to create this patch to explicitly disable ipv6, as with a newer cri-o version (1.25.2) some pods were getting dual stack ips and that broke the services, hoping that this will fix it

We passed the --disable-ipv4 flag as a value in the cilium configmap

we pass the --enable-ipv4 and --enable-ipv6 flag as a values in the cilium configmap

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/addons/cilium.go
+++ b/internal/pkg/skuba/addons/cilium.go
@@ -60,20 +60,15 @@ func (ciliumCallbacks) beforeApply(addonConfiguration AddonConfiguration, skubaC
 	if err != nil {
 		return errors.Wrap(err, "unable to determine if cilium secret exists")
 	}
-	ciliumConfigMapExists, err := cni.CiliumConfigMapExists(client)
-	if err != nil {
-		return errors.Wrap(err, "unable to determine if cilium config map exists")
-	}
 	if !ciliumSecretExists {
 		if err := cni.CreateCiliumSecret(client); err != nil {
 			return err
 		}
 	}
-	if !ciliumConfigMapExists {
-		if err := cni.CreateOrUpdateCiliumConfigMap(client); err != nil {
-			return err
-		}
+	if err := cni.CreateOrUpdateCiliumConfigMap(client); err != nil {
+		return err
 	}
+
 	return nil
 }
 
--- a/internal/pkg/skuba/addons/cilium.go
+++ b/internal/pkg/skuba/addons/cilium.go
@@ -191,7 +191,8 @@ spec:
           - "-t=vxlan"
           - "--kvstore=etcd"
           - "--kvstore-opt=etcd.config=/var/lib/etcd-config/etcd.config"
-          - "--disable-ipv4=$(DISABLE_IPV4)"
+          - "--enable-ipv4=$(ENABLE_IPV4)"
+          - "--enable-ipv6=$(ENABLE_IPV6)"
           - "--container-runtime=crio"
         ports:
           - name: prometheus
@@ -211,11 +212,16 @@ spec:
               configMapKeyRef:
                 name: cilium-config
                 key: debug
-          - name: "DISABLE_IPV4"
+          - name: "ENABLE_IPV4"
             valueFrom:
               configMapKeyRef:
                 name: cilium-config
-                key: disable-ipv4
+                key: enable-ipv4
+          - name: "ENABLE_IPV6"
+            valueFrom:
+              configMapKeyRef:
+                name: cilium-config
+                key: enable-ipv6
         livenessProbe:
           exec:
             command:
--- a/internal/pkg/skuba/cni/cilium.go
+++ b/internal/pkg/skuba/cni/cilium.go
@@ -125,9 +125,10 @@ func CreateOrUpdateCiliumConfigMap(client clientset.Interface) error {
 		return err
 	}
 	ciliumConfigMapData := map[string]string{
-		"debug":        "false",
-		"disable-ipv4": "false",
-		"etcd-config":  string(etcdConfigDataByte),
+		"debug":       "false",
+		"enable-ipv4": "true",
+		"enable-ipv6": "false",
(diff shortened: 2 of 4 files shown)


Itxaka:
tested and this definitively fixes the problem with the dual stack pods in cri-o 1.15.2

Itxaka:
@ereslibre seems a bit difficult to patch this upstream as from version 1.6, they already have a way of overriding this values using charts and the daemon does not need to have the flags explicitly passed, but having them in the configmap would be good enough for the agent to pick them up. 

And actually for us, we could drop some of the flags being passed as it would pick them from the configmap directly if Im reading this correctly: https://docs.cilium.io/en/v1.5/kubernetes/configuration/

internal/pkg/skuba/cni/cilium.go
```diff
@@ -125,9 +125,10 @@ func CreateOrUpdateCiliumConfigMap(client clientset.Interface) error {
 		return err
 	}
 	ciliumConfigMapData := map[string]string{
-		"debug":        "false",
-		"disable-ipv4": "false",
-		"etcd-config":  string(etcdConfigDataByte),
+		"debug":       "false",
+		"enable-ipv4": "true",
+		"enable-ipv6": "false",
+		"etcd-config": string(etcdConfigDataByte),
```
ereslibre: Please, bump the manifest version of `Cilium` in the `AddonsVersion`, so `skuba addon upgrade` will know that the manifests changed and that it has to reapply them: https://github.com/SUSE/skuba/blob/master/internal/pkg/skuba/kubernetes/versions.go#L79-L138, it can be done on all Kubernetes versions.

Itxaka: ahh perfect, that was what I was looking for regarding the update part, nice!

Itxaka: done!

Itxaka:
is there any reason to check if the configmap exists when we have a method called `CreateOrUpdateCiliumConfigMap` which should do the right thing? At least according to the upstream function: https://github.com/kubernetes/kubernetes/blob/v1.15.2/cmd/kubeadm/app/util/apiclient/idempotency.go#L45

Itxaka:
Also this may indicate that the end2end test for `addon upgrade apply` is not fully covering changes like this?

ereslibre:
I don't think there is a real reason for those checks to be performed aside from being a little faster, now with `skuba addon upgrade apply` we really need to reapply (it's fine if we skip the secret if it's already there, but the configmap check should be gone).

internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -89,7 +89,7 @@ var (
 				ToolingVersion:          "0.1.0",
 			},
 			AddonsVersion: AddonsVersion{
-				Cilium:  &AddonVersion{"1.5.3", 0},
+				Cilium:  &AddonVersion{"1.5.3", 1},
```
ereslibre: Can this be bumped on all Kubernetes versions? I know it's a little repetitive, but here we are favouring expliciteness.

ereslibre:
> Also this may indicate that the end2end test for addon upgrade apply is not fully covering changes like this?

There is a number of relevant tests that happen at the PR, we have to find the right balance between speedy PR feedback and a reasonable number of tests covering the PR level. It's fine if everything looks green at the PR level (running a subset of e2e tests), as long as we detect the problem on nightly jobs, and we have a plan to address those as soon as possible when they break.

Itxaka:
> I don't think there is a real reason for those checks to be performed aside from being a little faster, now with skuba addon upgrade apply we really need to reapply (it's fine if we skip the secret if it's already there, but the configmap check should be gone).

Sound good, do we want that change in here as a different commit or in a different PR?

> There is a number of relevant tests that happen at the PR, we have to find the right balance between speedy PR feedback and a reasonable number of tests covering the PR level. It's fine if everything looks green at the PR level (running a subset of e2e tests), as long as we detect the problem on nightly jobs, and we have a plan to address those as soon as possible when they break.

Good point!

internal/pkg/skuba/kubernetes/versions.go
Itxaka: done!

internal/pkg/skuba/addons/cilium.go
```diff
@@ -60,20 +60,15 @@ func (ciliumCallbacks) beforeApply(addonConfiguration AddonConfiguration, skubaC
 	if err != nil {
 		return errors.Wrap(err, "unable to determine if cilium secret exists")
 	}
-	ciliumConfigMapExists, err := cni.CiliumConfigMapExists(client)
```
ereslibre: This function can be removed as well now that it has no callers.

Itxaka: good catch, removed!

jenting:
Please run `make lint` to fix linter error

PR #763, opened 2019-10-10
correct return error

Hui-Zhi:
## Why is this PR needed?
I think it should only return when err is not nil, please reject this if it's just the author's coding style.

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


ereslibre:
However, it's strange that we are doing a `return err` there (when we know that `err == nil`). We should probably instead return a real error, something like:

```go
if err == nil {
  return errors.Errorf("[join] failed to join the node with name %q since a node with the same name already exists in the cluster\n", target.Nodename)
}

**Comment by ereslibre** [comment]
I think this was refactored to use regular `fmt` and return to just show the error to the user, but this will not properly handle skuba's exit code (that will be 0 I believe, because no error was raised, only an error was shown to the user on stdout).

We should probably audit how we are doing this in the codebase regarding this toplevel actions so we are consistent (e.g. `fmt` vs `klog` and also regarding exit status codes).

**Comment by ereslibre** [comment]
@Klaven WDYT?

**Comment by Hui-Zhi** [comment]
Ping @Klaven

File: pkg/skuba/actions/node/join/join.go [hunk 1]
```diff
@@ -83,10 +83,10 @@ func Join(joinConfiguration deployments.JoinConfiguration, target *deployments.T
 		"skuba-update.start",
 	}
 
+	// Try to find exist node which have the same name.
 	_, err = client.CoreV1().Nodes().Get(target.Nodename, metav1.GetOptions{})
 	if err == nil {
-		fmt.Printf("[join] failed to join the node with name %q since a node with the same name already exists in the cluster\n", target.Nodename)
-		return err
+		return errors.Errorf("[join] failed to join the node with name %q since a node with the same name already exists in the cluster\n", target.Nodename)
```
ereslibre:
```suggestion
		return errors.Errorf("[join] failed to join the node with name %q since a node with the same name already exists in the cluster", target.Nodename)
```

Hui-Zhi:
> Can we double check if we are consistent with other shown errors? I think some of them are using `fmt`, would you mind double checking and seeing if we can compose a PR that involves the rest as well to grasp it?

Sure.

jordimassaguerpla:
Hi!
Happy hacking

chentex:
Wow this has been open for too long, @Hui-Zhi can you `squash` this.

chentex:
Original repository does not exist anymore, this was more a nitpick. never driven forward.

PR #764, opened 2019-10-10
Avoid race condition on deploying lb with libvirt

Itxaka:
## Why is this PR needed?

Fixes an issue that can arise when manually deploying the terraform templates with libvirt

Fixes #

N/A

## What does this PR do?

There seems to be a race condition on the 2 resources
of the libvirt lb (reboot, haproxy) which are launched
inmediatly after the cloud init resource has finished
and can lead to the haproxy resource never completing
because the machine rebooted.

Instead, make the reboot dependant on the haproxy
resource, so its done in a linear fashion.

N/A

## Info for QA
N/A

### Related info
N/A

Creating libvirt resources with the terraform templates could lead to a timeout of the lb_push_haproxy_cfg resource

Creating libvirt resources with the terraform template will never timeout the lb_push_haproxy_cfg resource

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/libvirt/lb-instances.tf
+++ b/ci/infra/libvirt/lb-instances.tf
@@ -148,7 +148,7 @@ resource "null_resource" "lb_push_haproxy_cfg" {
 }
 
 resource "null_resource" "lb_reboot" {
-  depends_on = ["null_resource.lb_wait_cloudinit"]
+  depends_on = ["null_resource.lb_push_haproxy_cfg"]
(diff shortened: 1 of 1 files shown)


Itxaka:
oh, this seems to be fixed in a different manner here: https://github.com/SUSE/skuba/pull/758

PR #766, opened 2019-10-10
Handle RuntimeError in deployment fixture

pablochacin:
## Why is this PR needed?

This fixture joins all nodes and then waits until
the system pods are ready. In some circumstances,
some of the nodes may not respond and this causes
a RuntimeError which should be retried.

## What does this PR do?

Add the RuntimeError as a retriable exception to the wait function that checks for pod availability

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/tests/conftest.py
+++ b/ci/infra/testrunner/tests/conftest.py
@@ -50,7 +50,7 @@ def deployment(request, bootstrap, skuba, kubectl):
     if request.config.getoption("skip_setup") != 'deployed':
         skuba.join_nodes()
 
-    wait(kubectl.run_kubectl, 'wait --timeout=1m --for=condition=Ready pods --all --namespace=kube-system', wait_delay=60, wait_timeout=300, wait_backoff=30, wait_retries=5)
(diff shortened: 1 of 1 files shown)


ci/infra/testrunner/tests/conftest.py
```diff
@@ -50,7 +50,7 @@ def deployment(request, bootstrap, skuba, kubectl):
     if request.config.getoption("skip_setup") != 'deployed':
         skuba.join_nodes()
 
-    wait(kubectl.run_kubectl, 'wait --timeout=1m --for=condition=Ready pods --all --namespace=kube-system', wait_delay=60, wait_timeout=300, wait_backoff=30, wait_retries=5)
+    wait(kubectl.run_kubectl, 'wait --timeout=1m --for=condition=Ready pods --all --namespace=kube-system', wait_delay=60, wait_timeout=300, wait_backoff=30, wait_retries=5, wait_allow=(RuntimeError))
```
MaximilianMeister: can we just use `kubectl wait` with a timeout of 5m here? i don't see the need to have a nested `wait` procedure with 2 different waiting mechanisms, it looks a bit overcomplex to me

MaximilianMeister: this was explained in https://github.com/SUSE/skuba/pull/768

PR #768, opened 2019-10-11
Wait timeout enhancements

MaximilianMeister:
## Why is this PR needed?

waiting in e2e tests is not consistent

## What does this PR do?

make waiting consistent

### DIFF
--- a/ci/infra/testrunner/tests/test_cilium.py
+++ b/ci/infra/testrunner/tests/test_cilium.py
@@ -13,7 +13,7 @@ def test_cillium(deployment, kubectl):
 
     logger.info("Deploy deathstar")
     kubectl.run_kubectl("create -f https://raw.githubusercontent.com/cilium/cilium/v1.5/examples/minikube/http-sw-app.yaml")
-    kubectl.run_kubectl("wait --for=condition=ready pods --all --timeout=3m")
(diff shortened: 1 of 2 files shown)


ci/infra/testrunner/tests/conftest.py
```diff
@@ -50,7 +50,7 @@ def deployment(request, bootstrap, skuba, kubectl):
     if request.config.getoption("skip_setup") != 'deployed':
         skuba.join_nodes()
 
-    wait(kubectl.run_kubectl, 'wait --timeout=1m --for=condition=Ready pods --all --namespace=kube-system', wait_delay=60, wait_timeout=300, wait_backoff=30, wait_retries=5)
+    kubectl.run_kubectl('wait --timeout=5m --for=condition=Ready pods --all --namespace=kube-system')
```
Itxaka: FYI for me end2end tests were consistently failing (on test_reboot) and were fixed by the introduction of the wait in here by https://github.com/SUSE/skuba/commit/5e5d4bb258289747a39fdf23e758c22d782946b2 as it looked like the `kubectl wait` was stuck on some pods which were already on `Ready` status, but the command hanged there and turned into an error. Im sure the `wait_retries` helped here.

MaximilianMeister: Ok, thanks for the hint

PR #769, opened 2019-10-11
addon upgrades: only use manifest version to determine upgrade

ereslibre:
## Why is this PR needed?

When determining when an addon has to be upgraded (using `skuba addon
upgrade apply` or `skuba addon upgrade plan`), rely on the manifest
version exclusively, and if different, treat the addon version as a
regular string.

If the manifest version is always bumped with every single change,
necessarily it has to be a newer version of the addon, so no need to
perform a semver parsing of the actual version of the addon (that may
contain semver pre-releases suffixes, or metadata builds -- ignored
when comparing, or not deeply compared; e.g. any prerelease
information < no prerelease information, since it's considered
prerelease vs release [1]).

[1] https://semver.org/#spec-item-11

### DIFF
--- a/internal/pkg/skuba/upgrade/addon/version.go
+++ b/internal/pkg/skuba/upgrade/addon/version.go
@@ -46,8 +46,7 @@ func UpdatedAddons(clusterVersion *version.Version) (AddonVersionInfoUpdate, err
 	for addonName, version := range latestAddonVersions {
 		skubaConfigVersion := skubaConfig.AddonsVersion[addonName]
 		aviu.Current[addonName] = skubaConfigVersion
-		if skubaConfigVersion == nil ||
-			skubaConfigVersion != nil && (version.Version > skubaConfigVersion.Version || version.ManifestVersion > skubaConfigVersion.ManifestVersion) {
+		if skubaConfigVersion == nil || (version.ManifestVersion > skubaConfigVersion.ManifestVersion) {
 			aviu.Updated[addonName] = version
 		}
 	}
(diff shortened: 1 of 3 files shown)


internal/pkg/skuba/upgrade/addon/version.go
```diff
@@ -61,35 +60,22 @@ func PrintAddonUpdates(updatedAddons AddonVersionInfoUpdate) {
 			continue
 		}
 
-		hasVersionUpdate := hasAddonVersionUpdateWithAddon(updatedAddons, addonName)
-		hasManifestUpdate := hasAddonManifestUpdateWithAddon(updatedAddons, addonName)
-		if hasVersionUpdate && !hasManifestUpdate {
+		// At this point we know that this addon has a greater manifest version, if the addon version
+		// is different than the one stored, we will show that to the user. If the versions are equals
+		// (string comparison), we will show the manifest version bump as additional information.
+		if updatedAddons.Current[addonName].Version != versions.Version {
```
jenting: I think it should check manifest version first. Otherwise, it encounter user1 and user2 with different skuba version, always print out an addon need to upgrade

jenting:
Comment left, thanks Rafa.

internal/pkg/skuba/upgrade/addon/version.go
ereslibre: Plugins that are inside the `updatedAddons.Updated` structure are the ones that already had a newer `ManifestVersion`. I could make the explicit check here as well, but it would be redundant.

jenting: okay. I see it

PR #771, opened 2019-10-14
Replaced shelling out of kubectl drain command.

c3y1huang:
## Why is this PR needed?

Currently uses shelling out to kubectl to remove node. kubernetes/kubernetes#72827 can help to drain using kubectl as a library.

## What does this PR do?

Replace shelling out to `kubectl drain` with kubectl library

There are no change to behavior.

## Info for QA

No update in unit-test, code already covered by remove-node.

### Related info
* https://github.com/SUSE/avant-garde/issues/850
* https://github.com/kubernetes/kubernetes/pull/72827

Replace command with lib, there are no change in behavior.

Replace command with lib, there are no change in behavior.

## Docs

Replace command with lib, there are no change in behavior.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


pkg/skuba/actions/node/remove/remove.go
```diff
@@ -67,7 +67,9 @@ func Remove(client clientset.Interface, target string, drainTimeout time.Duratio
 		fmt.Printf("[remove-node] removing worker node %s (drain timeout: %s)\n", targetName, drainTimeout.String())
 	}
 
-	kubernetes.DrainNode(node, drainTimeout)
+	if err := kubernetes.DrainNode(client, node, drainTimeout); err != nil {
+		return errors.Wrapf(err, "[remove-node] could not drain node")
```
jenting: Use Wrap

internal/pkg/skuba/kubernetes/nodes.go
```diff
@@ -61,18 +59,47 @@ func IsControlPlane(node *v1.Node) bool {
...
+	drainer := &kubectldrain.Helper{
+		Client:              client,
+		Force:               true,
+		IgnoreAllDaemonSets: true,
+		DeleteLocalData:     true,
+		Timeout:             drainTimeout,
 	}
+	del, errs := drainer.GetPodsForDeletion(node.ObjectMeta.Name)
+	if errs != nil {
+		return errors.New(fmt.Sprintf("could not get pods for deletion %s", errs))
```
jenting: Consider use `fmt.Errorf`

pkg/skuba/actions/node/remove/remove.go
jenting: No argument, so `errors.Wrap`

internal/pkg/skuba/kubernetes/nodes.go
```diff
@@ -61,18 +59,47 @@ func IsControlPlane(node *v1.Node) bool {
...
+	}
+	for _, pod := range del.Pods() {
+		if len(policyGroupVersion) > 0 {
```
jenting: All pods in the same node should have the same eviction capability?

If yes, check `len(policyGroupVersion) > 0` before the loop `for _, pod := range del.Pods()` would have better performance.

jenting:
Link to https://github.com/SUSE/avant-garde/issues/854

internal/pkg/skuba/kubernetes/nodes.go
c3y1huang: I prefer to keep it within the loop. There is a fail prove to delete pod after eviction in case eviction doesn't work.

internal/pkg/skuba/kubernetes/nodes.go
c3y1huang: :+1:

pkg/skuba/actions/node/remove/remove.go
c3y1huang: :+1:

internal/pkg/skuba/kubernetes/nodes.go
c3y1huang: On the second look, I think you are right. I will change it. Thanks.

jordimassaguerpla:
> This is great @c3y1huang, thank you!
> 
> LGTM, deferring merge to @jordimassaguerpla, since it's not strictly a new feature I think we can wait to merge this one right after we tag a new version (to not include that many changes on the immediate next release).

@cclhsu : Please bring this at the next release coordination meeting.

PR #772, opened 2019-10-14
Enhance e2e tests for addon upgrade

davidcassany:
## Why is this PR needed?

This PR enhances the tests around `skuba addons upgrade plan|apply` to better align with the latest changes in skuba logic (#770).

## What does this PR do?

This commit enhances the e2e tests for addons to cover three different situations:

1. There is a new addon
2. There is a new manifest for an existing addon
3. There is a new container image for an exsisting addon

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/tests/test_skuba_addon_upgrade.py
+++ b/ci/infra/testrunner/tests/test_skuba_addon_upgrade.py
@@ -1,15 +1,18 @@
 import pytest
 import yaml
 import re
+import copy
 from io import StringIO
 
-
-def replace_skuba_config(kubectl, configmap_data):
-    new_configmap = kubectl.run_kubectl(
-        'create configmap skuba-config --from-literal {0} -o yaml --namespace kube-system --dry-run'.format(
-            configmap_data
+def create_skuba_config(kubectl, configmap_data, dry_run=False):
+    return kubectl.run_kubectl(
+        'create configmap skuba-config --from-literal {0} -o yaml --namespace kube-system {1}'.format(
+            configmap_data, '--dry-run' if dry_run else ''
         )
     )
+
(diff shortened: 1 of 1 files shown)


jenting:
CI failed due to test deployment timeout. This [PR](https://github.com/SUSE/skuba/pull/773/files) seems like to address the issue.

jenting:
Maybe we need to consider this case also https://github.com/SUSE/skuba/pull/760#issuecomment-542625777

davidcassany:
> Maybe we need to consider this case also [#760 (comment)](https://github.com/SUSE/skuba/pull/760#issuecomment-542625777)

I am not sure this is actually a new case from addons upgrade POV. If I understood the issue the failure is happening because of bad manifest upgrade, isn't it? The tests here are thought to validate the generic portions of addon apply and addon plan (not depending on addons specifics). 

Those are not regression tests for `skuba addon upgrade plan|apply` that any specific addon upgrade should relay on. To do that and test real addons upgrades (thus any addon upgrade commit would go through it) some refactor on Jenkins and/or testrunner is required. I can't see another way to test that without:

1. deploy a cluster with a skuba version before the patch
2. update skuba after the patch
3. run skuba upgrade commands

This can't be easily achieved with the current testrunner and Jenkins pipes. I would go for if I see a simple way to achieve that actually makes sense for automated Jenkins based executions and local executions.

Said all that I'd be more than happy to improve the coverage of these tests, so if you clearly see how this could be done feel free to share it or even commit over this branch :)

PR #773, opened 2019-10-14
Unify timeout handling in tests

pablochacin:
## Why is this PR needed?

Presently different tests using different approaches for handling timeouts when executing commands in the nodes, including those executed with `ssh and `kubectl`. 

All this diversity of mechanism makes hard to reason about how long are tests actually waiting before failing, making difficult to debug the problems. 

Fixes # https://github.com/SUSE/avant-garde/issues/939

## What does this PR do?

Unify the timeout handling in tests using the wait function.
Use --timeout=0 option for kubectl wait calls to return immediately
and use the backoff/retry mechanism in the wait function.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


MaximilianMeister:
by customized query i mean `kubectl get pods -o jsonpath` e.g. - this might be more reliable

pablochacin:
@MaximilianMeister I agree but I think this change should be a different PR just to keep this scope of this PR around handling the timeouts.

pablochacin:
@MaximilianMeister at the end I followed your suggestion of checking for the pod status because using the `wait` command caused some issues (basically, it generated an exception in the `run_shellcommand` function which was not properly handled) .

ci/infra/testrunner/tests/test_cilium.py
```diff
@@ -13,7 +15,14 @@ def test_cillium(deployment, kubectl):
 
     logger.info("Deploy deathstar")
     kubectl.run_kubectl("create -f https://raw.githubusercontent.com/cilium/cilium/v1.5/examples/minikube/http-sw-app.yaml")
-    kubectl.run_kubectl("wait --for=condition=ready pods --all --timeout=3m")
+
+    # FIXME: This check should be only get the star wars application's pods
```
Klaven: not sure I understand this, or why we can't fix it before merging this.

ci/infra/testrunner/tests/conftest.py
```diff
@@ -38,19 +39,33 @@ def cleanup():
...
 @pytest.fixture
 def deployment(request, bootstrap, skuba, kubectl):
     if request.config.getoption("skip_setup") != 'deployed':
         skuba.join_nodes()
 
-    wait(kubectl.run_kubectl, 'wait --timeout=5m --for=condition=Ready pods --all --namespace=kube-system', wait_delay=60, wait_timeout=300, wait_backoff=30, wait_retries=5, wait_allow=(RuntimeError))
+    wait(check_system_pods_ready,
+         kubectl,
+         wait_delay=60,
+         wait_timeout=10,
```
Klaven: I would prefer if we did not change the `wait_timeout` this pass as that is not what this PR is about. same with `wait_backoff` `wait_retires` and `wait_allow` it just seems like these changes are not related.

PR #774, opened 2019-10-14
Terraform 0.12 compatibility

bergmannf:
## Why is this PR needed?

I migrated the configuration files for terraform to version 0.12.

This also needs a more up-to-date version of the openstack provider: 1.23.0

The VMWare provider that worked with `vsphere` was using version: 1.13.0

## What does this PR do?

It migrates our terraform configuration files tot he new format - this is not backward compatible to version 0.11


ci/infra/openstack/variables.tf
```diff
@@ -70,8 +71,9 @@ variable "dnsdomain" {
 }
 
 variable "dnsentry" {
-  default     = 0
+  default     = false
```
mjura: Please also update `terraform.tfvars.example`

ci/infra/openstack/variables.tf
```diff
@@ -55,8 +55,9 @@ variable "workers" {
 }
 
 variable "workers_vol_enabled" {
-  default     = 0
+  default     = false
```
mjura: Please update also `terraform.tfvars.example`

bergmannf:
One thing I realized, when attempting to use the `testrunner` with this branch: this PR only works, when directly running `terraform apply`.

When using `terraform plan > some.tfplan` and `terraform apply some.tfplan` it will break (which is strange, but reproducible).

pablochacin:
@bergmannf That's weird. I'm wondering if we need to do these two steps in testrunner. :thinking:

bergmannf:
Newer up-to-date set of changes for terraform 0.12 is now in #872 .

As this also reintroduced the `template_file` resource again it should be preferred as it doesn't need the extraction into `locals`.

PR #776, opened 2019-10-14
Fix bsc#1153913

flavio:
## Why is this PR needed?

Do not try to guess the kubeconfig file to be used. The code slipped into by mistake (it was a POC).

Fixes [bsc#1153913](https://bugzilla.suse.com/show_bug.cgi?id=1153913)

## What does this PR do?

Go back to the old behaviour and rely on a hard-coded value.

Nothing.

## Info for QA

Just ensure you can do a `skuba node bootstrap` even if you have a `~/.kube/config` file inside on your machine.

Node bootstrap doesn't work. See bug report.

Node bootstrap works as expected.

## Docs

Nothing to change.

### DIFF
--- a/pkg/skuba/constants.go
+++ b/pkg/skuba/constants.go
@@ -19,7 +19,6 @@ package skuba
 
 import (
 	"fmt"
-	"os"
 	"path"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


bergmannf:
I just tested the changes locally - this fixes the problem.

It only happens during `bootstrap`, because at that time not `admin.conf` exists yet - maybe this might be used to workaround this problem in the future.

PR #777, opened 2019-10-14
Fix addon upgrade breaking generated manifests by invalid control plane value

innobead:
## Why is this PR needed?

Fix broken generated addon manifests from `skuba addon upgrade apply` caused by invalid control plane value.

Fixes https://github.com/SUSE/avant-garde/issues/944

## What does this PR do?

Use the correct control plane value w/o port to generate correct addon manifests, but this fix doesn’t handle customized manifest changed by user that should be tackled by another PR based on kustomize based on a RFC proposal.

N/A

## Info for QA

Please just follow up the steps mentioned in https://github.com/SUSE/avant-garde/issues/944 to verify gangway addon upgrade w/ valid gangway configmap, but this does not handle customized manifest changed by user.

### Related info
n/a

After applying gangway addon upgrade, it's config map should has invalid URLs.

```
redirectURL: "https://10.84.153.111:6443:32001/callback"
authorizeURL: "https://10.84.153.111:6443:32000/auth"
tokenURL: "https://10.84.153.111:6443:32000/token"
apiServerURL: "https://10.84.153.111:6443:6443"
```

After applying gangway addon upgrade, it's config map should has valid URLs.

```
redirectURL: "https://10.84.153.111:32001/callback"
authorizeURL: "https://10.84.153.111:32000/auth"
tokenURL: "https://10.84.153.111:32000/token"
apiServerURL: "https://10.84.153.111:6443"
```

## Docs
n/a

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/pkg/skuba/actions/addon/upgrade/apply.go
+++ b/pkg/skuba/actions/addon/upgrade/apply.go
@@ -25,6 +25,7 @@ import (
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
 	"github.com/SUSE/skuba/internal/pkg/skuba/upgrade/addon"
 	"github.com/pkg/errors"
+	kubeadmutil "k8s.io/kubernetes/cmd/kubeadm/app/util"
 )
 
 // Apply implements the `skuba addon upgrade apply` command.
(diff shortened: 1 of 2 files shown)


jenting:
It’s not covers when user changes other manifest setting. The solution should involves https://github.com/kubernetes-sigs/kustomize.

innobead:
Yes, that should be tackled by another implementation by the proposal RFC from Rafa first. This only fixes the incorrect control plane value.

thehejik:
Now it works correctly, thanks for the quick fix

jenting:
> I think this is hiding potentially important information for addons. IMO we should pass the raw control plane endpoint, but each addon should then choose if they will use the full URI or parse it and get the host only, or if they will use the control plane endpoint port as well for anything.

`skuba cluster init` render with control-plane flag
https://github.com/SUSE/skuba/blob/master/pkg/skuba/actions/cluster/init/init.go#L127
https://github.com/SUSE/skuba/blob/master/internal/pkg/skuba/addons/addons.go#L213

`skuba node bootstrap` and `skuba addon upgrade apply` render with control-plane-endpoint
https://github.com/SUSE/skuba/blob/master/internal/pkg/skuba/addons/addons.go#L123
https://github.com/SUSE/skuba/blob/master/internal/pkg/skuba/addons/addons.go#L236

So, this means we need an extra field to records the control-plane-endpoint. Good?

ereslibre:
Created https://github.com/SUSE/skuba/pull/781 as an alternative, I'm going to validate if it works as expected and remove the WIP flag.

innobead:
Sorry for the late reply. Let Rafa continues the fixing, closing this.

PR #779, opened 2019-10-15
Add validator job

MaximilianMeister:
## Why is this PR needed?

We need to run validator on a nightly base, and on demand

## What does this PR do?

Add a job to run a simple validator scenario

Pipeline and codebase is here https://gitlab.suse.de/mkravec/scripts
Prerequisite is https://gitlab.suse.de/mkravec/scripts/merge_requests/10

### DIFF
--- a/ci/jenkins/templates/validator-nightly-template.yaml
+++ b/ci/jenkins/templates/validator-nightly-template.yaml
@@ -1,0 +1,27 @@
+- job-template:
+    name: 'validator-join-nightly'
+    project-type: pipeline
+    number-to-keep: 30
+    days-to-keep: 30
+    branch: master
+    wrappers:
+      - timeout:
+          timeout: 120
+          fail: true
+    triggers:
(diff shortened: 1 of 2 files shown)


ci/jenkins/templates/validator-nightly-template.yaml
```diff
@@ -0,0 +1,27 @@
...
+    wrappers:
+      - timeout:
+          timeout: 120
+          fail: true
+    triggers:
+        - timed: 'H H(3-5) * * *'
+    parameters:
+        - string:
+            name: VALIDATOR_ARGS
+            default: '-t join -n 3:2 -c'
```
kravciak: Just to make it obvious, default might change
```suggestion
            default: '-p openstack -t join -n 3:2 -c'
```

MaximilianMeister: thanks updated

PR #781, opened 2019-10-16
Add `ControlPlaneHost` and `ControlPlaneHostAndPort` for templates

ereslibre:
## Why is this PR needed?

Templates can choose whether they want to use `ControlPlaneHost`, what
will remove the port of the control plane from the result, or
`ControlPlaneHostAndPort` that will respect the provided port for the
control plane (if any; otherwise, default to 6443).

This also ensures that we are explicit about what we are rendering,
since using `ControlPlane` only was very open and depended on whether
a user gave a port or not to the `--control-plane` argument before
defaulting, and behave wrongly after defaulting (adding port twice).

Fixes https://github.com/SUSE/avant-garde/issues/944

## Info for QA

Please just follow up the steps mentioned in SUSE/avant-garde#944 to verify gangway addon upgrade w/ valid gangway configmap, but this does not handle customized manifest changed by user.

### Related info

This is an alternative to https://github.com/SUSE/skuba/pull/777

After applying gangway addon upgrade, its config map should have invalid URLs.

```
redirectURL: "https://10.84.153.111:6443:32001/callback"
authorizeURL: "https://10.84.153.111:6443:32000/auth"
tokenURL: "https://10.84.153.111:6443:32000/token"
apiServerURL: "https://10.84.153.111:6443:6443"
```

After applying gangway addon upgrade, its config map should have valid URLs.

```
redirectURL: "https://10.84.153.111:32001/callback"
authorizeURL: "https://10.84.153.111:32000/auth"
tokenURL: "https://10.84.153.111:32000/token"
apiServerURL: "https://10.84.153.111:6443"
```

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/tests/conftest.py
+++ b/ci/infra/testrunner/tests/conftest.py
@@ -50,7 +50,7 @@ def deployment(request, bootstrap, skuba, kubectl):
     if request.config.getoption("skip_setup") != 'deployed':
         skuba.join_nodes()
 
-    wait(kubectl.run_kubectl, 'wait --timeout=1m --for=condition=Ready pods --all --namespace=kube-system', wait_delay=60, wait_timeout=300, wait_backoff=30, wait_retries=5, wait_allow=(RuntimeError))
+    wait(kubectl.run_kubectl, 'wait --timeout=5m --for=condition=Ready pods --all --namespace=kube-system', wait_delay=60, wait_timeout=300, wait_backoff=30, wait_retries=5, wait_allow=(RuntimeError))
 
 
 @pytest.fixture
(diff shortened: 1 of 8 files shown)


ereslibre:
This PR also covers a special case not handled in https://github.com/SUSE/skuba/pull/777: if a user passes a port on `skuba cluster init --control-plane <host>:<port>` for any reason. In that case, #777, still assumes in the templates that the apiserver location is <host>:6443, despite the user provided a specific `<port>`. This PR also aims to fix that as well, by making it explicit whether a template needs the host only, or the host and port (if no port was provided, it will default to 6443).

innobead:
Sorry for the late reply due to traveling, let’s work on this fix, and I will close the other ;)

ereslibre:
CI is green after replacing a 1m timeout with a 5m timeout for `kubectl wait` on deployments: https://github.com/SUSE/skuba/pull/781/commits/a2c8e18399ee412e8d0620afc9074bbe72ab4649

PR #767, opened 2019-10-10
Ssh errors should show errors regardless of v level (bsc#1151689)

suseclee:
If there is any error in sshing nodes, the error messages
   should be printed.
   The klog.Errorf() is show messages regardless of verbosity level.

## Why is this PR needed?

Fixes # https://bugzilla.suse.com/show_bug.cgi?id=1151689

## Anything else a reviewer needs to know?

Side effects for this commit is that there are errors already existed in normal successful processes. It is because we wanted to reset before bootstrap or join node. When I tried this, it took a while after Skuba printed out the error message. IMO, some people can misunderstand that Skuba is hanging from an error, which is not true. So after this commit, I need to tackle the noise.  
```
** This is an UNTAGGED version and NOT intended for production usage. **
[bootstrap] updating init configuration with target information
[bootstrap] writing init configuration for node
[bootstrap] applying init configuration to node
E1010 12:35:49.352456   24771 ssh.go:182] stderr | W1010 19:35:49.283153    5905 removeetcdmember.go:79] [reset] No kubeadm config, using etcd pod spec to get data directory
E1010 12:39:26.738068   24771 ssh.go:182] stderr | Created symlink /etc/systemd/system/timers.target.wants/skuba-update.timer → /usr/lib/systemd/system/skuba-update.timer.
[bootstrap] successfully bootstrapped core components on node "10.86.0.179" with Kubernetes: "1.15.2"
[bootstrap] downloading secrets from bootstrapped node "10.86.0.179"
[bootstrap] deploying core add-ons on node "10.86.0.179"
[bootstrap] successfully bootstrapped core add-ons on node "10.86.0.179"
```

## Info for QA
The problem was if we use default verbosity (level 0), vebosity level 0 hid error messages as well.
skuba node bootstrap --user sles --sudo --target 10.86.3.11 caasp-master-clee-0
skuba node join --role worker --user sles --sudo --target 10.86.0.108 caasp-worker-clee-0

Now when we run skuba commands such as bootstrap, join , remove, or upgrade with default verbosity, you need to see detailed error messages from ssh outputs when Skuba occurs errors.  

# Merge restrictions

(Please do not edit this)

We are in *v4-maintenance phase*, so we will restrict what can be merged to prevent unexpected surprises:

    What can be merged (merge criteria):
        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine
        there is a PR for updating documentation (or a statement that this is not needed)

PR #784, opened 2019-10-17
Add a progress indicator(spinner) in terminal

suseclee:
1. Add spinner while waiting ssh commands

## Why is this PR needed?

When users runs skuba commands, users may misunderstand if skuba does not print out anything for a while. So, skuba needs a progress indicator to terminal while Skuba commands are executed.

Fixes #

https://github.com/SUSE/avant-garde/issues/950

There are merge conflict between this PR and https://github.com/SUSE/skuba/pull/767
ssh.go needs to be changed to the following after #767 is merged.Or vice verse
```
result.Write([]byte(scanner.Text()))
		if description == "stdout" && !silent {
			spin.Stop()
			klog.V(1).Infof("%s", scanner.Text())
			spin.Start()
		} else if description == "stderr" {
			spin.Stop()
			klog.Errorf("%s", scanner.Text())
			spin.Start()
		}
``` 
# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


suseclee:
When I redirect to output file, it included scraps. 
Added `--no-progress`

dannysauer:
> When I redirect to output file, it included scraps.
> Added `--no-progress`

"is output a tty" should be detected and avoid this situation.

internal/pkg/skuba/deployments/ssh/ssh.go
```diff
@@ -169,10 +174,17 @@ func (t *Target) internalSshWithStdin(silent bool, stdin string, command string,
...
+		spin.Start()
+		defer spin.Stop()
+	}
 	stdoutChan := make(chan string)
 	stderrChan := make(chan string)
-	go readerStreamer(stdoutReader, stdoutChan, "stdout", silent)
-	go readerStreamer(stderrReader, stderrChan, "stderr", silent)
+
+	go readerStreamer(stdoutReader, stdoutChan, "stdout", silent, spin, t.no_progress)
+	go readerStreamer(stderrReader, stderrChan, "stderr", silent, spin, t.no_progress)
```
dannysauer: This is probably the right place to add the tty check.  If stdout or stderr are not connected to TTYs, just pass in false always, ignoring t.no_progress.  Maybe something like "isatty(sys.stderr) || t.no_progress", or however that would be properly written in Go. :D

suseclee: @dannysauer Thanks for the different viewpoint, I will look into it. If this works, then I do not need to have the `--no-progress` option, which is even greater.

go.sum
```diff
@@ -79,6 +81,8 @@ github.com/exponent-io/jsonpath v0.0.0-20151013193312-d6023ce2651d h1:105gxyaGwC
 github.com/exponent-io/jsonpath v0.0.0-20151013193312-d6023ce2651d/go.mod h1:ZZMPRZwes7CROmyNKgQzC3XPs6L/G2EJLHddWejkmf4=
 github.com/fatih/camelcase v0.0.0-20160318181535-f6a740d52f96 h1:5e8GDOdG6jKeeqNGbR+tlmqhf4vQVs3atTTMEWeEcAk=
 github.com/fatih/camelcase v0.0.0-20160318181535-f6a740d52f96/go.mod h1:yN2Sb0lFhZJUdVvtELVWefmrXpuZESvPmqwoZc+/fpc=
+github.com/fatih/color v1.7.0 h1:DkWD4oS2D8LGGgTQ6IvwJJXSL5Vp2ffcQg58nFV38Ys=
+github.com/fatih/color v1.7.0/go.mod h1:Zm6kSWBoL9eyXnKyktHP6abPY2pDugNf5KwzbycvMj4=
```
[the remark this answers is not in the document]
suseclee: Spinner has dependency on color.

internal/pkg/skuba/deployments/ssh/ssh.go
```diff
@@ -181,13 +192,19 @@ func (t *Target) internalSshWithStdin(silent bool, stdin string, command string,
...
 	for scanner.Scan() {
 		result.Write([]byte(scanner.Text()))
 		if !silent {
+			if isatty.IsTerminal(os.Stdout.Fd()) {
+				spin.Stop()
+			}
 			klog.V(1).Infof("%s | %s", description, scanner.Text())
+			if isatty.IsTerminal(os.Stdout.Fd()) {
+				spin.Start()
+			}
```
ereslibre: In terms of code changes this is the only thing I'm not sure about, calling to `Stop()` and `Start()` here when we are sharing a single `spin` and we have always two outputs that will be "fighting" at some extent for this resource.

suseclee: We need to stop spinner before write something in terminal.
Without stop and start, spinner marks will be remained in terminal. 
[==>   ] I1016 14:19:22.989369   19390 states.go:35] === applying state kubeadm.reset === 
So the outputs can be either terminal or file output. If a user redirects the outputs to file, spinner will not start.
I am not sure if I get your question correctly.

internal/pkg/skuba/deployments/ssh/ssh.go
```diff
@@ -181,13 +192,19 @@ func (t *Target) internalSshWithStdin(silent bool, stdin string, command string,
...
+func readerStreamer(reader io.Reader, outputChan chan<- string, description string, silent bool, spin *spinner.Spinner) {
 	result := bytes.Buffer{}
 	scanner := bufio.NewScanner(reader)
 	for scanner.Scan() {
 		result.Write([]byte(scanner.Text()))
 		if !silent {
+			if isatty.IsTerminal(os.Stdout.Fd()) {
+				spin.Stop()
+			}
 			klog.V(1).Infof("%s | %s", description, scanner.Text())
```
jenting: Why not change to `klog.Infof` or `fmt.Printf`? Then we don't need spinner

Klaven: `klog.Infof` should not be used. `klog.V(x).Infof` should be used as this logs all `info` under a verbosity setting.

ereslibre: Couldn't it be used for output where we want to respect the `klog` output settings, while printing with any verbosity level?

internal/pkg/skuba/deployments/ssh/ssh.go
innobead: It depends on start and stop implementation. From my understanding, it does take care.

internal/pkg/skuba/deployments/ssh/ssh.go
suseclee: This need to be verbosity level 1 because this includes details of ssh outputs. If we use Klog.Infof, we do not need verbosity level option in Skuba. We print all log output to customers. fmt.Print is not an option because Skuba output needs to be  consistent with timestamp. 
Spinner has simple task, giving an indicator to user while waiting. I saw, even with -v 1 that some of ssh cmds takes some times.

jenting: I am not challenge spinner, but keep check the isTerminal in a for loop a kind of repeated job. It could be check outside for once.

Further, I am not sure stdout and stderr share the same spin, any corner case on this? Do we have some race condition unit tests on this?

jenting: I am not challenge spinner, but keep check the isTerminal in a for loop a kind of repeated job. It could be check outside for once.

Further, I am not sure stdout and stderr share the same spin, any corner case on this? Do we have some race condition unit tests on this?

innobead: Taking back my approval because it seems more concern on the thread safe even it seems the lib already taking care, but do agree to have adding unit tests to strengthen the implementation to avoid impact user experience. 

but thanks @suseclee keep improving this PR :)

suseclee: I do not think we need to worry about sharing because stdout and stderr also share same tty. I can store the value of isTerminal and use it instead of calling the function every time.

dannysauer: I'm inclined to think that the spinner should only emit to stdout. Stderr should only be errors, and errors shouldn't need progress.

suseclee: The error is outcome of ssh cmd processes. We cannot expect which one will come out stderr or stdout.

suseclee:
spinner(pointer) is sharing with two threads by 
```
go readerStreamer(stdoutReader, stdoutChan, "stdout", silent, spin)
go readerStreamer(stderrReader, stderrChan, "stderr", silent, spin)
```
We do not know which one comes first. So whatever it come first, spin will be stopped, give output and  spin will be start again. Then it will be safe for next thread. At the end `defer` will stop spin when cmd ends.  I believe this is thread safe design. 

My concern is user point of view of spinner.  
In a bug https://bugzilla.suse.com/show_bug.cgi?id=1140526, stated the following
```
At Daimler we had to realize that a node bootstrap can take some long time for several task (i.e. depending on timeouts or internet speed).
While this shows as "nothing happens" - the admin gets nervous and is asking "is it broken? what is it doing etc.?
```

So, my suggestion was 

```
IMO,combining two (Case 1 and Case 2)  together can give high satisfaction to users of Skuba. Currently Skuba is not nice in the case of verbosity 0 (no outputs to users). 

Case 1: So we can give minimal outputs to users about what is going on in Skuba with verbosity 0.

https://github.com/SUSE/skuba/pull/767

skuba node bootstrap --user sles --sudo --target 10.86.0.191 caasp-master-clee-0
I1016 14:18:47.506221   19390 states.go:35] === applying state kubernetes.install-node-pattern ===
I1016 14:19:22.791149   19390 states.go:40] === state kubernetes.install-node-pattern applied successfully ===
I1016 14:19:22.791171   19390 bootstrap.go:101] updating init configuration with target information
I1016 14:19:22.989190   19390 bootstrap.go:125] writing init configuration for node
I1016 14:19:22.989355   19390 bootstrap.go:135] applying init configuration to node
I1016 14:19:22.989369   19390 states.go:35] === applying state kubeadm.reset ===
I1016 14:19:23.193533   19390 states.go:40] === state kubeadm.reset applied successfully ===
I1016 14:19:23.193547   19390 states.go:35] === applying state kubernetes.bootstrap.upload-secrets ===
I1016 14:19:23.193605   19390 states.go:40] === state kubernetes.bootstrap.upload-secrets applied successfully ===
I1016 14:19:23.193614   19390 states.go:35] === applying state kernel.load-modules ===
I1016 14:19:24.664246   19390 states.go:40] === state kernel.load-modules applied successfully ===
I1016 14:19:24.664257   19390 states.go:35] === applying state kernel.configure-parameters ===
I1016 14:19:25.809829   19390 states.go:40] === state kernel.configure-parameters applied successfully ===
I1016 14:19:25.809841   19390 states.go:35] === applying state apparmor.start ===
I1016 14:19:31.990669   19390 states.go:40] === state apparmor.start applied successfully ===
I1016 14:19:31.990680   19390 states.go:35] === applying state cri.configure ===
I1016 14:19:32.917628   19390 states.go:40] === state cri.configure applied successfully ===
I1016 14:19:32.917638   19390 states.go:35] === applying state cri.start ===
E1016 14:19:33.115137   19390 ssh.go:208] Created symlink /etc/systemd/system/multi-user.target.wants/crio.service → /usr/lib/systemd/system/crio.service.
I1016 14:19:34.291270   19390 states.go:40] === state cri.start applied successfully ===
I1016 14:19:34.291281   19390 states.go:35] === applying state kubelet.configure ===
I1016 14:19:35.931711   19390 states.go:40] === state kubelet.configure applied successfully ===
I1016 14:19:35.931726   19390 states.go:35] === applying state kubelet.enable ===
E1016 14:19:36.131051   19390 ssh.go:208] Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /usr/lib/systemd/system/kubelet.service.
I1016 14:19:36.444041   19390 states.go:40] === state kubelet.enable applied successfully ===
I1016 14:19:36.444061   19390 states.go:35] === applying state kubeadm.init ===
I1016 14:21:01.709979   19390 states.go:40] === state kubeadm.init applied successfully ===
I1016 14:21:01.710001   19390 states.go:35] === applying state skuba-update.start ===
E1016 14:21:01.922950   19390 ssh.go:208] Created symlink /etc/systemd/system/timers.target.wants/skuba-update.timer → /usr/lib/systemd/system/skuba-update.timer.
I1016 14:21:02.314847   19390 states.go:40] === state skuba-update.start applied successfully ===
I1016 14:21:02.314886   19390 bootstrap.go:155] successfully bootstrapped core components on node "10.86.0.191" with Kubernetes: "1.15.2"
I1016 14:21:02.314978   19390 bootstrap.go:162] downloading secrets from bootstrapped node "10.86.0.191"
I1016 14:21:03.991743   19390 bootstrap.go:72] deploying core add-ons on node "10.86.0.191"
I1016 14:21:14.395704   19390 bootstrap.go:86] successfully bootstrapped core add-ons on node "10.86.0.191"

Case 2: Another will be spinner because some of processes still takes some time (mainly depending on its network condition, I think). This can give more kindness to user such as many other products who people used in these days.
 
https://github.com/SUSE/skuba/pull/784
```

I am open to use (case 1) or (case 2) or (case 1 and case 2).

PR #785, opened 2019-10-18
Skip the psp image when listing all images for airgap

bergmannf:
The psp image is not an actual container, but it pretends to be one for use
cluster update logic.

## Why is this PR needed?

Currently the `psp` image is listed as an image - but the image does not actually exists.

This PR introduces a list of images that should be skipped when outputting images that are needed to bootstrap the cluster.

Fixes SUSE/avant-garde/issues/951

## Info for QA

The list of images output now should be `pull`able using any of the container tools.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


bergmannf:
Closing in favor of #786

PR #786, opened 2019-10-18
Skip the psp image when listing all images for airgap

bergmannf:
The psp image is not an actual container, but it pretends to be one for use
cluster update logic.

This is done using a sentinel object for the FakeVersion used by the psp image.
This can then be detected and skipped over.

This is an alternative implementation for #785 

## Why is this PR needed?

Currently the psp image is listed as an image - but the image does not actually exists.

This PR introduces a list of images that should be skipped when outputting images that are needed to bootstrap the cluster.

Fixes SUSE/avant-garde/issues/951

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubernetes/versions.go
+++ b/internal/pkg/skuba/kubernetes/versions.go
@@ -93,7 +93,7 @@ var (
 				Kured:   &AddonVersion{"1.2.0", 0},
 				Dex:     &AddonVersion{"2.16.0", 2},
 				Gangway: &AddonVersion{"3.1.0-rev4", 3},
-				PSP:     &AddonVersion{"1.0.0", 0},
+				PSP:     &AddonVersion{"", 0},
 			},
 		},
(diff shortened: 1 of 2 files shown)


bergmannf:
Using `nil` seems to not change what needs to change compared to the sentinel.

However I can not run an actual `upgrade` on my cluster, but I didn't find any references at all for the `PSP` key, so I guess nothing else needs to be adjusted.

bergmannf:
I think that proposal won't work, because you can't assign a `nil` value to a string in `go`:

`LSP: cannot convert nil (untyped nil value) to string`

I changed it the empty string now - but it is obviously far less explicit and won't easily break when being used somewhere, where `nil` would throw a `SegmentationFault`.

PR #791, opened 2019-10-23
Run more linters

aojea:
## Why is this PR needed?

Linters are very useful to keep code clean and avoid common problems

Fixes https://github.com/SUSE/avant-garde/issues/970

appear in the changelog.

## What does this PR do?

Installs a golang and shell linter so we can run them with `make lint`

I'll need some guidance here to understand what needs to be changed in the CI
Also, this will need to clean the errors in the code or remove some checks in the linters

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ lint: deps
 	# check terraform fmt
 	$(TERRAFORM) fmt -check=true -write=false -diff=true ci/infra
 	# run golangci-lint
-	$(BINPATH)/golangci-lint run
+	$(BINPATH)/golangci-lint -v run --timeout 3m
 	# run bash linter
 	find ci -type f -name "*.sh" | xargs $(BINPATH)/shellcheck
 
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ GO_MD2MAN ?= go-md2man
 LN = ln
 RM = rm
 
+BINPATH       := $(abspath ./bin)
 GOBINPATH     := $(shell $(GO) env GOPATH)/bin
 COMMIT        := $(shell git rev-parse HEAD)
 BUILD_DATE    := $(shell date +%Y%m%d)
@@ -66,6 +67,7 @@ install: go-version-check
 clean:
 	$(GO) clean -i ./...
 	$(RM) -f ./skuba
+	$(RM) -rf $(BINPATH)
 
 .PHONY: distclean
 distclean: clean
@@ -85,13 +87,24 @@ go-version-check:
 		[ $(GO_VERSION_MAJ) -eq 1 -a $(GO_VERSION_MIN) -ge 12 ] || (echo "FATAL: Go version should be >= 1.12.x" ; exit 1 ; )
 
 .PHONY: lint
-lint:
+lint: deps
 	# explicitly enable GO111MODULE otherwise go mod will fail
 	GO111MODULE=on go mod tidy && GO111MODULE=on go mod vendor && GO111MODULE=on go mod verify
+	# run go vet
 	$(GO) vet ./...
+	# run go gmt
 	test -z `$(GOFMT) -l $(SKUBA_SRCS)` || { $(GOFMT) -d $(SKUBA_SRCS) && false; }
+	# check terraform fmt
 	$(TERRAFORM) fmt -check=true -write=false -diff=true ci/infra
-	find ci -type f -name "*.sh" | xargs bashate
+	# run golangci-lint
+	$(BINPATH)/golangci-lint run
+	# run bash linter
+	find ci -type f -name "*.sh" | xargs $(BINPATH)/shellcheck
+
+.PHONY: deps
+deps:
+	test -f $(BINPATH)/golangci-lint || curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(BINPATH) v1.21.0
+	test -f $(BINPATH)/shellcheck || curl -sfL "https://storage.googleapis.com/shellcheck/shellcheck-v0.4.7.linux.x86_64.tar.xz" | tar -xJv --strip-components=1 -C $(BINPATH)
 
 .PHONY: suse-package
 suse-package:
--- a/pkg/skuba/cloud/cloud.go
+++ b/pkg/skuba/cloud/cloud.go
@@ -43,8 +43,7 @@ func ConfigHasRestrictedPermissions(file string) bool {
 	}
 	mode := info.Mode()
 	var mask os.FileMode = 077
-	if (mode & mask) > 0 {
-		return false
-	}
-	return true
+	// if (mode & mask) > 0 { return false }
+	// (mode & mask) is uint32
+	return (mode & mask) == 0
 }
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,82 @@
+linters-settings:
+  govet:
+    check-shadowing: true
+  golint:
+    min-confidence: 0
+  gocyclo:
+    min-complexity: 15
+  maligned:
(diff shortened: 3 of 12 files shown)


Makefile
```diff
@@ -18,6 +18,7 @@ GO_MD2MAN ?= go-md2man
 LN = ln
 RM = rm
 
+BINPATH:=$(abspath ./bin)
```
ereslibre: Nit: alignment

Makefile
```diff
@@ -85,13 +87,21 @@ go-version-check:
...
 .PHONY: lint
-lint:
+lint: deps
 	# explicitly enable GO111MODULE otherwise go mod will fail
 	GO111MODULE=on go mod tidy && GO111MODULE=on go mod vendor && GO111MODULE=on go mod verify
 	$(GO) vet ./...
 	test -z `$(GOFMT) -l $(SKUBA_SRCS)` || { $(GOFMT) -d $(SKUBA_SRCS) && false; }
 	$(TERRAFORM) fmt -check=true -write=false -diff=true ci/infra
-	find ci -type f -name "*.sh" | xargs bashate
+	$(BINPATH)/golangci-lint run cmd/... pkg/... internal/...
```
ereslibre:
```suggestion
	$(BINPATH)/golangci-lint run {cmd,pkg,internal}/...
```

Does this work?

Makefile
```diff
@@ -85,13 +87,21 @@ go-version-check:
...
+	find ci -type f -name "*.sh" | xargs $(BINPATH)/shellcheck
+
+.PHONY: deps
+deps:
+	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(BINPATH) v1.21.0
```
ereslibre: Nit: spaces around pipe

Makefile
```diff
@@ -85,13 +87,21 @@ go-version-check:
...
-lint:
+lint: deps
 	# explicitly enable GO111MODULE otherwise go mod will fail
 	GO111MODULE=on go mod tidy && GO111MODULE=on go mod vendor && GO111MODULE=on go mod verify
 	$(GO) vet ./...
 	test -z `$(GOFMT) -l $(SKUBA_SRCS)` || { $(GOFMT) -d $(SKUBA_SRCS) && false; }
 	$(TERRAFORM) fmt -check=true -write=false -diff=true ci/infra
-	find ci -type f -name "*.sh" | xargs bashate
+	$(BINPATH)/golangci-lint run cmd/... pkg/... internal/...
+	find ci -type f -name "*.sh" | xargs $(BINPATH)/shellcheck
```
ereslibre: Wouldn't this make sense to be added on the `ci` `Makefile`, and maybe calling it from here?

Makefile
saschagrunert:
```suggestion
BINPATH     := $(abspath ./bin)
```

Makefile
```diff
@@ -85,13 +87,21 @@ go-version-check:
...
+
+.PHONY: deps
+deps:
+	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(BINPATH) v1.21.0
+	$(BINPATH)/golangci-lint --version
+	wget -qO- "https://storage.googleapis.com/shellcheck/shellcheck-v0.4.7.linux.x86_64.tar.xz" | tar -xJv --strip-components=1 -C $(BINPATH)
+	$(BINPATH)/shellcheck --version
```
saschagrunert: I'd add a check if the binaries area already available to avoid re-downloads.

Makefile
saschagrunert: Do we want to pin the `master` to a dedicated version/commit too?

saschagrunert:
I think you could consider adding a `.golangci.yml` later on.

Makefile
aojea: this is the recommended procedure in the upstream golangci-lint repo, the binary is pinned but I think that they will be monitoring the install script.
Let's keep it this way and see what happens

jenting:
Here is `.golangci.yml` example I used before

```
run:
  deadline: 5m

linters-settings:
  dupl:
    threshold: 100
  errcheck:
    check-type-assertions: true
    check-blank: true
  govet:
    check-shadowing: true
  golint:
    min-confidence: 0
  gocyclo:
    min-complecity: 10
  gofmt:
    simplify: true
  maligned:
    suggest-new: true

linters:
  enable-all: true
  disable:
    - prealloc
    - dupl
    - lll
    - gochecknoinits
    - gochecknoglobals
```

saschagrunert:
We use this one in CRI-O:
https://github.com/cri-o/cri-o/blob/83b1c32bc09907b868505b471f830237c705ff64/.golangci.yml

I suggest in adding a comment to the yaml to see which linters are disabled on purpose. :innocent:

Makefile
aojea: this is replacing current configuration, I prefer not to use this PR to do major changes

Makefile
```diff
@@ -85,13 +87,24 @@ go-version-check:
...
 	GO111MODULE=on go mod tidy && GO111MODULE=on go mod vendor && GO111MODULE=on go mod verify
+	# run go vet
 	$(GO) vet ./...
+	# run go gmt
 	test -z `$(GOFMT) -l $(SKUBA_SRCS)` || { $(GOFMT) -d $(SKUBA_SRCS) && false; }
+	# check terraform fmt
 	$(TERRAFORM) fmt -check=true -write=false -diff=true ci/infra
-	find ci -type f -name "*.sh" | xargs bashate
+	# run golangci-lint
+	$(BINPATH)/golangci-lint run {cmd,internal,pkg}/...
```
saschagrunert: Doesn't it work the same if we just:
```suggestion
	$(BINPATH)/golangci-lint run
```

ereslibre: Unsure if that would include `vendor`, not sure if `golangci-lint` skips it by default...

saschagrunert: Yep, since the default config states:
```yaml
# default is true. Enables skipping of directories:
#   vendor$, third_party$, testdata$, examples$, Godeps$, builtin$
skip-dirs-use-default: true
```

aojea:
Folks PTAL

I've decided to keep the PR simple so we can review and merge without high risks.
First commit is adding the linters commands and subsequent commits per different linter fixes that are very simple.

We should follow up on `errcheck` errors , but let's get this merged and address that in a new PR

pkg/skuba/cloud/cloud.go
```diff
@@ -43,8 +43,7 @@ func ConfigHasRestrictedPermissions(file string) bool {
 	}
 	mode := info.Mode()
 	var mask os.FileMode = 077
-	if (mode & mask) > 0 {
-		return false
-	}
-	return true
+	// if (mode & mask) > 0 { return false }
+	// (mode & mask) is uint32
+	return (mode & mask) == 0
```
aojea: please verify this is correct

Makefile
```diff
@@ -97,7 +97,7 @@ lint: deps
 	# check terraform fmt
 	$(TERRAFORM) fmt -check=true -write=false -diff=true ci/infra
 	# run golangci-lint
-	$(BINPATH)/golangci-lint run
+	$(BINPATH)/golangci-lint -v run --timeout 3m
```
aojea: previous job timed out  in 1 min, this has to be something with the slave, it takes only a few seconds in my laptop and in this new run.

pkg/skuba/cloud/cloud.go
jenting: I think yes

PR #793, opened 2019-10-24
skuba node upgrade shows component downgrade (bsc#1154085)

jenting:
## Why is this PR needed?

The components `kubelet` and `cri-o` are handled by `skuba-update`, therefore, the patch version might mismatch compares to the kubernetes version. This would encounter a corner case that when the user runs `skuba node upgrade plan <worker-node>`, skuba shows component `kubelet` or `cri-o` downgrade.

Fixes https://github.com/SUSE/avant-garde/issues/959

appear in the changelog.

## What does this PR do?

- Function `HasMajorOrMinorUpdate()`:
  - The node is control plane, and if _updated_ apiserver major/minor version > _current_ apiserver major/minor version, return `true`; otherwise, return `false`.
  - The node is worker node, and if _updated_ kubelet major/minor version > _current_ kubelet major/minor version or if updated _cri-o_ major/minor version > current _cri-o_ major/minor version, return `true`; otherwise, return `false`.

- Function `IsUpdated()`:
  Check both `kubelet` and `cri-o` version is updated when:
  - Current major and minor version the same as updated major and minor version.
  - Current patch version large than or equals to updated patch version.

- Add unit tests to test the above functions.

N/A

## Info for QA

### Related info

N/A

1. Deploy a cluster (1 master node, 1 worker node) with version `1.15.0`
```
skuba cluster init --control-plane <IP/FQDN> --kubernetes-version 1.15.0 <cluster-name>
```
2. Run node upgrade plan on worker node
```
skuba node upgrade plan <worker-node-ip-fqdn>
```
3. Console output indicates `kubelet` downgrade
```
Current Kubernetes cluster version: 1.15.0

Current Node version: 1.15.2

Component versions in 10.86.0.32
  - kubelet: 1.15.2 -> 1.15.0
  - cri-o: 1.15.0 -> 1.15.0
```

1. Deploy a cluster (1 master node, 1 worker node) with version `1.15.0`
```
skuba cluster init --control-plane <IP/FQDN> --kubernetes-version 1.15.0 <cluster-name>
```
2. Run node upgrade plan on worker node
```
skuba node upgrade plan <worker-node-ip-fqdn>
```
3. Console output indicates worker node is up to date
```
Current Kubernetes cluster version: 1.15.0

Current Node version: 1.15.2

Node 10.86.0.32 is up to date
```

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/upgrade/node/versions.go
+++ b/internal/pkg/skuba/upgrade/node/versions.go
@@ -52,17 +52,22 @@ func (nviu NodeVersionInfoUpdate) HasMajorOrMinorUpdate() bool {
 		}
 	}
 	return nviu.Update.KubeletVersion.Major() > nviu.Current.KubeletVersion.Major() ||
-		nviu.Update.KubeletVersion.Minor() > nviu.Current.KubeletVersion.Minor()
+		nviu.Update.KubeletVersion.Minor() > nviu.Current.KubeletVersion.Minor() ||
+		nviu.Update.ContainerRuntimeVersion.Major() > nviu.Current.ContainerRuntimeVersion.Major() ||
+		nviu.Update.ContainerRuntimeVersion.Minor() > nviu.Current.ContainerRuntimeVersion.Minor()
 }
 
 func (nviu NodeVersionInfoUpdate) IsUpdated() bool {
-	return reflect.DeepEqual(nviu.Current.KubeletVersion, nviu.Update.KubeletVersion) &&
-		reflect.DeepEqual(nviu.Current.APIServerVersion, nviu.Update.APIServerVersion) &&
+	return reflect.DeepEqual(nviu.Current.APIServerVersion, nviu.Update.APIServerVersion) &&
 		reflect.DeepEqual(nviu.Current.ControllerManagerVersion, nviu.Update.ControllerManagerVersion) &&
 		reflect.DeepEqual(nviu.Current.SchedulerVersion, nviu.Update.SchedulerVersion) &&
 		reflect.DeepEqual(nviu.Current.EtcdVersion, nviu.Update.EtcdVersion) &&
+		nviu.Current.KubeletVersion.Major() == nviu.Update.KubeletVersion.Major() &&
+		nviu.Current.KubeletVersion.Minor() == nviu.Update.KubeletVersion.Minor() &&
+		nviu.Current.KubeletVersion.Patch() >= nviu.Update.KubeletVersion.Patch() &&
 		nviu.Current.ContainerRuntimeVersion.Major() == nviu.Update.ContainerRuntimeVersion.Major() &&
-		nviu.Current.ContainerRuntimeVersion.Minor() == nviu.Update.ContainerRuntimeVersion.Minor()
+		nviu.Current.ContainerRuntimeVersion.Minor() == nviu.Update.ContainerRuntimeVersion.Minor() &&
+		nviu.Current.ContainerRuntimeVersion.Patch() >= nviu.Update.ContainerRuntimeVersion.Patch()
 }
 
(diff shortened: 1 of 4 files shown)


internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -50,19 +50,24 @@ func (nviu NodeVersionInfoUpdate) HasMajorOrMinorUpdate() bool {
...
 	return nviu.Update.KubeletVersion.Major() > nviu.Current.KubeletVersion.Major() ||
-		nviu.Update.KubeletVersion.Minor() > nviu.Current.KubeletVersion.Minor()
+		nviu.Update.KubeletVersion.Minor() > nviu.Current.KubeletVersion.Minor() ||
+		nviu.Update.ContainerRuntimeVersion.Major() > nviu.Current.ContainerRuntimeVersion.Major() ||
+		nviu.Update.ContainerRuntimeVersion.Minor() > nviu.Current.ContainerRuntimeVersion.Minor()
 }
 
 func (nviu NodeVersionInfoUpdate) IsUpdated() bool {
-	return reflect.DeepEqual(nviu.Current.KubeletVersion, nviu.Update.KubeletVersion) &&
-		reflect.DeepEqual(nviu.Current.APIServerVersion, nviu.Update.APIServerVersion) &&
```
innobead: Why `reflect.DeepEqual(nviu.Current.APIServerVersion, nviu.Update.APIServerVersion)` got removed?

jenting: good catch, my bad

internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -50,19 +50,25 @@ func (nviu NodeVersionInfoUpdate) HasMajorOrMinorUpdate() bool {
 			nviu.Update.APIServerVersion.Minor() > nviu.Current.APIServerVersion.Minor() {
 			return true
 		}
+		return false
```
ereslibre: This should not be here, we want to fall through to checking kubelet and cri-o still.

On control-planes we first check the control plane components, then, if it doesn't have updates on those components, we check the kubelet and cri-o as we do on the rest of the nodes.

jenting: Thanks,@ereslibre , I got more clear on the checking logic

internal/pkg/skuba/upgrade/node/versions_test.go
```diff
@@ -356,21 +406,50 @@ func TestWorkerUpdateStatusWithAvailableVersions(t *testing.T) {
 			node:                   workerNode("worker1"),
 			expectedError:          true,
 		},
+		{
+			name:                   "worker with outdated control plane; patch version",
```
innobead:
```suggestion
			name:                   "worker with updated control plane; patch version",
```

internal/pkg/skuba/upgrade/node/versions_test.go
```diff
@@ -288,34 +296,72 @@ func TestControlPlaneUpdateStatusWithAvailableVersions(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
 			nodeVersionInfoUpdate, err := controlPlaneUpdateStatusWithAvailableVersions(tt.currentClusterVersion, tt.allNodesVersioningInfo, &tt.node, tt.versionInquirer)
-			if err == nil && tt.expectedError {
-				t.Errorf("error expected on %s, but no error reported", tt.name)
+			if tt.expectedError {
+				if err == nil {
+					t.Errorf("error expected on %s, but no error reported", tt.name)
+				}
 				return
 			} else if err != nil && !tt.expectedError {
```
innobead:
```suggestion
			} else if err != nil {
```

PR #794, opened 2019-10-24
Do not discard error values when retrieving pods

bergmannf:
Instead return the errors as any non-retrieved version will break during
SemanticParse anyhow

## Why is this PR needed?

Additional debugging information to see why some nightly CI runs break

PR #795, opened 2019-10-25
Fix node upgrade not success if patch version bump (bsc#1155173)

jenting:
## Why is this PR needed?

When the difference between _current_ and _update_ cluster version is only patch version, upgrade the first control plane node does not _really_ successfully.

This happens when we have a patched kubernetes version, as `1.15.0` and `1.15.2`. During `1.15.0` cluster deploying, the RPM package component `kubelet` will installed to the latest version `1.15.2`.
This causes the check logic `IsFirstControlPlaneNodeToBeUpgraded` always as `false` when only the patch version updated.

Fixes https://github.com/SUSE/avant-garde/issues/993

appear in the changelog.

## What does this PR do?

Fix the function `IsFirstControlPlaneNodeToBeUpgraded()`.

RFC about how to check first/second control plane to be upgraded.
- https://github.com/SUSE/caasp-rfc/blob/master/accepted/0040-platform-upgrade.md#first-control-plane-to-be-upgraded
- https://github.com/SUSE/caasp-rfc/blob/master/accepted/0040-platform-upgrade.md#secondary-control-plane-to-be-upgraded

## Info for QA

### Related info

Related PR is #793, but this PR address different behavior.

1. Deploy a version `1.15.0` cluster (1 master, 1 worker).
2. Execute node upgrade apply, it displays upgrade successfully.
```
skuba node upgrade apply --sudo --user sles --target <master-node-ip-fqdn> -v 5
```
3. Run upgrade plan on the master node, it displays still have a version to be upgraded from `1.15.0` to `1.15.2`.
```
skuba node upgrade plan <master-node-name>
```

1. Deploy a version `1.15.0` cluster (1 master, 1 worker).
2. Execute node upgrade apply, it displays upgrade successfully.
```
skuba node upgrade apply --sudo --user sles --target <master-node-ip-fqdn> -v 5
```
3. Run upgrade plan on the master node, it displays the node version is up to date.
```
skuba node upgrade plan <master-node-name>
```

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubernetes/node_version.go
+++ b/internal/pkg/skuba/kubernetes/node_version.go
@@ -25,7 +25,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/version"
-	"k8s.io/client-go/kubernetes"
+	clientset "k8s.io/client-go/kubernetes"
 )
 
 var (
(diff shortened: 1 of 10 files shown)


internal/pkg/skuba/kubernetes/node_version.go
```diff
@@ -185,10 +179,10 @@ func nodeVersioningInfoWithClientset(client kubernetes.Interface, nodeName strin
 
 	// find out the container image tags, depending on the role of the node
 	if IsControlPlane(nodeObject) {
```
mssola: Note that this will probably conflict with #794. We should merge that one first and then rebase this.

PR #801, opened 2019-10-28
Fix dockercaps test

pablochacin:
## Why is this PR needed?

Changes the way pod readiness is checked in Docker CAPS test to make it more reliable.

Fixes https://github.com/SUSE/avant-garde/issues/995

## What does this PR do?

Refactors the logic used for testing system pod readiness in the deployment fixture to reuse it as a generic function for testing readiness of pods in a namespace. Uses this function in the Docker CAPS tests.

Additionally, simplifies the test by using `run_kubectl` function's `stdin` parameter to pass a manifest without having to write it to a file first.
 
# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/tests/test_dockercaps.py
+++ b/ci/infra/testrunner/tests/test_dockercaps.py
@@ -4,7 +4,7 @@ import textwrap
 
 import pytest
 
-from tests.utils import wait
+from tests.utils import (check_pods_ready, wait)
 
 logger = logging.getLogger("testrunner")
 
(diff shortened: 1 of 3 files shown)


ci/infra/testrunner/tests/test_dockercaps.py
```diff
@@ -50,29 +50,19 @@
...
 @pytest.mark.flaky
-def test_dockercaps(deployment, kubectl, setup_manifest):
+def test_dockercaps(deployment, kubectl):
     logger.info("Deploy testcases")
     kubectl.run_kubectl(
-        "apply -f {}".format(setup_manifest))
+        "apply -f -", stdin=MANIFEST.encode())
 
-    wait(kubectl.run_kubectl,
-         "wait --for=condition=ready pods --all --timeout=0",
```
bergmannf: I think the problem is that this is using `--timeout=0`. which according to the docs:

`--timeout=30s: The length of time to wait before giving up.  Zero means check once and don't wait, negative means
wait for a week.`

So it will only every check once and not really do a wait - I guess just passing any timeouts to the `kubectl wait` would already help, as then it would actually wait for the resources.

pablochacin: Not sure I follow you. I'm removing the `wait` in favor of pulling the status and retrying using the wait function.

drpaneas:
Let me test this again with the new changes.

PR #803, opened 2019-10-29
Cloud integration: refactor openstack

ereslibre:
## Why is this PR needed?

Refactor openstack so it behaves like AWS integration, rendering all
needed settings upon `skuba cluster init` instead of overwriting
certain parameters when running `skuba node bootstrap`, what was
rather opaque.

This follows the approach of rendering as much as possible during the
`skuba cluster init` command, removing configuration overrides logic
from `skuba node bootsrap` and `skuba node join`.

Some logic, like adding target information to the configuration or the
token injection needs to happen at `skuba node join` time, however,
other settings can still be refactored to be rendered at `skuba
cluster init` time, like setting the admission plugins. This makes it
easier for a customer or user to inspect what configuration will be
applied to the cluster before they run `skuba node bootstrap`since
the number of overriden settings after `skuba cluster init` will be
smaller.

This contains many go template rendering logic. On a next refactor it
would be interesting to drop many go templating in favour of
initializing native kubeadm types from skuba, filling them and
marshaling them to disk, so there's no need to have complex logic on
go templates.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


pkg/skuba/actions/cluster/init/init.go
```diff
@@ -66,6 +78,59 @@ func (initConfiguration InitConfiguration) ControlPlaneHostAndPort() string {
...
+	case "aws":
+		if kubernetesComponent == ControllerManager {
+			extraArgs["allocate-node-cidrs"] = "false"
+		}
+	case "openstack":
+		switch kubernetesComponent {
+		case APIServer, ControllerManager, Kubelet:
+			extraArgs["cloud-config"] = skuba.OpenstackConfigRuntimeFile()
+		}
+	}
```
jenting: I would prefer cloud provider first, Kubernetes component later. [1]

This makes the logic more clear.

```
// APIServer, ControllerManager and Kubelet needs `cloud-provider` arg set if CPI is enabled
switch initConfiguration.CloudProvider {
case "aws":
	switch kubernetesComponent {
	case APIServer, Kubelet:
		extraArgs["cloud-provider"] = initConfiguration.CloudProvider
	case ControllerManager:
		extraArgs["cloud-provider"] = initConfiguration.CloudProvider
		extraArgs["allocate-node-cidrs"] = "false"
	}
case "openstack":
	switch kubernetesComponent {
	case APIServer, ControllerManager, Kubelet:
		extraArgs["cloud-provider"] = initConfiguration.CloudProvider
		extraArgs["cloud-config"] = skuba.OpenstackConfigRuntimeFile()
	}
}
```

pkg/skuba/actions/cluster/init/init.go
```diff
@@ -66,6 +78,59 @@ func (initConfiguration InitConfiguration) ControlPlaneHostAndPort() string {
...
+	return extraArgs
+}
+
+func (initConfiguration InitConfiguration) ComponentExtraVolumes(kubernetesComponent KubernetesComponent) []kubeadmapi.HostPathMount {
+	extraVolumes := []kubeadmapi.HostPathMount{}
+	switch kubernetesComponent {
+	case APIServer, ControllerManager:
+		if initConfiguration.CloudProvider == "openstack" {
```
jenting: Same as [1].

pkg/skuba/actions/cluster/init/init.go
ereslibre: I will update, I have no real preference, but that duplication of `cloud-provider` seems odd :)

ereslibre:
PR updated, ideally this changeset should require some QA cycles to verify that everything works as expected both with no cloud integration, AWS and OpenStack.

ereslibre:
Closing this because I'm exploring to directly use kubeadm types that we marshal. This will help in having generic mutators that also act upon upgrade, being able to modify cluster settings of existing clusters as if they were new deployments.

That logic has a lot in common with a different approach that will not leverage templating that much, but adapting native types and marshaling them.

Thanks to reviewers!

PR #808, opened 2019-11-06
Implement delay when joining nodes in testrunner

pablochacin:
## Why is this PR needed?

The join nodes command in the testrunner joins multiple nodes in sequence.

This introduces errors for example when a new master is added, because the cluster may not be stable until the node is ready and trying to join another node immediately fails.
Fixes 

## What does this PR do?

Implements a mechanism for waiting until the newly joined node appears as part of the cluster before attempting joining the next node in the sequence
rces or anything else that could be helpful to the reviewer.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/skuba/skuba.py
+++ b/ci/infra/testrunner/skuba/skuba.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import stat
+import time
 
 import platforms
 from utils.format import Format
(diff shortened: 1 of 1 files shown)


ci/infra/testrunner/skuba/skuba.py
```diff
@@ -104,12 +105,30 @@ def join_nodes(self, masters=None, workers=None):
...
 
-        for n in range(1, masters):
-            self.node_join("master", n)
-
-        for n in range(0, workers):
-            self.node_join("worker", n)
-
+        nodes = [("master", n) for n in range(1, masters)] + \
+                [("worker", n) for n in range(0, workers)] 
+        for r, n in nodes:
```
Itxaka: can we expand this to read `role, node_number` instead? Makes it clearer to know what we are doing :)

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -104,12 +105,30 @@ def join_nodes(self, masters=None, workers=None):
...
+            self.node_join(r, n)
+            self._wait_node_ready(r, n, timeout=180, backoff=20)
+
+
+    def _wait_node_ready(self, role, nr, timeout=60, backoff=10):
```
Itxaka: Same here with `rn` Im guessing its just the node number?

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -104,12 +105,30 @@ def join_nodes(self, masters=None, workers=None):
...
-            self.node_join("worker", n)
-
+        nodes = [("master", n) for n in range(1, masters)] + \
+                [("worker", n) for n in range(0, workers)] 
+        for role, node in nodes:
+            self.node_join(role, node)
+            self._wait_node_ready(role, node, timeout=180, backoff=20)
+
+
+    def _wait_node_ready(self, role, node, timeout=60, backoff=10):
```
MaximilianMeister: i'd call it `_wait_node_joined` as ready implies the `Ready` status which we don't check here

PR #810, opened 2019-11-07
Fix `skuba cluster images` returns wrong values (bsc#1155324)

jenting:
##  Why is this PR needed?

Partial container images cannot be pulled and do not match as the cluster images.
```
# cluster images list
registry.suse.com/caasp/v4/dex:2.16.0
# crictl images
registry.suse.com/caasp/v4/caasp-dex 2.16.0

# images not visible in cluster images output
registry.suse.com/caasp/v4/cilium-init 1.5.3
registry.suse.com/caasp/v4/cilium-operator 1.5.3

# image does not exist in registry.suse.de/com
1.15.2 registry.suse.com/caasp/v4/tooling:0.1.0
```

Fixes #https://github.com/SUSE/avant-garde/issues/1013

appear in the changelog.

## What does this PR do?

1. Change the `suba cluster images` help workding.
2. Fix incompatible container image name.

N/A

## Info for QA

### Related info

N/A

Run the following commands, partial container images cannot be pulled.
```
skuba cluster images | grep '1.15.2' | awk '{print $2}' | xargs -I{} docker pull {}
```

Run the following commands, all container images can be pulled.
```
skuba cluster images | grep '1.15.2' | awk '{print $2}' | xargs -I{} docker pull {}
```

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/cmd/skuba/cluster/images.go
+++ b/cmd/skuba/cluster/images.go
@@ -18,6 +18,8 @@
 package cluster
 
 import (
+	"os"
+
 	"github.com/spf13/cobra"
 	"k8s.io/klog"
 
@@ -28,10 +30,11 @@ import (
 func NewImagesCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:   "images",
-		Short: "Show images to be pulled",
+		Short: "Show a list of images being used in the cluster",
 		Run: func(cmd *cobra.Command, args []string) {
 			if err := cluster.Images(); err != nil {
 				klog.Errorf("unable to get cluster images: %s", err)
+				os.Exit(1)
 			}
 		},
 		Args: cobra.NoArgs,
--- a/internal/pkg/skuba/addons/addons.go
+++ b/internal/pkg/skuba/addons/addons.go
@@ -47,10 +47,11 @@ const (
 var Addons = map[kubernetes.Addon]Addon{}
 
 type Addon struct {
-	addon         kubernetes.Addon
-	templater     addonTemplater
-	callbacks     addonCallbacks
-	addonPriority addonPriority
+	addon             kubernetes.Addon
+	templater         addonTemplater
+	callbacks         addonCallbacks
+	addonPriority     addonPriority
+	getImageCallbacks []getImageCallback
 }
 
 type addonCallbacks interface {
@@ -65,6 +66,8 @@ type AddonConfiguration struct {
 }
 
 type addonTemplater func(AddonConfiguration) string
+
+type getImageCallback func(imageTag string) string
 
 type ApplyBehavior uint
 
@@ -99,12 +102,13 @@ func (renderContext renderContext) ManifestVersion() string {
 	return fmt.Sprintf("%s-%d", addonVersion.Version, addonVersion.ManifestVersion)
 }
 
-func registerAddon(addon kubernetes.Addon, addonTemplater addonTemplater, callbacks addonCallbacks, addonPriority addonPriority) {
+func registerAddon(addon kubernetes.Addon, addonTemplater addonTemplater, callbacks addonCallbacks, addonPriority addonPriority, getImageCallbacks []getImageCallback) {
 	Addons[addon] = Addon{
-		addon:         addon,
-		templater:     addonTemplater,
-		callbacks:     callbacks,
-		addonPriority: addonPriority,
+		addon:             addon,
+		templater:         addonTemplater,
+		callbacks:         callbacks,
+		addonPriority:     addonPriority,
+		getImageCallbacks: getImageCallbacks,
 	}
 }
 
@@ -261,6 +265,14 @@ func (addon Addon) Apply(addonConfiguration AddonConfiguration, skubaConfigurati
 		}
 	}
 	return updateSkubaConfigMapWithAddonVersion(addon.addon, addonConfiguration.ClusterVersion, skubaConfiguration)
+}
+
+func (addon Addon) Images(imageTag string) []string {
+	images := []string{}
+	for _, cb := range addon.getImageCallbacks {
+		images = append(images, cb(imageTag))
+	}
+	return images
 }
 
 func addonVersionLower(current *kubernetes.AddonVersion, updated *kubernetes.AddonVersion) bool {
--- a/internal/pkg/skuba/kubernetes/versions.go
+++ b/internal/pkg/skuba/kubernetes/versions.go
@@ -19,10 +19,13 @@ package kubernetes
 
 import (
 	"fmt"
-	"log"
 	"sort"
 
 	"k8s.io/apimachinery/pkg/util/version"
+	"k8s.io/klog"
+	"k8s.io/kubernetes/cmd/kubeadm/app/images"
+
+	"github.com/SUSE/skuba/pkg/skuba"
 )
 
 type Addon string
@@ -35,8 +38,8 @@ const (
 	Gangway Addon = "gangway"
 	PSP     Addon = "psp"
 
-	ContainerRuntime Component = "cri-o"
 	Kubelet          Component = "kubelet"
+	ContainerRuntime Component = "cri-o"
 
 	Hyperkube Component = "hyperkube"
 	Etcd      Component = "etcd"
@@ -46,18 +49,17 @@ const (
 	Tooling Component = "tooling"
 )
 
-type ControlPlaneComponentsVersion struct {
-	HyperkubeVersion string
-	EtcdVersion      string
+type ComponentHostVersion struct {
+	KubeletVersion          string
+	ContainerRuntimeVersion string
 }
 
-type ComponentsVersion struct {
-	ContainerRuntimeVersion string
-	KubeletVersion          string
-	CoreDNSVersion          string
-	PauseVersion            string
-	ToolingVersion          string
+type ContainerImageTag struct {
+	Name string
+	Tag  string
 }
+
+type ComponentContainerVersion map[Component]*ContainerImageTag
 
 type AddonVersion struct {
 	Version         string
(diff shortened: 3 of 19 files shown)


internal/pkg/skuba/addons/addons.go
```diff
@@ -66,6 +67,8 @@ type AddonConfiguration struct {
 
 type addonTemplater func(AddonConfiguration) string
 
+type getImageCb func(imageTag string) string
```
mssola: Nitpick: I would rename it to `getImageCallback`, since we are using the full `callback` word instead of `cb` in other places.

c3y1huang: +1

kravciak:
Should I file separate bug about `1.14.1` and `1.15.0` images in the output?
Those are used only in devel skuba.

cmd/skuba/cluster/images.go
```diff
@@ -28,10 +30,11 @@ import (
 func NewImagesCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:   "images",
-		Short: "Show images to be pulled",
+		Short: "Show a list of images will use",
```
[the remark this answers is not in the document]
jenting: I choose the first one :+1:

internal/pkg/skuba/addons/addons.go
jenting: Renamed :)

internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -174,7 +207,11 @@ func AddonVersionWithAvailableVersions(addon Addon, clusterVersion *version.Vers
 }
 
 func AddonVersionForClusterVersion(addon Addon, clusterVersion *version.Version) *AddonVersion {
-	return AddonVersionWithAvailableVersions(addon, clusterVersion, Versions)
+	return AddonVersionWithAvailableVersions(addon, clusterVersion, supportVersions)
+}
+
+func AddonsVersionForClusterVersion(clusterVersion *version.Version) AddonsVersion {
```
bergmannf: I would maybe try to name this function differently `AllAddonVersionsForClusterVersion`, just to distinguish it more from the function defined right above it `AddonVersionForClusterVersion`.

internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -142,27 +153,49 @@ var (
...
 }
 
 func ComponentVersionForClusterVersion(component Component, clusterVersion *version.Version) string {
-	return ComponentVersionWithAvailableVersions(component, clusterVersion, Versions)
+	return ComponentVersionWithAvailableVersions(component, clusterVersion, supportVersions)
+}
+
+func ComponentContainerImageWithAvailableVersions(component Component, clusterVersion *version.Version, availableVersions KubernetesVersions) string {
+	imageTag := ComponentVersionWithAvailableVersions(component, clusterVersion, availableVersions)
+	switch component {
```
mssola: I have to admit that I don't fully understand why we need the `Component` type at all (it's just a `string` alias with no methods associated with it). That being said, if this is just a string, couldn't this whole switch statement be modified to be simply something like:

```go
return images.GetGenericImage(skuba.ImageRepository, component, imageTag)
```

To be honest, I'm not sure when is that "unknown component" case ever reached. Do we even expect that to happen? (I think we don't, that's why we have the `panic` there, but still, it feels weird). To counter this I'd log the image repository being picked before returning. Otherwise, I'm sure there has to be a more clever way to handle this (on the top of my head: instead of having separate variables, have a slice of possible components or a map if it's easier, then check that the given component exists on that map). Otherwise, it feels like this switch would grow indefinitely in the future (i.e. new components and such), but maybe I'm wrong.

jenting: > I have to admit that I don't fully understand why we need the `Component` type at all (it's just a `string` alias with no methods associated with it). That being said, if this is just a string, couldn't this whole switch statement be modified to be simply something like:
> 
> ```go
> return images.GetGenericImage(skuba.ImageRepository, component, imageTag)
> ```
Because we have a container image called `skuba-tooling`, the variable cannot be `skuba-tooling` :)

> To be honest, I'm not sure when is that "unknown component" case ever reached. Do we even expect that to happen? (I think we don't, that's why we have the `panic` there, but still, it feels weird). To counter this I'd log the image repository being picked before returning. Otherwise, I'm sure there has to be a more clever way to handle this (on the top of my head: instead of having separate variables, have a slice of possible components or a map if it's easier, then check that the given component exists on that map). Otherwise, it feels like this switch would grow indefinitely in the future (i.e. new components and such), but maybe I'm wrong.

You are right, this is my general thought before, maybe it's better than current.
In this way, we could cover one case that the control plane container image name might be changed in the future. Also, no need to manually add a new component to the switch case. 
 WDYT?
```
type ContainerImageTag struct {
	Name string
	Tag  string
}
type ControlPlaneHostVersion struct {
	KubeletVersion          string
	ContainerRuntimeVersion string
}
type ControlPlaneContainerVersion map[Component]*ContainerImageTag
type AddonsVersion map[Addon]*AddonVersion

type KubernetesVersion struct {
	ControlPlaneHostVersion      ControlPlaneHostVersion
	ControlPlaneContainerVersion ControlPlaneContainerVersion
	AddonsVersion                AddonsVersion
}

type KubernetesVersions map[string]KubernetesVersion

var supportVersions = KubernetesVersions{
		"1.15.2": KubernetesVersion{
			ControlPlaneHostVersion: ControlPlaneHostVersion{
				KubeletVersion:          "1.15.2",
				ContainerRuntimeVersion: "1.15.0",
			},
			ControlPlaneContainerVersion: ControlPlaneContainerVersion{
				Hyperkube: &ContainerImageTag{"hyperkube", "v1.15.2"},
				Etcd:      &ContainerImageTag{"etcd", "3.3.11"},
				CoreDNS:   &ContainerImageTag{"coredns", "1.3.1"},
				Pause:     &ContainerImageTag{"pause", "3.1"},
				Tooling:   &ContainerImageTag{"skuba-tooking", "0.1.0"},
			},
			AddonsVersion: AddonsVersion{
				Cilium:  &AddonVersion{"1.5.3", 1},
				Kured:   &AddonVersion{"1.2.0", 0},
				Dex:     &AddonVersion{"2.16.0", 3},
				Gangway: &AddonVersion{"3.1.0-rev4", 3},
				PSP:     &AddonVersion{"", 0},
			},
		},
	}

func ComponentContainerImageWithAvailableVersions(component Component, clusterVersion *version.Version, availableVersions KubernetesVersions) string {
	currentKubernetesVersion := availableVersions[clusterVersion.String()]
	if componentVersion, found := currentKubernetesVersion.ControlPlaneContainerVersion[component]; found {
		return images.GetGenericImage(skuba.ImageRepository, componentVersion.Name, componentVersion.Tag)
	}
	return ""
}
```

mssola: > Because we have a container image called skuba-tooling, the variable cannot be skuba-tooling :)

Gotcha :sweat_smile: 

> You are right, this is my general thought before, maybe it's better than current.
In this way, we could cover one case that the control plane container image name might be changed in the future. Also, no need to manually add a new component to the switch case.
WDYT?

LGTM

chentex: OK, I think I finally managed to understand the change here, I like the new map approach :+1:

bergmannf: Couldn't we use the same logic for all of the `versions` if we just make all of the `ContainerImageTag` types?

internal/pkg/skuba/kubernetes/versions.go
jenting: Changed!

internal/pkg/skuba/kubernetes/versions.go
jenting: Well, cilium addon is much complicated to proceed because it has _cilium-init_, _cilium-operator_, and _cilium_. I am still thinking a way to simplify the code and make it readable and easier to maintain/developing.

jenting:
> Should I file separate bug about `1.14.1` and `1.15.0` images in the output?
> Those are used only in devel skuba.

Finally, I realize the risk that `1.14.1` and `1.15.0` are not release to `registry.suse.de/suse/sle-15-sp1/update/products/casp40/containers/caasp/v4` and `registry.suse.com/caasp/v4`.

I fix the issue in the current PR.

internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -46,19 +49,18 @@ const (
 	Tooling Component = "tooling"
 )
 
-type ControlPlaneComponentsVersion struct {
-	HyperkubeVersion string
-	EtcdVersion      string
+type ControlPlaneHostVersion struct {
```
ereslibre: I'm not sure about this naming, this is not only about control planes, but also about worker nodes, right?

jenting: @ereslibre Thanks for pointing out. Change to `ComponentHostVersion` and `ComponentContainerVersion`. Hope this naming is better than before :)

internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -67,26 +69,26 @@ type AddonVersion struct {
...
+	ComponentHostVersion      ComponentHostVersion
+	ComponentContainerVersion ComponentContainerVersion
+	AddonsVersion             AddonsVersion
 }
 
 type KubernetesVersions map[string]KubernetesVersion
 
 var (
-	Versions = KubernetesVersions{
+	supportVersions = KubernetesVersions{
```
ereslibre: I don't see a strong need for renaming this (aside from not making it public), it also makes the whole diff bigger. However, I think it's fine, but I'd use `supportedVersions` instead, WDYT?

jenting: Sure, I made it private since the originally same package uses this variable directly, some use API to access it. Will change to `supportedVersions`.

PR #814, opened 2019-11-08
Fix long hostnames when running in ci

pablochacin:
## Why is this PR needed?

In ci, the hostnames assigned to the nodes include the
stackname, which is taken from the ci jobname. In some
cases this name exceeds the maximum of 63 characters.

Fixes https://github.com/SUSE/avant-garde/issues/1021

## What does this PR do?

This fix ensures the name is valid by truncating the
stack name passed to testrunner.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/platforms/terraform.py
+++ b/ci/infra/testrunner/platforms/terraform.py
@@ -82,8 +82,8 @@ class Terraform(Platform):
         return len(self.get_nodes_ipaddrs(role))
 
     def get_nodes_names(self, role):
-        stack_name = self.conf.terraform.stack_name
-        return [f'caasp-{role}-{stack_name.lower()}-{i}' for i in range(self.get_num_nodes(role))] 
+        stack_name = self.stack_name()
+        return [f'caasp-{role}-{stack_name}-{i}' for i in range(self.get_num_nodes(role))] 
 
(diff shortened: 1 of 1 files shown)


ci/infra/testrunner/platforms/terraform.py
```diff
@@ -115,10 +115,18 @@ def _generate_tfvars_file(self):
             with open(tfvars_final, "w") as f:
                 json.dump(tfvars, f)
 
+    #take up to 50 characters from stackname to give room to the fixed part
+    # in the node name: caasp-[master|worker]-xx (total length must be <= 63)
+    def stack_name(self):
+         stack_name = self.conf.terraform.stack_name
+         while len(stack_name) > 45:
```
Klaven: question. could we not remove this while loop in favor of just getting the sub string at length 45. And then if it is important to break it at a `-` just run the `rfind('-')` once after?

My thought is this. what if there is 59 characters before the first `-`? this would loop forever right? hopfully that cant happen, right?

pablochacin: Good point. The intention of this loop is to remove whole components of the name. Truncating may lead to weird names.

PR #816, opened 2019-11-11
Fix invalid characters in node names

pablochacin:
## Why is this PR needed?

The stackname is used as part of the nodename. The node name has several restrictions including valid characters (alphanumeric, "." and "-") and start and end with alphanumeric. 

## What does this PR do?

Ensures "_" is replaced by "-" and that the stackname starts and ends with
alphanumeric characters (to avoid spurious "-" due to truncation).

Follow up to: https://github.com/SUSE/skuba/pull/814

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/platforms/terraform.py
+++ b/ci/infra/testrunner/platforms/terraform.py
@@ -115,11 +115,18 @@ class Terraform(Platform):
             with open(tfvars_final, "w") as f:
                 json.dump(tfvars, f)
 
-    #take up to 50 characters from stackname to give room to the fixed part
-    # in the node name: caasp-[master|worker]-xxx (total length must be <= 63)
+    # take up to 45 characters from stackname to give room to the fixed part
+    # in the node name: caasp-[master|worker]-<stack name>-xxx (total length
+    # must be <= 63).
+    # Also ensure that only valid character are present and that the string
+    # starts and ends with alphanumeric characters and all lowercase.
     def stack_name(self):
-         stack_name = self.conf.terraform.stack_name
-         return stack_name[:45].lower()
+         stack_name = self.conf.terraform.stack_name[:45]
+         stack_name = stack_name.replace("_","-").replace("/","-")
+         stack_name = stack_name.strip("-.")
+         stack_name = stack_name.lower()
+   
+         return stack_name
 
     def _update_tfvars(self, tfvars):
         new_vars = {


ci/infra/testrunner/platforms/terraform.py
```diff
@@ -115,11 +115,18 @@ def _generate_tfvars_file(self):
...
+    # in the node name: caasp-[master|worker]-<stack name>-xxx (total length
+    # must be <= 63).
+    # Also ensure that only valid character are present and that the string
+    # starts and ends with alphanumeric characters and all lowercase.
     def stack_name(self):
-         stack_name = self.conf.terraform.stack_name
-         return stack_name[:45].lower()
+         stack_name = self.conf.terraform.stack_name[:45]
+         stack_name = stack_name.replace("_","-").replace("/","-")
+         stack_name = stack_name.strip("-.").lstrip("-.")
```
Klaven: I believe `lstrip` is redundant here unless I misunderstand. `strip` should remove all cases of it. This should not cause a problem. just extra code unless i'm missing something.

ci/infra/testrunner/platforms/terraform.py
```diff
@@ -115,11 +115,18 @@ def _generate_tfvars_file(self):
...
+    # in the node name: caasp-[master|worker]-<stack name>-xxx (total length
+    # must be <= 63).
+    # Also ensure that only valid character are present and that the string
+    # starts and ends with alphanumeric characters and all lowercase.
     def stack_name(self):
-         stack_name = self.conf.terraform.stack_name
-         return stack_name[:45].lower()
+         stack_name = self.conf.terraform.stack_name[:45]
+         stack_name = stack_name.replace("_","-").replace("/","-")
+         stack_name = stack_name.strip("-.")
```
dannysauer: The comment says this ensures that only valid characters are present and that it starts/ends with alnum.  But the code says "remove underscores and slashes, and make sure it doesn't start or end with a period or hyphen."

Why doesn't the code do what the comment says?  Is there a guarantee elsewhere that underscore, slash, and period are the only non-alnum chars which could possibly be in this value?

dannysauer: I'd personally be _strongly_ inclined to use something like `stack_name = re.sub(r'[^.A-Za-z\d]', '-', stack_name)` to replace all chars that aren't period or alnum with a hyphen.

pablochacin: Hi Dany. I agree that this is the safest solution. I was working on the assumption that the problem was detected in the context of the ci, where the only possible invalid character is "_". But your suggestion would be a good follow up to this. Thanks

PR #820, opened 2019-11-13
Use datastore cluster instead of datastore

jordimassaguerpla:
## Why is this PR needed?

We're facing performance issues with the datastore
backed by iSCSI storage (3PAR). Using the new
datastore cluster (LOCAL-DISKS-CLUSTER) backed
by local disks should improve the performances.

fixes issue#1034

## What does this PR do?

Fix vmware ci flakiness regarding accessing storage

## Info for QA

This PR is for CI, nothing for QA.

## Docs

This PR is for CI, nothing for Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/vmware/README.md
+++ b/ci/infra/vmware/README.md
@@ -53,6 +53,7 @@ Copy the `terraform.tfvars.example` to `terraform.tfvars` and provide reasonable
 ## Variables
 
 `vsphere_datastore` - Provide the datastore to use in vSphere\
+`vsphere_datastore_cluster` - Provide the datastore cluster to use on the vSphere server\
 `vsphere_datacenter` - Provide the datacenter to use in vSphere\
 `vsphere_network` - Provide the network to use in vSphere - this network must be able to access the ntp servers and the nodes must be able to reach each other\
 `vsphere_resource_pool` - Provide the resource pool the machines will be running in\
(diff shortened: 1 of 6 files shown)


jordimassaguerpla:
> We can't use it like that, see the node from the first PR #385
> 
> We would need to somehow patch the terraform files during a CI run because we can't just push it like this to customers.

I don't know much about terraform, but could we make the following:

datastore_cluster_id = ${data.vsphere_datastore_cluster.datastore_cluster} == {} ? "${data.vsphere_datastore.datastore.id}": "${data.vsphere_datastore_cluster.datastore_cluster.id}"

in 

https://github.com/SUSE/skuba/pull/820/files#diff-8c492ece61d9a77075f0c3f341c3784e
https://github.com/SUSE/skuba/pull/820/files#diff-e18dbe9b97d84e82471263e08e7656f7
https://github.com/SUSE/skuba/pull/820/files#diff-6d4320b64cc875b1da335f1c70ae28e9

Makes sense?

lcavajani:
@jordimassaguerpla unfortunately we can't do that because `datastore_id` and  `datastore_cluster_id` are two things completely differents :/

You can't use a datastore ID as a datastore_cluster ID for example

jordimassaguerpla:
> @jordimassaguerpla unfortunately we can't do that because `datastore_id` and `datastore_cluster_id` are two things completely differents :/
> 
> You can't use a datastore ID as a datastore_cluster ID for example

How could we fix it instead?

jordimassaguerpla:
Closing PR. It was a nice try :) . We will have to think of something different.

PR #822, opened 2019-11-13
CI: Fix upgrade apply user lock

ereslibre:
## Why is this PR needed?

Variable shadowing was not working because of a typo, so we were
trying to annotate twice instead of the intended action of annotating
and getting the result.

Fixes https://github.com/SUSE/avant-garde/issues/1040


jordimassaguerpla:
Thanks @ereslibre for your quick response on fixing this test.

PR #825, opened 2019-11-14
Add monitoring test to validator

MaximilianMeister:
## What does this PR do?

Add a test for prometheus and grafana deployment

* SUSE helm charts for prometheus and grafana are not yet merged https://github.com/SUSE/kubernetes-charts-suse-com/pull/103
* container images are not yet available in registry.suse.com


MaximilianMeister:
I've added a workaround now into the test in validator for the meantime, to use registry.suse.de images until they become available. So this CI job is not blocked.

PR #828, opened 2019-11-15
Do not create kubelet sysconfig file

flavio:
## Why is this PR needed?

skuba always overwrite `/etc/sysconfig/kubelet` with an empty file when a node is joined/bootstrapped.

This can lead user's customization to be lost.

For example: user can specify node labels at infrastructure creation time by providing a properly crafted `/etc/sysconfig/kubelet`. The file should not be overwritten by skuba when the node is added to a cluster.

## What does this PR do?

Our empty `/etc/sysconfig/kubelet` file is **never** created ~only when the file doesn't exist on the target node~.

Nothing special.

## Info for QA

How to verify the feature.

Ensure new regression:
  * Create a new node
  * Do not create the `/etc/sysconfig/kubelet` file
  * Use the node via skuba during a bootstrap/join operation
  * At the end of the process the `/etc/sysconfig/kubelet` file exists won't exist ~and contains a boilerplate created by skuba~

Ensure new behaviour works as expected:
  * Create a new node
  * Create a `/etc/sysconfig/kubelet` file with some custom comments
  * Use the node via skuba during a bootstrap/join operation
  * At the end of the process the `/etc/sysconfig/kubelet` file exists and is still the same one created by the user

## Docs

No need to change docs.

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/kubelet.go
+++ b/internal/pkg/skuba/deployments/ssh/kubelet.go
@@ -141,9 +141,6 @@ func kubeletConfigure(t *Target, data interface{}) error {
 		if err := t.UploadFileContents("/usr/lib/systemd/system/kubelet.service.d/10-kubeadm.conf", assets.KubeadmService); err != nil {
 			return err
 		}
-		if err := t.UploadFileContents("/etc/sysconfig/kubelet", assets.KubeletSysconfig); err != nil {
-			return err
-		}
 	} else {
 		if err := t.UploadFileContents("/lib/systemd/system/kubelet.service", assets.KubeletService); err != nil {
 			return err
--- a/internal/pkg/skuba/deployments/ssh/assets/assets.go
+++ b/internal/pkg/skuba/deployments/ssh/assets/assets.go
@@ -42,6 +42,4 @@ EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env
 EnvironmentFile=-/etc/sysconfig/kubelet
 ExecStart=
 ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS`
-
-	KubeletSysconfig = "KUBELET_EXTRA_ARGS="
 )


internal/pkg/skuba/deployments/ssh/kubelet.go
```diff
@@ -41,8 +41,13 @@ func kubeletConfigure(t *Target, data interface{}) error {
...
 		}
-		if err := t.UploadFileContents("/etc/sysconfig/kubelet", assets.KubeletSysconfig); err != nil {
-			return err
+
+		const kubelet_sysconfig = "/etc/sysconfig/kubelet"
+		if contents, _ := t.DownloadFileContents(kubelet_sysconfig); contents == "" {
+			// Upload file only when it doesn't alreayd exist.
+			if err := t.UploadFileContents(kubelet_sysconfig, assets.KubeletSysconfig); err != nil {
+				return err
+			}
```
kkaempf: This is changing `skuba` semantics and skuba should be more verbose of this change. Like adding an `else` case saying that `/etc/sysconfig/kubelet` already exists and will not be touched.

flavio:
> In my opinion I think leaving `/etc/sysconfig/kubelet` entirely for the user to use would be the best approach, so skuba should not own it.

Yes, I totally agree about that. I just wonder if *not* having `/etc/sysconfig/kubelet` won't cause any issue to the systemd drop-in we upload:
https://github.com/SUSE/skuba/blob/5108fab9e25421661548f3708953c1988da17015/internal/pkg/skuba/deployments/ssh/assets/assets.go#L42

Itxaka:
we should use `/etc/systemd/system/kubelet.service.d/` for any overriding parameters that we want. Otherwise that file should be treated as config (`%config(noreplace)`), that is how its done in upstream: https://github.com/kubernetes/release/pull/658/files

Itxaka:
> Yes, I totally agree about that. I just wonder if not having /etc/sysconfig/kubelet won't cause any issue to the systemd drop-in we upload:

@flavio no issues there, systemd accounts for that (the minus in front of the file)[0]:

```
The argument passed should be an absolute filename or wildcard expression, optionally prefixed with "-", which indicates that if the file does not exist, it will not be read and no error or warning message is logged.
```

[0] https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=

ereslibre:
> we should use /etc/systemd/system/kubelet.service.d/ for any overriding parameters that we want.

We are using `/usr/lib/systemd/system/kubelet.service.d` though, `kubeadm` writes the `10-kubeadm.conf` drop in under this location.

Itxaka:
> We are using /usr/lib/systemd/system/kubelet.service.d though, kubeadm writes the 10-kubeadm.conf drop in under this location.

That is actually the proper path indeed! if only because of loading order!

flavio:
So is everybody fine with me changing this PR to **remove** the code that writes the empty sysconfig file?

ereslibre:
> So is everybody fine with me changing this PR to remove the code that writes the empty sysconfig file?

I am.

Itxaka:
> So is everybody fine with me changing this PR to remove the code that writes the empty sysconfig file?

:+1:

flavio:
I've updated the PR to reflect what we've discussed above

flavio:
ping @ereslibre / @Itxaka

evrardjp:
woops, clicked too fast. Here is the comment I wanted to add on the review: "Not tested, looks okay according to my grep-ing skills."

Issue #792, opened 2019-10-23
command-line completion for skuba command


dannysauer:
### Summary

As a skuba user, I'd like the ability to press the completion key in my shell, and have command line arguments to skuba be expanded.  This would reduce the likelihood of misspelling long arguments, reduce the time it takes to check the "help" output, and generally provide a more polished feel.  The upstream commands (kubectl/kubeadm) have an option which emits completion code for shell inclusion, so this would bring us more feature parity as well.

As a developer, I'd like to know more about how skuba parses command line arguments.  Developing shell completion code will give me practical experience learning about argument parsing in skuba.

As a Bash user, I'd like to increase my depth of knowledge about how programmable completion works in Bash.

### Acceptance Criteria

- [x] stand-alone programmable completion for Bash is written in a way that allows relatively simple expansion as new CLI arguments are added
- [x] ticket to implement automated generation of Bash completion code directly within skuba is created
- [x] assessment of feasibility for also generating zsh completion code is documented in this ticket
- [ ] completion ideas (guess nodename based on target IP, predict hostnames from known hosts file or /etc/hosts, etc) which might take "longer than reasonable to implement" are documented in this ticket

kravciak:
You might be able to reuse some code from PR https://github.com/SUSE/skuba/pull/95

dannysauer:
Neat; there's a generator built in. Well, that counts as learning about the arg parser. :D

https://github.com/spf13/cobra/blob/master/bash_completions.go#L527

(clearly I should've searched closed issues too; thanks @kravciak)

dannysauer:
Marking off the checkboxes while I clean stuff up this morning:
 * This _is_ the ticket for developing the code now
 * Since it's built-in to Cobra, addition of new options is automatic
 * Feasibility of zsh code: 100% doable at the same time. :D

tdaines42:
I think the extra stuff should be it's own issue if it's something we want to do. Is there anything that needs to be done by me to get this done?

dannysauer:
I think we can probably leave enhancements for later, after people have used this and given feedback.

PR #829, opened 2019-11-15
skuba: Add new 'completion' command

tdaines42:
## Why is this PR needed?

As skuba is growing, new commands can be added. It's beneficial to
have a bash and zsh autocompletion commands which can be used to
generate a script which can autocomplete all skuba commands.

Fixes https://github.com/SUSE/skuba/issues/792

## What does this PR do?

Adds the commands `skuba completion bash` and `skuba completion zsh` that generate script for cli completion

Dockerfile that can be used to test:

```
FROM opensuse/leap

RUN zypper in -y git go1.12 make zsh bash-completion
ENV GOPATH=$HOME/go
ENV PATH=$GOPATH/bin:$PATH
RUN git clone https://github.com/tdaines42/skuba.git
RUN cd skuba && git checkout add-bash-completion && make install
RUN echo "source <( /go/bin/skuba completion bash | sed -n '/^#/,\$p' )" > /etc/bash_completion.d/skuba.sh
RUN skuba completion zsh | sed -n '/^#/,$p' > /etc/zsh_completion.d/_skuba
```

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/cmd/skuba/completion/bash_completions.go
+++ b/cmd/skuba/completion/bash_completions.go
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2019 SUSE LLC. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package completion
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+	"k8s.io/klog"
+)
+
+// NewBashCompletion creates a `skuba completion bash` cobra command
+func NewBashCompletion() *cobra.Command {
+	return &cobra.Command{
+		Use:   "bash",
+		Short: "Generates bash completion scripts",
+		// FIXME: sed can be dropped when stdout from main() is not polluted anymore
+		Long: `To manually load skuba bash completion, run the following command
+
+source <( ~/go/bin/skuba completion bash | sed -n '/^#/,$p' )`,
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := cmd.Parent().Parent().GenBashCompletion(os.Stdout); err != nil {
+				klog.Errorf("unable generate bash completion script: %s", err)
+			}
+		},
+	}
+}
--- a/cmd/skuba/completion/zsh_completion.go
+++ b/cmd/skuba/completion/zsh_completion.go
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2019 SUSE LLC. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package completion
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+	"k8s.io/klog"
+)
+
+// NewZshCompletion creates a `skuba completion zsh` cobra command
+func NewZshCompletion() *cobra.Command {
+	return &cobra.Command{
+		Use:   "zsh",
+		Short: "Generates zsh completion scripts",
+		// FIXME: sed can be dropped when stdout from main() is not polluted anymore
+		Long: `To manually load skuba zsh completion, run the following command
+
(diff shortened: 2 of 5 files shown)


tdaines42:
Based on code from https://github.com/SUSE/skuba/pull/95

tdaines42:
One thing I'm not sure about is changing the commands in the help messages to something like `skuba completion bash | tail -n +2 > "/etc/bash_completion.d/skuba.sh"` and `skuba completion zsh | tail -n +2 > "/etc/zsh_completion.d/_skuba"`

dannysauer:
I think we should package files in the *.d directories with the skuba command, and just remove that part from the help.

We really just need to sed that first line off of the generated output, using something like `$(skuba completion bash)` in the file. Then the file contents don't have to change.  Do you know how to get another couple of files into the rpm, @tdaines?

tdaines42:
> I think we should package files in the *.d directories with the skuba command, and just remove that part from the help.
> 
> We really just need to sed that first line off of the generated output, using something like `$(skuba completion bash)` in the file. Then the file contents don't have to change. Do you know how to get another couple of files into the rpm, @tdaines42?

I haven't had experience with packaging. I would assume there's something in IBS that could even generate the files after building skuba. 🤷‍♀

suseclee:
> > I think we should package files in the *.d directories with the skuba command, and just remove that part from the help.
> > We really just need to sed that first line off of the generated output, using something like `$(skuba completion bash)` in the file. Then the file contents don't have to change. Do you know how to get another couple of files into the rpm, @tdaines42?
> 
> I haven't had experience with packaging. I would assume there's something in IBS that could even generate the files after building skuba.

IMO, after this PR is merged,  we need to have another issue card for changing skuba spec file to apply this.

dannysauer:
The completion file really just needs to contain a single line in order to work.  In the case of bash, the /etc/bash_completion.d/skuba.sh file I was testing with looks like:
```bash
eval "$( ~sauer/go/bin/skuba completion bash | sed -n '/^#/,$p' )"
```
(print everything from the first comment line to the end of the file, for people who hate sed)

Though, the file should clearly use the right path for skuba. :D  That would be a nuiscance to generate in the spec file due to the quotes.  Perhaps we should just have the spec file run `skuba completion bash > skuba.sh` (probably also with the `sed` above, to clean up untagged versions) in the install stage and then copy skuba.sh into the right directory instead of the static file contents.  And repeat for zsh, of course.  That'd be pretty easy, and I think is what you're proposing, @tdaines42 

The only drawback is that it would potentially break if someone compiled a new skuba without installing the package, but do we care about that developers-only situation?  I'm thinking that we don't; people who break packages should probably expect breakage.

cmd/skuba/completion/bash_completions.go
```diff
@@ -0,0 +1,42 @@
...
+	"k8s.io/klog"
+)
+
+// NewBashCompletion creates a `skuba completion bash` cobra command
+func NewBashCompletion() *cobra.Command {
+	return &cobra.Command{
+		Use:   "bash",
+		Short: "Generates bash completion scripts",
+		// FIXME: tail -n +2 can be dropped when stdout from main() is not polluted anymore
+		Long: `To load skuba bash completion add the following to your ~/.bashrc or ~/.profile
```
dannysauer: If we generate the completion.d files, then we probably don't want people putting commands in their .profile / .bashrc.  Maybe something like:
```suggestion
		Long: `To manually load skuba bash completion, run the following command
```

cmd/skuba/completion/zsh_completion.go
```diff
@@ -0,0 +1,42 @@
...
+
+// NewZshCompletion creates a `skuba completion zsh` cobra command
+func NewZshCompletion() *cobra.Command {
+	return &cobra.Command{
+		Use:   "zsh",
+		Short: "Generates zsh completion scripts",
+		// FIXME: tail -n +2 can be dropped when stdout from main() is not polluted anymore
+		Long: `To load skuba zsh completion run the following command
+
+skuba completion zsh | tail -n +2 > "${fpath[1]}/_skuba"`,
```
dannysauer: Same comments as above in the bash section, re "manually add" and using the `sed`

dannysauer:
Should Marcos be removed from the authors in the commit so we're "all green" on the contributor check, or do we not care about that (since he did, technically, write some of the code - and was authorized to do so at the time).

cmd/skuba/completion/bash_completions.go
```diff
@@ -0,0 +1,42 @@
...
+
+. <(skuba completion bash | tail -n +2)`,
```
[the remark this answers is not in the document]
tdaines42: I updated the comment to match this

cmd/skuba/completion/bash_completions.go
tdaines42: same

cmd/skuba/completion/zsh_completion.go
tdaines42: updated this one as well

tdaines42:
I made some changes and updated the dockerfile to reflect that

tdaines42:
> LGTM.
> 
> Though, A dockerfile is mentioned, but I don't see it in the commit - is that intentional?

It's in the comment for the PR for manual testing

dannysauer:
I updated the skuba spec file in my skuba branch at https://build.suse.de/package/show/home:dannysauer:branches:Devel:CaaSP:4.0/skuba to build bash and zsh completion files.  When this gets merged, that spec should create two new packages - skuba-bash-completion and skuba-zsh-completion.

dannysauer:
Binary packages at https://build.suse.de/package/binaries/home:dannysauer:branches:Devel:CaaSP:4.0/skuba/SLE_15_SP1

PR #833, opened 2019-11-19
Add DoesResourceExistWithError unit-test

c3y1huang:
## Why is this PR needed?

Increase unit test code coverage
Ref: https://github.com/SUSE/avant-garde/issues/744

## What does this PR do?

Test includes:
* secret exist with no error
* secret not exist
* secret exist in other namespace

```
github.com/SUSE/skuba/internal/pkg/skuba/kubernetes     0.024s  coverage: 27.9% of statements
```

```
github.com/SUSE/skuba/internal/pkg/skuba/kubernetes     0.040s  coverage: 29.8% of statements
```

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubernetes/resources_test.go
+++ b/internal/pkg/skuba/kubernetes/resources_test.go
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2019 SUSE LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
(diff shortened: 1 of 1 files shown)


internal/pkg/skuba/kubernetes/resources_test.go
```diff
@@ -0,0 +1,152 @@
...
+		fakeClientset    *fake.Clientset
+		searchNamespace  string
+		searchSecret     string
+		expectReturn     bool
+		expectErrMessage error
+		fakeErrMessage   error
+	}{
+		{
+			name: "secret exist with no error",
+			fakeClientset: fake.NewSimpleClientset(
```
innobead: The same `fakeClientset` exists in three test data, I would suggest to create one on the top then reuse to avoid duplicated creation.

internal/pkg/skuba/kubernetes/resources_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func Test_DoesSecretResourceExistWithError(t *testing.T) {
```
jenting: Minor: naming `Test_DoesResourceExistWithError` makes easier to find which function on the test.

PR #836, opened 2019-11-19
CI: Remove junit log collection from e2e tests

pablochacin:
## Why is this PR needed?

e2e test are not configured to generated junit output.
This causes the pipeline to fail when trying to collect
this log.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


pablochacin:
@chentex IIRC @tdaines42 added this functionality for making easier to analyze the reports from the e2e tests. AFIAK no one is actually doing so.

PR #826, opened 2019-11-14
Kubernetes 1.16

ereslibre:
## Why is this PR needed?

This PR includes Kubernetes 1.16 into the supported versions map.

Fixes https://github.com/SUSE/avant-garde/issues/1023

**WIP** because I'm testing it at the moment.

## Info for QA

It's necessary to check that it's possible to bootstrap a cluster with 1.15, 1.16 and that the migration works without any issues.

## Docs

Tracked in https://github.com/SUSE/avant-garde/issues/934

# Merge restrictions

(Please do not edit this)

We are in *v4-maintenance phase*, so we will restrict what can be merged to prevent unexpected surprises:

    What can be merged (merge criteria):
        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine
        there is a PR for updating documentation (or a statement that this is not needed)

PR #837, opened 2019-11-19
Evolve settings on upgrade

ereslibre:
Supersedes: https://github.com/SUSE/skuba/pull/803
Reintroduces: https://github.com/SUSE/skuba/pull/697
Fixes: https://github.com/SUSE/avant-garde/issues/1016

## Why is this PR needed?

This changeset moves away from templating for `kubeadm-init.conf` and `kubeadm-join.conf.d`'s `master.conf.template` and `worker.conf.template`.

The idea behind this changeset is that we have a base configuration, set in terms of internal kubeadm API's, avoiding many templating logic. Instead, we will run mutations on top of that configuration.

Mutations will happen on two places mainly:

* During `init`: when `skuba cluster init` is executed, we will modify the default init, cluster and join configuration depending on the settings that were passed to `skuba cluster init`. When we are done with all tweaks, we will directly marshal the kubeadm types to disk, instead of relying on text templating.

* During `upgrade apply`: when `skuba node upgrade apply` is executed on the first control plane instance, the cluster configuration will be downloaded, unmarshaled, mutated, `kubeadm upgrade apply --config` with the mutated configuration, what will in turn re-upload the mutated `ClusterConfiguration` to the cluster, so subsequent `skuba node upgrade apply` (that will in reality be called internally as `kubadm upgrade node`), will use this mutated configuration.

So, for example, the admission plugins setting that was introduced recently will be applied to new clusters, but will also be applied to existing clusters upon `skuba node upgrade apply`, when we run that command on the first control plane instance of the cluster.

Also, there was a difference between AWS and OpenStack integrations in which AWS was rendering all required configuration during `cluster init`, whereas OpenStack integration was in contrast overriding some configurations during `skuba node bootstrap`. This change also modifies this behavior, so both behave as AWS, rendering all configuration on `skuba cluster init` on disk.

## Info for QA

This is an internal refactor to make easier for configuration changes to be applied to existing clusters in a way that is reasonable to track and maintain.

Ideally this PR needs to be tested in many contexts: deploying new clusters (with, and without cloud integration), and also upgrading existing clusters, so we make sure that the mutation logic is working as intended.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


ereslibre:
Waiting for reviews. This is ready.

New cluster creation tested (without cloud integration), and upgrade from 1.15 upgrade to 1.16 tested as well (along with https://github.com/SUSE/skuba/pull/826), without cloud integration. Both worked as expected.

pkg/skuba/actions/cluster/init/init.go
```diff
@@ -176,3 +195,173 @@ func renderTemplate(templateContents string, initConfiguration InitConfiguration
...
+	}
+	joinCfgContents, err := kubeadmutil.MarshalToYamlForCodecs(&joinCfg, schema.GroupVersion{
+		Group:   "kubeadm.k8s.io",
+		Version: kubeadm.GetKubeadmApisVersion(initConfiguration.KubernetesVersion),
+	}, kubeadmscheme.Codecs)
+	if err != nil {
+		return err
+	}
+	if err := ioutil.WriteFile(skuba.WorkerConfTemplateFile(), joinCfgContents, 0600); err != nil {
+		return errors.Wrap(err, "error writing master join configuration")
```
jenting: worker :)

internal/pkg/skuba/kubeadm/configmap.go
```diff
@@ -114,3 +117,46 @@ func RemoveAPIEndpointFromConfigMap(client clientset.Interface, node *v1.Node) e
...
+		"ServiceAccount",
+		"TaintNodesByCondition",
+		"Priority",
+		"DefaultTolerationSeconds",
+		"DefaultStorageClass",
+		"PersistentVolumeClaimResize",
+		"MutatingAdmissionWebhook",
+		"ValidatingAdmissionWebhook",
+		"ResourceQuota",
+	)
```
jenting: https://github.com/kubernetes/website/blob/master/content/en/docs/reference/access-authn-authz/admission-controllers.md#which-plugins-are-enabled-by-default

Any reason missing StorageObjectInUseProtection and RuntimeClass?

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -96,7 +96,9 @@ func Apply(clientSet clientset.Interface, target *deployments.Target) error {
 			if err := node.AddTargetInformationToInitConfigurationWithClusterVersion(target, initCfg, nodeVersionInfoUpdate.Update.APIServerVersion); err != nil {
 				return errors.Wrap(err, "error adding target information to init configuration")
 			}
-			kubeadm.SetContainerImagesWithClusterVersion(initCfg, nodeVersionInfoUpdate.Update.APIServerVersion)
+			if err := kubeadm.UpdateClusterConfigurationWithClusterVersion(initCfg, nodeVersionInfoUpdate.Update.APIServerVersion); err != nil {
+				return errors.Wrap(err, "error mutating the cluster configuration")
+			}
 			initCfgContents, err = kubeadmconfigutil.MarshalInitConfigurationToBytes(initCfg, schema.GroupVersion{
 				Group:   "kubeadm.k8s.io",
 				Version: "v1beta2",
```
jenting: Does this needs to change as `Version: kubeadm.GetKubeadmApisVersion(currentClusterVersion),`?

internal/pkg/skuba/kubeadm/configmap.go
ereslibre: Thank you, added them as a separate commit with an explanation of both.

pkg/skuba/actions/node/upgrade/apply.go
ereslibre: Updated, `kubeadm` is the first component to be upgraded that will use this configuration, so it's even safer to retrieve the version according to the "future" version of the cluster instead of the current one.

internal/pkg/skuba/kubeadm/configmap.go
```diff
@@ -114,3 +117,48 @@ func RemoveAPIEndpointFromConfigMap(client clientset.Interface, node *v1.Node) e
...
+	if err := setApiserverAdmissionPlugins(initConfiguration); err != nil {
+		return err
+	}
+	if err := setContainerImagesWithClusterVersion(initConfiguration, version); err != nil {
+		return err
+	}
+	return nil
+}
+
+func setApiserverAdmissionPlugins(initConfiguration *kubeadmapi.InitConfiguration) error {
```
MaximilianMeister: do we need to return an `error` here?

internal/pkg/skuba/kubeadm/images.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/SUSE/skuba/pkg/skuba"
 )
 
-func SetContainerImagesWithClusterVersion(initConfiguration *kubeadmapi.InitConfiguration, clusterVersion *version.Version) {
+func setContainerImagesWithClusterVersion(initConfiguration *kubeadmapi.InitConfiguration, clusterVersion *version.Version) error {
```
MaximilianMeister: also do we need to return an `error` here?

internal/pkg/skuba/kubeadm/configmap.go
ereslibre: Thank you, addressed.

internal/pkg/skuba/kubeadm/images.go
ereslibre: Thanks! Addressed.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -96,7 +96,7 @@ func Apply(clientSet clientset.Interface, target *deployments.Target) error {
 			if err := node.AddTargetInformationToInitConfigurationWithClusterVersion(target, initCfg, nodeVersionInfoUpdate.Update.APIServerVersion); err != nil {
 				return errors.Wrap(err, "error adding target information to init configuration")
 			}
-			kubeadm.SetContainerImagesWithClusterVersion(initCfg, nodeVersionInfoUpdate.Update.APIServerVersion)
+			kubeadm.UpdateConfigurationWithClusterVersion(initCfg, nodeVersionInfoUpdate.Update.APIServerVersion)
```
drpaneas: If you decide to keep the `UpdateConfigurationWithClusterVersion` function to return `error` then it makes sense to check for it as well. I would be in favor of dropping all of your `error` returns in the following functions:

- `UpdateConfigurationWithClusterVersion`
- `setContainerImagesWithClusterVersion`
- `setContainerImagesWithClusterVersion`

this is because I think they will always return `nil`.

internal/pkg/skuba/kubeadm/configmap.go
```diff
@@ -133,11 +134,29 @@ func setApiserverAdmissionPlugins(initConfiguration *kubeadmapi.InitConfiguratio
 	if initConfiguration.APIServer.ControlPlaneComponent.ExtraArgs == nil {
 		initConfiguration.APIServer.ControlPlaneComponent.ExtraArgs = map[string]string{}
 	}
+	admissionPlugins := []string{}
+	if len(initConfiguration.APIServer.ExtraArgs["enable-admission-plugins"]) > 0 {
+		admissionPlugins = strings.Split(initConfiguration.APIServer.ExtraArgs["enable-admission-plugins"], ",")
```
drpaneas: this looks like a good place for simple *unit-test*.

internal/pkg/skuba/util/slices.go
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package util
+
+func UniqueStringSlice(slice []string) []string {
```
drpaneas: please provide a `// comment` description for the function

internal/pkg/skuba/util/slices.go
```diff
@@ -0,0 +1,30 @@
...
+	result := []string{}
+	seen := make(map[string]struct{}, len(slice))
+	for _, value := range slice {
+		if _, ok := seen[value]; !ok {
+			seen[value] = struct{}{}
```
drpaneas: Can you explain a bit what this `{}{}` is supposed to mean? Looks like an arcane Perl spell script to me :P

drpaneas:
LGTM for me as well, I just pointed some minor comments. But it looks good. Thanks for the great work @ereslibre  (writing tests as well, kudos) on this :dagger:

internal/pkg/skuba/util/slices.go
bergmannf: I assume it is for memory optimization - it seems a empty `struct{}` consumes no memory, whereas a `bool` would consume 1 byte already to track if a string was already seen before.

pkg/skuba/actions/node/upgrade/apply.go
ereslibre: I think you reviewed an older version here, I updated that already :), not your fault, I've been force pushing a lot lately on this one!

internal/pkg/skuba/util/slices.go
ereslibre: Yes, exactly, in this case we are only interested in `seen` values, that are keys in the map, the values of those keys are irrelevant, that's why an empty struct pays the bill. `struct{}{}` is just another way of saying `make(struct{})`.

internal/pkg/skuba/kubeadm/configmap.go
ereslibre: Adding that one, yup!

internal/pkg/skuba/kubeadm/configmap.go
```diff
@@ -114,3 +117,44 @@ func RemoveAPIEndpointFromConfigMap(client clientset.Interface, node *v1.Node) e
...
+		"DefaultStorageClass",
+		"PersistentVolumeClaimResize",
+		"MutatingAdmissionWebhook",
+		"ValidatingAdmissionWebhook",
+		"ResourceQuota",
+		"StorageObjectInUseProtection",
+	)
+	if clusterVersion.AtLeast(version.MustParseSemantic("1.16.0")) {
+		admissionPlugins = append(admissionPlugins, "RuntimeClass")
+	}
```
jenting: 💯

internal/pkg/skuba/kubeadm/configmap.go
ereslibre: Adding that one, yup!

ereslibre:
> Do we need docs for that? or just QA?

I don't think we need public docs for this one, in the sense that is enabling us to evolve the cluster configuration over time. QA can happen post-merge along with the 1.16 PR that was merged already.

ereslibre:
Added some tests to the kubeadm configmap logic as requested by @drpaneas, this should be good to go :)

evrardjp:
I sure hope there was no race condition between the force push and @drpaneas review :)
One must love github review system... Ahem.

internal/pkg/skuba/kubeadm/configmap_test.go
```diff
@@ -0,0 +1,97 @@
...
+				currentAdmissionPlugins := strings.Join(tt.currentAdmissionPlugins, ",")
+				if initCfg.APIServer.ControlPlaneComponent.ExtraArgs == nil {
+					initCfg.APIServer.ControlPlaneComponent.ExtraArgs = map[string]string{}
+				}
+				initCfg.APIServer.ControlPlaneComponent.ExtraArgs["enable-admission-plugins"] = currentAdmissionPlugins
+			}
+			UpdateClusterConfigurationWithClusterVersion(&initCfg, tt.clusterVersion)
+			// Check admission plugins
+			gotAdmissionPlugins := initCfg.APIServer.ControlPlaneComponent.ExtraArgs["enable-admission-plugins"]
+			if gotAdmissionPlugins != expectedAdmissionPlugins {
```
Itxaka: any reason we are not using `assert` here and manually checking and raising the errors?

ereslibre: For `assert` we'd need an external package

evrardjp:
I like the concept, thought I am not savvy enough yet to vote positively on this. Most of the commits looks fine at first sight though.

internal/pkg/skuba/kubeadm/configmap_test.go
Itxaka: :o I would have expected assert to be part of the standard lib!

Thanks!

internal/pkg/skuba/kubeadm/configmap.go
```diff
@@ -114,3 +117,44 @@ func RemoveAPIEndpointFromConfigMap(client clientset.Interface, node *v1.Node) e
...
+func setApiserverAdmissionPlugins(initCfg *kubeadmapi.InitConfiguration, clusterVersion *version.Version) {
+	if initCfg.APIServer.ControlPlaneComponent.ExtraArgs == nil {
+		initCfg.APIServer.ControlPlaneComponent.ExtraArgs = map[string]string{}
+	}
+	admissionPlugins := []string{}
+	if len(initCfg.APIServer.ExtraArgs["enable-admission-plugins"]) > 0 {
+		admissionPlugins = strings.Split(initCfg.APIServer.ExtraArgs["enable-admission-plugins"], ",")
+	}
+	// List of recommended plugins: https://git.io/JemEu
+	admissionPlugins = append(admissionPlugins,
```
evrardjp: nit:

else?

I am not sure if it's fine to append to an existing list of admissionPlugins in the case you already have them.
Or maybe my brain farts, and I should totally go learn go.

Itxaka:
lgtm at least code-wise seems really good! Not sure about the whole upgrade+init thing as I miss some context in there, but +1!

internal/pkg/skuba/kubeadm/configmap.go
ereslibre: Can you please elaborate? not sure I'm following.

jenting: @evrardjp This function call deletes duplicate admission plugins https://github.com/SUSE/skuba/blob/fa14c5c46877ffe4231909d3cfacc85588b03752/internal/pkg/skuba/kubeadm/configmap.go#L158

Itxaka: shouldnt it be easier to have an `AppendNoDuplicates` so we only append if not in there?

Maybe its even more complicated of a func than the `UniqueStringSlice`, probably needs to iterate over the slice for each element, which `UniqueStringSlice` seems to iterate just once over the slice

Go its weird :)

evrardjp: Discussed outside this, and it's basically the intended behaviour. Good then!

ereslibre:
Had to resolve conflicts with latest `master` and had to push again.

PR #843, opened 2019-11-20
Release: check tags from multiple branches

jordimassaguerpla:
## Why is this PR needed?

Initially we were releasing only from master branch.
However, in order not to block master branch, we are now also releasing
from branches called release-caasp-x.y.z

The script that was creating the changelog was taking commits that have
been merged into the master branch and so it was including entries that
have already been released.

Removing the "--merge" option from the git command will take into
account all versions.

Given we are not maintaining more than one version, this will work.

## What does this PR do?

See above

Nothing else

## Info for QA

N/A

The generated changelog was including commits from previous releases.

Include only commits that are new since the last release.

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/packaging/suse/changelog_maker.sh
+++ b/ci/packaging/suse/changelog_maker.sh
@@ -7,13 +7,13 @@ changes=${1:-skuba.changes.append}
 
 [ -f "${changes}" ] && rm "${changes}"
 
-mapfile -t tags < <( git tag --merged | sort -rV | head -n2 )
-scope="${tags[1]}..${tags[0]}"
+mapfile -t tags < <( git tag | sort -rV | head -n2 )
+scope="${tags[1]}...${tags[0]}"
 
 {
     git show --format="${header}%n%n- Update to ${tags[0]}:" \
         --date="format-local:${datef}" -s "${tags[0]}^{commit}"
-    git log -s --format="%w(77,2,12)* %h %s" --no-merges "${scope}"
+    git log -s --cherry-pick --format="%w(77,2,12)* %h %s" --no-merges "${scope}"
     # Add empty line
     echo
 } >> "${changes}"


jordimassaguerpla:
> While I believe this change makes sense if we consider a skuba to be released in a rolling release fashion, understanding that the release branches are unlikely to be touched after the official release is approved.
> 
> However note that this change does not prevent including the backports that were already included in any release branch. For instance v1.1.2 generated changelog will include the three commits that were backported in v1.1.1 during the QA testing. At least this is the behavior I am experimenting in my host. I could not find an easy way to omit them with a one liner git command, even looks like it should be possible according the `man git-log` docs by using the `--cherry-pick` flag, there might be something I do not properly understand.

We could look for duplicates using patch-ids.

If we do this:

`git log -s --format="%w(77,2,12)* %h %s" --no-merges v1.1.1...v1.1.2 | grep autoyast`

we see
```

  * 849959df Add release packages for autoyast
  * 7fbace5a Add release packages for autoyast
```

This is because we actually have the same commit in both branches but with different commit-ids.

However, they have the same patch-id:

```
git show 849959df | git patch-id
572e493023f55b43952998d8be9f8aae3823b2f3 849959df9a978865781540bdc2a562b8d9af4b7a
```

```
git show 7fbace5a | git patch-id
572e493023f55b43952998d8be9f8aae3823b2f3 7fbace5a66392be1196ac65978286f32c636ee85

```

I think we could use this to filter commits that have a duplicated patchid.

Makes sense?

ci/packaging/suse/changelog_maker.sh
```diff
@@ -7,7 +7,7 @@ changes=${1:-skuba.changes.append}
 
 [ -f "${changes}" ] && rm "${changes}"
 
-mapfile -t tags < <( git tag --merged | sort -rV | head -n2 )
+mapfile -t tags < <( git tag | sort -rV | head -n2 )
 scope="${tags[1]}..${tags[0]}"
```
davidcassany: Please change that to

```bash
scope="${tags[1]}...${tags[0]}"
```

ci/packaging/suse/changelog_maker.sh
```diff
@@ -7,7 +7,7 @@ changes=${1:-skuba.changes.append}
...
 
 {
```
davidcassany: Also change the git-log command from:

```bash
    git log -s --format="%w(77,2,12)* %h %s" --no-merges "${scope}"
```

to:

```bash
    git log -s --cherry-pick --format="%w(77,2,12)* %h %s" --no-merges "${scope}"
```

jordimassaguerpla:
> > We could look for duplicates using patch-ids.
> 
> Well this is what `--cherry-pick` flag is supposed to do
> 
> > If we do this:
> > `git log -s --format="%w(77,2,12)* %h %s" --no-merges v1.1.1...v1.1.2 | grep autoyast`
> > we see
> > ```
> > 
> >   * 849959df Add release packages for autoyast
> >   * 7fbace5a Add release packages for autoyast
> > ```
> 
> God! You gave me the key joy_cat I never saw those duplicated entries in my tests, neither never appeared in the output of `make suse-changelog`, just one was there. I checked whats different from your command and I realized you are using three dots `...` between tags, while in the script there are only two `..`!! argh! Why is git not failing if the syntax is wrong astonished !!
> 
> If we use `...` then the `--cherry-pick` command does exactly what it is expected to do.

.. or ... have different meaning but both are accepted. This is why it does not complain. I can't remember the difference thought.

davidcassany:
> .. or ... have different meaning but both are accepted. This is why it does not complain. I can't remember the difference thought.

You are right:

```
   Dotted Range Notations
       The .. (two-dot) Range Notation
           The ^r1 r2 set operation appears so often that there is a shorthand for it. When you have two commits r1 and r2 (named according to the syntax explained in SPECIFYING REVISIONS above), you can ask for commits that are reachable
           from r2 excluding those that are reachable from r1 by ^r1 r2 and it can be written as r1..r2.

       The ... (three-dot) Symmetric Difference Notation
           A similar notation r1...r2 is called symmetric difference of r1 and r2 and is defined as r1 r2 --not $(git merge-base --all r1 r2). It is the set of commits that are reachable from either one of r1 (left side) or r2 (right side)
           but not from both.
```

from `man gitrevisions`. Damn this syntax is really non intuitive :smile:

jordimassaguerpla:
@davidcassany : I did the changes you suggested and rebased everything. Could you please review again?

jordimassaguerpla:
@dannysauer @cclhsu Could you also review? Thanks.

PR #846, opened 2019-11-21
Add git pre-commit checks

jenting:
## Why is this PR needed?

When file PR for review, the CI runs linter check.
It could change to locally before the user executes `git commit` to do some basic check.

## What does this PR do?

Add file `.pre-commit-config.yaml` and when the user executes `make pre-commit-install`, it will generates file `.git/hooks/pre-commit` to perform code linter check.

So later when the user executes `git commit` command, it will do the corresponding checks described in `.pre-commit-config.yaml`.

The installed hooks could be uninstalled `make pre-commit-uninstall` if the user wants to remove the git hooks locally.

This PR adds the checks based on the skuba Makefile current linter, we could increase/shrink checks in the future.
https://github.com/SUSE/skuba/blob/72adcddbcf690582448fd3ff4924c8675cf593f4/Makefile#L93-L100

By the way, git hooks cannot be force installed when cloning a repo to local. Therefore, the pre-commit tools need to manually installed when the user executes `make pre-commit-install`.

References
- https://pre-commit.com
- https://pre-commit.com/hooks.html

Here is the sample output when executes `git commit` command.
image

Demo:
- Golang: https://asciinema.org/a/uGChT4WTdptlZkbrGLJYiGu0T
- Python: https://asciinema.org/a/bjq8Ba5Nr0MvdzS5uYbxuF7Se
- Terraform: https://asciinema.org/a/fa9iLfCyJzZRdpyNI5FswjaGw

## Info for QA

N/A

### Related info

N/A

N/A

N/A

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


Itxaka:
Looks good but jenkins failed with:

```
test -f /home/jenkins/workspace/p-jobs_caasp-v4-code-lint_PR-846/bin/bin/pre-commit || curl -sfL https://pre-commit.com/install-local.py | HOME=/home/jenkins/workspace/p-jobs_caasp-v4-code-lint_PR-846/bin python -

/bin/sh: python: command not found
```

Maybe you need to specify `python3` instead of plain python?

Makefile
```diff
@@ -106,6 +106,16 @@ deps:
...
 	test -f $(BINPATH)/shellcheck || curl -sfL "https://storage.googleapis.com/shellcheck/shellcheck-v0.4.7.linux.x86_64.tar.xz" | tar -xJv --strip-components=1 -C $(BINPATH)
 
+.PHONY: pre-commit-install
+pre-commit-install:
+	test -f $(BINPATH)/bin/pre-commit || curl -sfL https://pre-commit.com/install-local.py | HOME=$(BINPATH) python -
+	$(BINPATH)/bin/pre-commit install -t pre-commit -t commit-msg
+
+.PHONY: pre-commit-uninstall
+pre-commit-uninstall:
+	test -f $(BINPATH)/bin/pre-commit || curl -sfL https://pre-commit.com/install-local.py | HOME=$(BINPATH) python -
```
aojea: Is it possible to pin to a specific version to avoid issues with new releases?

jenting: For `install-local.py`, it does not support pass specific version when pip install.

Other installation ways are https://pre-commit.com/#install

Any better idea?

aojea: it was mostly curiosity, not a big deal :sweat_smile: 
it's fine with me, but I defer to others the approval
Thanks for the improvement

Itxaka: if pre-commit is not installed and we are calling uninstall, why install and uninstall it? we should just exit directly

scripts/commit-msg.sh
```diff
@@ -0,0 +1,37 @@
...
+      exit 1
+    fi
+  else
+    # Any line in body must not exceed 72 characters
+    if [ $length -gt 72 ]; then
+      echo "The line \"$line\" exceeds 72 characters."
+      exit 1
+    fi
+  fi
+done < "$1"
```
ereslibre: Given this check is not `Required` I would remove it from here honestly. It's sometimes very strict (e.g. if we add a URL inside the body it will very likely go over 72 characters, or if the subject is 51 characters long because it cannot be shorter...).

Makefile
jenting: The purpose is to remove the .git/hook/pre-commit which is installed by pre-commit, maybe the better way is to delete the file `.git/hook/pre-commit` directly.

PR #850, opened 2019-11-22
CI: Increase delay after joining masters

pablochacin:
## Why is this PR needed?

Several e2e tests fail due to problems derived from etcd not ready after joining a master node. 

## What does this PR do?

Increase the time between joining a master and joining
another node, to allow etcd to stabilize.

It is a followup of https://github.com/SUSE/skuba/pull/844 where the delay was introduced, increasing its value.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


pablochacin:
> this /lgtm as long as we know it's just that etcd needs a little more time and not just a random thing.

The problem we don't know how much time it needs. It depends on many factors. I could rise the time to 5 minutes, but this makes each test last 15 minutes more. So I'm trying to find a number that works reliably without introducing too much delay for tests.

PR #851, opened 2019-11-25
Add node relate unit-tests.

c3y1huang:
## Why is this PR needed?

Increase unit test code coverage.
Ref: https://github.com/SUSE/avant-garde/issues/744

## What does this PR do?

New Tests Include:
* GetControlPlaneNodes
  * get control plane node when cluster has 1 master.
  * get control plane node when cluster has 2 master.
  * get control plane node when cluster has 3 master.
  * get control plane node when cluster has 3 master, 1 worker.
* GetNodeWithMachineID
  * get node name matching master node machine ID.
  * get node name matching worker node machine ID.
  * get node name when machin ID does not exist.
* DrainNode
  * drain master node by eviction.
  * drain worker node by eviction.
  * drain worker node by deletion.
  * drain when node does not exist.
  * drain by eviction when pod does not exist.
  * drain by deletion when pod does not exist.
* getPodContainerImageTag
  * get pod container image tag with pod not exist.
  * get pod container image tag with pod exist in different namespace.

Lint Changes:
* Add comments to some exported functions.
* Change machineId to machineID.

Other Changes:
* Function change to use clientset interface.
* Uniform clientset interface name.
* Move getPodContainerImageTag to pod.go.

github.com/SUSE/skuba/internal/pkg/skuba/kubernetes/nodes.go (16.7%)

github.com/SUSE/skuba/internal/pkg/skuba/kubernetes/nodes.go (89.7%)

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubernetes/nodes_test.go
+++ b/internal/pkg/skuba/kubernetes/nodes_test.go
@@ -18,37 +18,326 @@
 package kubernetes
 
 import (
-	"reflect"
+	"fmt"
 	"testing"
 
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/fake"
+	ktest "k8s.io/client-go/testing"
 )
 
-func TestGetPodContainerImageTag(t *testing.T) {
-	podName := "etcd-my-master-0"
-	namespace := metav1.NamespaceSystem
-	expectedEtcdContainerImageTag := "3.3.11"
-	t.Run("get pod container image tag with clientset", func(t *testing.T) {
-		clientset := fake.NewSimpleClientset(&v1.Pod{
+var m1 = createNode("master1", true, "m1")
+var m2 = createNode("master2", true, "m2")
+var m3 = createNode("master3", true, "m3")
+var w1 = createNode("worker1", false, "w1")
+var w2 = createNode("worker2", false, "w2")
+var w3 = createNode("worker3", false, "w3")
+var invalid = createNode("invalid", false, "invalid")
+
+func addEvictionSupport(fakeClientset *fake.Clientset) {
+	podsEviction := metav1.APIResource{
+		Name:    "pods/eviction",
+		Kind:    "Eviction",
+		Group:   "",
+		Version: "v1",
+	}
+	coreResources := &metav1.APIResourceList{
+		GroupVersion: "v1",
+		APIResources: []metav1.APIResource{podsEviction},
+	}
+
+	policyResources := &metav1.APIResourceList{
+		GroupVersion: "policy/v1",
+	}
+	fakeClientset.Resources = append(fakeClientset.Resources, coreResources, policyResources)
+}
+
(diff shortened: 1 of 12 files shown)


internal/pkg/skuba/kubernetes/nodes_test.go
```diff
@@ -18,37 +18,409 @@
...
+					t.Errorf("error expected on %s, but no error reported", tt.name)
+					return
+				}
+				if err.Error() != tt.expectErrMsg {
+					t.Errorf("returned error (%v) does not match the expected one (%v)", err.Error(), tt.expectErrMsg)
+					return
+				}
+				return
+			}
+			if tt.expectErrMsg == "" {
```
jenting: This if case does not need since above has returned. [1]

internal/pkg/skuba/kubernetes/nodes_test.go
```diff
@@ -18,37 +18,409 @@
...
```
jenting: [1]

internal/pkg/skuba/kubernetes/nodes_test.go
```diff
@@ -18,37 +18,409 @@
...
+	}
+}
+
+func Test_getPodContainerImageTag(t *testing.T) {
+	fakeClientset := fake.NewSimpleClientset(
+		&corev1.NodeList{
+			Items: []corev1.Node{
+				m1, w1,
+			},
+		},
```
jenting: For testing pod, node information could be optional.

internal/pkg/skuba/kubernetes/nodes_test.go
```diff
@@ -18,37 +18,409 @@
...
+			namespace:    metav1.NamespaceSystem,
+			podName:      "invalid",
+			expectErrMsg: "could not retrieve pod object: pods \"invalid\" not found",
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt // Parallel testing
+		t.Run(tt.name, func(t *testing.T) {
+			actual, err := getPodContainerImageTag(fakeClientset, tt.namespace, tt.podName)
```
jenting: Not about this PR. But it will be great to improve it.
If possible, it would better creates two files `pod.go` and `pod_test.go` and move functions `getPodContainerImageTag` and `Test_getPodContainerImageTag` into it :)

internal/pkg/skuba/kubernetes/nodes_test.go
```diff
@@ -18,37 +18,409 @@
...
+										Name:  "test",
+										Image: "registry.suse.com/caasp/v4/test:1.1.1",
+									},
+								},
+							},
+						},
+					},
+				}
+				return true, obj, nil
+			})
```
jenting: To be honest, I don't fully understand what this purpose for. But I will not block this PR.

c3y1huang: The kubernetes fake clientset doesn't handle *selector as expected when called constraints it does not support. 
Instead of returning pods on specific node or errors, it will return all pods as there are no constraints. This leads to a false result when testing case such `drain by eviction when pod does not exist` where expecting errors. 

The address to issue, [here](https://github.com/kubernetes/client-go/issues/326) suggested to use reactor where can implement the expect matching behavior.

internal/pkg/skuba/kubernetes/nodes_test.go
c3y1huang: Good suggestion.

PR #857, opened 2019-11-27
Enhance skuba auth login error message (bsc#1157802)

jenting:
## Why is this PR needed?

Enhance error message when the user inputs incorrect OIDC dex server URL.

Fixes https://github.com/SUSE/avant-garde/issues/1092

appear in the changelog.

## What does this PR do?

Function `oidc.NewProvider` will do HTTP Get method to OIDC dex URL `.well-known/openid-configuration`  to scrapes the basic OIDC dex server information. We could check on this URL is accessible or not to verify the user inputs the correct OIDC dex server URL.

N/A

## Info for QA

### Related info

N/A

The user inputs the OIDC dex server URL to gangway URL
```
skuba auth login -s https://10.86.3.10:32001 --insecure
Enter your username:
Enter your password:
```

Output error message:
```
F1127 15:06:34.778465   13400 login.go:88] error on login: auth failed: failed to query provider "https://10.86.3.10:32001": oidc: failed to decode provider discovery object: expected Content-Type = application/json, got "text/html; charset=utf-8": invalid character '<' looking for beginning of value
```

The user inputs the OIDC dex server URL to gangway URL
```
skuba auth login -s https://10.86.3.10:32001 --insecure
Enter your username:
Enter your password:
```

Output error message:
```
F1127 15:06:56.026300   13428 login.go:88] error on login: auth failed: failed to query provider https://10.86.3.10:32001 (is this the right URL? maybe missing --root-ca or --insecure, or incorrect port number?)
```

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/pkg/skuba/actions/auth/auth.go
+++ b/pkg/skuba/actions/auth/auth.go
@@ -39,6 +39,7 @@ import (
 	"github.com/pkg/errors"
 	"golang.org/x/net/html"
 	"golang.org/x/oauth2"
+	"k8s.io/klog"
 )
 
 const (
@@ -186,14 +187,15 @@ func doAuth(authReq request) (*response, error) {
 	ctx := oidc.ClientContext(context.Background(), client)
 	provider, err := oidc.NewProvider(ctx, authReq.IssuerURL)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to query provider %q", authReq.IssuerURL)
+		klog.Errorf("failed to query provider %s: %v", authReq.IssuerURL, err)
+		return nil, fmt.Errorf("failed to query provider %s (is this the right URL? maybe missing --root-ca or --insecure, or incorrect port number?)", authReq.IssuerURL)
 	}
 
 	var s struct {
 		ScopesSupported []string `json:"scopes_supported"`
 	}
 	if err := provider.Claims(&s); err != nil {
-		return nil, errors.Wrapf(err, "failed to parse provider scopes_supported")
+		return nil, errors.Wrap(err, "failed to parse provider scopes_supported")
 	}
 
 	authReq.provider = provider
@@ -206,13 +208,13 @@ func doAuth(authReq request) (*response, error) {
 	authCodeURL := oauth2Config(authReq).AuthCodeURL("", oauth2.AccessTypeOffline)
 	resp, err := client.Get(authCodeURL)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed on get auth code url")
+		return nil, errors.Wrap(err, "failed on get auth code url")
 	}
 	defer resp.Body.Close()
 
 	respDump, err := httputil.DumpResponse(resp, true)
 	if err != nil {
-		return nil, errors.Wrapf(err, "http dump response")
+		return nil, errors.Wrap(err, "http dump response")
 	}
 
 	loginURL := authReq.IssuerURL
@@ -232,7 +234,7 @@ func doAuth(authReq request) (*response, error) {
 			reader := bufio.NewReader(os.Stdin)
 			authConnector, err := reader.ReadString('\n')
 			if err != nil {
-				return nil, errors.Wrapf(err, "read user input")
+				return nil, errors.Wrap(err, "read user input")
 			}
 			authReq.AuthConnector = strings.TrimSpace(authConnector)
 		}
@@ -270,7 +272,7 @@ func doAuth(authReq request) (*response, error) {
 
 	loginResp, err := client.PostForm(loginURL, formValues)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed on post login url")
+		return nil, errors.Wrap(err, "failed on post login url")
 	}
 	defer loginResp.Body.Close()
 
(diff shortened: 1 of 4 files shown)


pkg/skuba/actions/auth/auth.go
```diff
@@ -186,7 +186,7 @@ func doAuth(authReq request) (*response, error) {
 	ctx := oidc.ClientContext(context.Background(), client)
 	provider, err := oidc.NewProvider(ctx, authReq.IssuerURL)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to query provider %q", authReq.IssuerURL)
+		return nil, fmt.Errorf("URL %s is not a OIDC dex server (maybe missing --root-ca or --insecure, or incorrect port number?)", authReq.IssuerURL)
```
c3y1huang:
```suggestion
		return nil, fmt.Errorf("failed to query provider %s (is this the right URL? maybe missing --root-ca or --insecure, or incorrect port number?)", authReq.IssuerURL)
```

c3y1huang: Failure doesn't necessary always the case of `URL %s is not a OIDC dex server` and seems inconsistent with the followed suggest fix -> `maybe missing --root-ca or --insecure, or incorrect port number?` 

IMHO the original message is clear without the wrapped up errors.

pkg/skuba/actions/auth/auth.go
```diff
@@ -186,7 +186,7 @@ func doAuth(authReq request) (*response, error) {
 	ctx := oidc.ClientContext(context.Background(), client)
 	provider, err := oidc.NewProvider(ctx, authReq.IssuerURL)
 	if err != nil {
```
Itxaka: Is this everything we can do? Like guess the issue? Dont we have any kind of info coming from the error so we can try to better guess the actual issue? No connection VS insecure its a very different issue so I would expect the error to point to what is wrong*

* knows nothing of go errors or the oidc lib :)

ereslibre: I think we can always improve later and evolve this kind of errors, but for the ratio effort/result I think this is good to have and then see if there's need to improve further.

pkg/skuba/actions/auth/login_test.go
```diff
@@ -59,6 +59,7 @@ func Test_Login(t *testing.T) {
 		cfg                LoginConfig
 		expectedKubeConfCb func(string, string) *clientcmdapi.Config
 		expectedError      bool
+		expectedErrorMsg   string
```
innobead:
```suggestion
		expectedErrorMsg    string
		expectedSubErrorMsg string
```

pkg/skuba/actions/auth/login_test.go
```diff
@@ -325,8 +328,13 @@ func Test_Login(t *testing.T) {
 				if err == nil {
 					t.Errorf("error expected on %s, but no error reported", tt.name)
 				}
+				if err.Error() != tt.expectedErrorMsg {
+					t.Errorf("got error msg %s, want %s", err.Error(), tt.expectedErrorMsg)
+				}
```
innobead: Suggest to change to below.

```
				if tt.expectedErrorMsg != "" {
					if !strings.Contains(err.Error(), tt.expectedErrorMsg) {
						t.Errorf("got error msg %s, want %s", err.Error(), tt.expectedErrorMsg)
					}
				}
```

pkg/skuba/actions/auth/auth.go
```diff
@@ -186,14 +186,14 @@ func doAuth(authReq request) (*response, error) {
...
-		return nil, errors.Wrapf(err, "failed to query provider %q", authReq.IssuerURL)
+		return nil, fmt.Errorf("failed to query provider %s (is this the right URL? maybe missing --root-ca or --insecure, or incorrect port number?)", authReq.IssuerURL)
```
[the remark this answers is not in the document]
jenting: Bumps out the root cause error message would cause the output message ugly like the bug description.

pkg/skuba/actions/auth/login_test.go
jenting: There is already a check and return.

https://github.com/SUSE/skuba/blob/5e73c11862bb4a7cc24ef3e843f798eeb16dacb8/pkg/skuba/actions/auth/login_test.go#L327-L335

innobead: This is due to above requested change. Given that, the transitive error has no need to specify.

```
errors.Wrapf(err, "failed to query provider %s (is this the right URL? maybe missing --root-ca or --insecure, or incorrect port number?)", authReq.IssuerURL)
```

pkg/skuba/actions/auth/auth.go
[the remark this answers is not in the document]
innobead: Does it deserve missing the root cause here for pretty message? or at least need to dump by log if you don't want to embed the original error.

pkg/skuba/actions/auth/auth.go
Itxaka: Dont get me wrong, presenting solutions to an user instead of plainly telling the error is always good, but Im wondering if it would be better to spend a bit more time and fine tune this rather than blindly guessing what could be wrong.

innobead:
This is related to https://github.com/SUSE/avant-garde/issues/1092

pkg/skuba/actions/auth/auth.go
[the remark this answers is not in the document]
jenting: Add `klog.Errorf`.

pkg/skuba/actions/auth/auth.go
jenting: Yes, but this would to the same logic as `oidc.NewProvider` does.

I could investigate is it possible to check the certificate error.

Itxaka: having a look at this on the net/http pkg, seems that we would need to regex match the error message as both would come back as `Error`.

Pretty poor error handling IMO, doesnt really merits to spend anymore time into a flaky solution :/

jenting: go 1.13 have an enhance on the error handling.

PR #876, opened 2019-12-04
Check preconditions on worker nodes before upgrading

ereslibre:
## Why is this PR needed?

Fixes https://github.com/SUSE/avant-garde/issues/1127
Fixes https://github.com/SUSE/avant-garde/issues/1128

## What does this PR do?

Worker nodes have preconditions: they cannot be upgraded if all
control planes are not in the same version (the cluster version).

This patch adds help in case any of the preconditions for any node is
not matching at `skuba node upgrade plan` and `skuba node upgrade
apply`.

Different cases have different preconditions, and this makes it clear
to the user what precondition is not met and how they can fix the
problem.

- When upgrading the platform in an HA environment, it's possible to `skuba node upgrade plan/apply` a worker node when not all control plane nodes have been updated yet.

- When upgrading the platform in an HA environment, `skuba node upgrade plan/apply` will show a clear message if any of the preconditions for any node is not met, so the upgrade cannot be started.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/upgrade/node/versions.go
+++ b/internal/pkg/skuba/upgrade/node/versions.go
@@ -18,6 +18,7 @@
 package node
 
 import (
+	"bytes"
 	"fmt"
 	"reflect"
 
@@ -29,6 +30,7 @@ import (
 
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubeadm"
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
+	"github.com/SUSE/skuba/internal/pkg/skuba/upgrade/addon"
 	upgradecluster "github.com/SUSE/skuba/internal/pkg/skuba/upgrade/cluster"
 )
 
@@ -86,6 +88,64 @@ func (nviu NodeVersionInfoUpdate) IsFirstControlPlaneNodeToBeUpgraded(client cli
 		currentClusterVersion.Patch() <= nviu.Current.KubeletVersion.Patch()
 
 	return isControlPlane && allControlPlanesMatchVersion && matchesClusterVersion, nil
+}
+
+// NodeUpgradeableCheck returns whether a given node is upgradeable or not, taking global cluster restrictions into account.
+// If all preconditions are met for the given node, no error will be returned.
+func (nviu NodeVersionInfoUpdate) NodeUpgradeableCheck(client clientset.Interface, currentClusterVersion *version.Version) error {
+	errorMessages := []string{}
+	isFirstControlPlaneNodeToBeUpgraded, err := nviu.IsFirstControlPlaneNodeToBeUpgraded(client)
+	if err != nil {
+		return err
+	}
+	if isFirstControlPlaneNodeToBeUpgraded {
+		// First check if all schedulable workers will tolerate the version we are upgrading to. If they don't, they need to be upgraded first.
+		upgradeable, err := kubernetes.AllWorkerNodesTolerateVersion(client, nviu.Update.APIServerVersion)
+		if err != nil {
+			return err
+		}
+		if !upgradeable {
+			errorMessages = append(errorMessages, fmt.Sprintf("Make sure all schedulable worker nodes match the current cluster version: %s, and retry after upgrading them", currentClusterVersion))
+		}
+		// Then check if we have addon upgrades available that would need to be applied first.
+		updatedAddons, err := addon.UpdatedAddons(client, currentClusterVersion)
+		if err != nil {
+			return err
+		}
+		if addon.HasAddonUpdate(updatedAddons) {
+			errorMessages = append(errorMessages, fmt.Sprintf("There are addon upgrades available for the current cluster version (%s) that need to be applied first", currentClusterVersion))
+		}
+	} else {
+		// There is already at least one updated control plane node
+		if nviu.Current.IsControlPlane() {
+			// Secondary control plane, check if all schedulable worker nodes tolerate the current cluster version
+			upgradeable, err := kubernetes.AllWorkerNodesTolerateVersion(client, currentClusterVersion)
+			if err != nil {
+				return err
+			}
+			if !upgradeable {
+				errorMessages = append(errorMessages, fmt.Sprintf("Make sure all schedulable worker nodes match the current cluster version: %s, and retry after upgrading them", currentClusterVersion))
+			}
+		} else {
+			// Worker node, check if all control plane nodes match the current cluster version
+			upgradeable, err := kubernetes.AllControlPlanesMatchVersion(client, currentClusterVersion)
+			if err != nil {
+				return err
+			}
+			if !upgradeable {
+				errorMessages = append(errorMessages, fmt.Sprintf("Make sure all control plane nodes match the current cluster version: %s, and retry after upgrading them", currentClusterVersion))
+			}
+		}
+	}
+	if len(errorMessages) > 0 {
+		var errorMessage bytes.Buffer
+		errorMessage.WriteString(fmt.Sprintf("node %s cannot be upgraded yet. The following errors were detected:\n", nviu.Current.Nodename))
+		for _, error := range errorMessages {
+			errorMessage.WriteString(fmt.Sprintf(" - %s\n", error))
+		}
+		return errors.New(errorMessage.String())
+	}
+	return nil
 }
 
 func UpdateStatus(client clientset.Interface, nodeName string) (NodeVersionInfoUpdate, error) {
--- a/pkg/skuba/actions/node/upgrade/apply.go
+++ b/pkg/skuba/actions/node/upgrade/apply.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
@@ -33,6 +32,7 @@ import (
 	"github.com/SUSE/skuba/internal/pkg/skuba/kured"
 	"github.com/SUSE/skuba/internal/pkg/skuba/node"
 	upgradenode "github.com/SUSE/skuba/internal/pkg/skuba/upgrade/node"
+	"github.com/pkg/errors"
 )
 
 func Apply(client clientset.Interface, target *deployments.Target) error {
@@ -46,18 +46,25 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 	}
 	currentVersion := currentClusterVersion.String()
 	latestVersion := kubernetes.LatestVersion().String()
-	fmt.Printf("Current Kubernetes cluster version: %s\n", currentVersion)
-	fmt.Printf("Latest Kubernetes version: %s\n", latestVersion)
-	fmt.Println()
-
 	nodeVersionInfoUpdate, err := upgradenode.UpdateStatus(client, target.Nodename)
 	if err != nil {
 		return err
 	}
 
+	fmt.Printf("Current Kubernetes cluster version: %s\n", currentVersion)
+	fmt.Printf("Latest Kubernetes version: %s\n", latestVersion)
+	fmt.Printf("Current Node version: %s\n", nodeVersionInfoUpdate.Current.KubeletVersion.String())
+	fmt.Println()
+
 	if nodeVersionInfoUpdate.IsUpdated() {
 		fmt.Printf("Node %s is up to date\n", target.Nodename)
 		return nil
+	}
+
+	// Check if the node is upgradeable (matches preconditions)
+	if err := nodeVersionInfoUpdate.NodeUpgradeableCheck(client, currentClusterVersion); err != nil {
+		fmt.Println()
+		return err
 	}
 
 	// Check if skuba-update.timer is already disabled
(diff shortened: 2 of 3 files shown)


internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -88,6 +89,51 @@ func (nviu NodeVersionInfoUpdate) IsFirstControlPlaneNodeToBeUpgraded(client cli
 	return isControlPlane && allControlPlanesMatchVersion && matchesClusterVersion, nil
 }
 
+// IsNodeUpgradeable returns whether a given node is upgradeable or not, taking global cluster restrictions into account.
+// If all preconditions are met for the given node, no error will be returned.
+func (nviu NodeVersionInfoUpdate) IsNodeUpgradeable(client clientset.Interface, currentClusterVersion *version.Version) error {
```
MaximilianMeister: from the naming of the function it implies that a bool is returned, we have that covention in some other places as well. As it's easier to just return `err` here and not a bool and an err, maybe I'd rename the function to something like `NodeUpgradeableCheck`

internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -88,6 +89,51 @@ func (nviu NodeVersionInfoUpdate) IsFirstControlPlaneNodeToBeUpgraded(client cli
...
+	var upgradeable bool
+	isFirstControlPlaneNodeToBeUpgraded, err := nviu.IsFirstControlPlaneNodeToBeUpgraded(client)
+	if err != nil {
+		return err
+	}
+	if isFirstControlPlaneNodeToBeUpgraded {
+		upgradeable, err = kubernetes.AllWorkerNodesTolerateVersion(client, nviu.Update.APIServerVersion)
+		if err != nil {
```
mssola: Here and the two below: couldn't this be reconstructed as:

```golang
if isFirstControlPlaneNodeToBeUpgraded {
    upgradeable, err := myFunction(...)
    if upgradeable || err != nil {
        return err
    }
    error.WriteString(...)
} else {
    // The other cases
}

return error
```

Just saying because it looks like if it can be upgraded, `err` would be nil anyways, no? (and if isn't it wouldn't matter much, then we'd need to add an extra if inside of the current one). All in all, it would allow us to get rid off redundant if-else statements at the end of this function.

ereslibre: Not sure what you mean, can you please elaborate?

ereslibre:
There's a small detail missing. Will update the PR tomorrow.

jenting:
2 master + 2 workers without load-balancer, deploy version `1.15.2` and upgrade first control plane node, I encountered this error :thinking: 
```
I1205 15:15:40.107106   31598 states.go:40] === state kubernetes.install-node-pattern applied successfully ===
I1205 15:15:40.107124   31598 states.go:35] === applying state kubeadm.upgrade.apply ===
I1205 15:15:40.107145   31598 files.go:31] uploading to remote file "/tmp/kubeadm-upgrade.conf" with contents
I1205 15:15:41.949403   31598 ssh.go:168] running command: "sudo sh -c 'kubeadm upgrade apply --config /tmp/kubeadm-upgrade.conf -y'"
E1205 15:15:42.334532   31598 ssh.go:193] no InitConfiguration or ClusterConfiguration kind was found in the YAML file
E1205 15:15:42.334567   31598 ssh.go:193] To see the stack trace of this error execute with --v=5 or higher
I1205 15:15:42.535423   31598 ssh.go:168] running command: "sudo sh -c 'rm /tmp/kubeadm-upgrade.conf'"
Unable to apply node upgrade: failed to apply state kubeadm.upgrade.apply: Process exited with status 1
```

ereslibre:
> 2 master + 2 workers without load-balancer, deploy version 1.15.2 and upgrade first control plane node, I encountered this error 

Yes, the patch was not complete :grin:

ereslibre:
Checked al possible cases I could come up with and this PR is working as intended now. Please, review :)

internal/pkg/skuba/upgrade/node/versions.go
suseclee: Or
```
if ! upgradeable || err != nil {
      errorMessages = append(errorMessages, fmt.Sprintf("There are addon upgrades available for the current cluster version (%s) that need to be applied first", currentClusterVersion))
    }

```

internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -88,6 +90,64 @@ func (nviu NodeVersionInfoUpdate) IsFirstControlPlaneNodeToBeUpgraded(client cli
...
+		}
+		// Then check if we have addon upgrades available that would need to be applied first.
+		updatedAddons, err := addon.UpdatedAddons(client, currentClusterVersion)
+		if err != nil {
+			return err
+		}
+		if addon.HasAddonUpdate(updatedAddons) {
+			errorMessages = append(errorMessages, fmt.Sprintf("There are addon upgrades available for the current cluster version (%s) that need to be applied first", currentClusterVersion))
+		}
+	} else {
```
suseclee: Or,
```
if addon.HasAddonUpdate(updatedAddons) || err != nil  {
   errorMessages = append(errorMessages, fmt.Sprintf("There are addon upgrades available for the current cluster version (%s) that need to be applied first", currentClusterVersion))
```

internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -88,6 +90,64 @@ func (nviu NodeVersionInfoUpdate) IsFirstControlPlaneNodeToBeUpgraded(client cli
...
+		// There is already at least one updated control plane node
+		if nviu.Current.IsControlPlane() {
+			// Secondary control plane, check if all schedulable worker nodes tolerate the current cluster version
+			upgradeable, err := kubernetes.AllWorkerNodesTolerateVersion(client, currentClusterVersion)
+			if err != nil {
+				return err
+			}
+			if !upgradeable {
```
suseclee: Or,
```
if !upgradeable || err != nil {
    errorMessages = append(errorMessages, fmt.Sprintf("Make sure all schedulable worker nodes match the current cluster version: %s, and retry after upgrading them", currentClusterVersion))
			}
```

internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -88,6 +90,64 @@ func (nviu NodeVersionInfoUpdate) IsFirstControlPlaneNodeToBeUpgraded(client cli
...
+			}
+		} else {
+			// Worker node, check if all control plane nodes match the current cluster version
+			upgradeable, err := kubernetes.AllControlPlanesMatchVersion(client, currentClusterVersion)
+			if err != nil {
+				return err
+			}
+			if !upgradeable {
+				errorMessages = append(errorMessages, fmt.Sprintf("Make sure all control plane nodes match the current cluster version: %s, and retry after upgrading them", currentClusterVersion))
+			}
```
suseclee: Or, 
```
if !upgradeable || err != nil {
   errorMessages = append(errorMessages, fmt.Sprintf("Make sure all control plane nodes match the current cluster version: %s, and retry after upgrading them", currentClusterVersion))
			}
```

pkg/skuba/actions/node/upgrade/plan.go
```diff
@@ -54,6 +53,12 @@ func Plan(client clientset.Interface, nodeName string) error {
 		}
 		fmt.Printf("  - kubelet: %s -> %s\n", nodeVersionInfoUpdate.Current.KubeletVersion.String(), nodeVersionInfoUpdate.Update.KubeletVersion.String())
 		fmt.Printf("  - cri-o: %s -> %s\n", nodeVersionInfoUpdate.Current.ContainerRuntimeVersion.String(), nodeVersionInfoUpdate.Update.ContainerRuntimeVersion.String())
+
+		// Check if the node is upgradeable (matches preconditions)
+		if err := nodeVersionInfoUpdate.NodeUpgradeableCheck(client, currentClusterVersion); err != nil {
+			fmt.Println()
+			return err
+		}
```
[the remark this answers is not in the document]
ereslibre: We have many non-optimized code when comparing versions (like retrieving every component is a separate request), we could improve this and this process would be faster.

I think it's good that plan prints as many known information as possible, so they get the whole idea of what's going to happen, that's why I added this at the end.

ereslibre:
@suseclee 

> IMO, customers want to all in one for upgrade too like skuba cluster upgrade apply for easy upgrade instead

There's no such command right now, this would need to somehow solve how to pass the IP addresses, user and sudo for each machine at once. To the best of my knowledge it's good if they can script these commands away, but adding a `skuba cluster upgrade apply` would require an RFC most probably and this PR is trying to help with the upgrade procedure for this immediate release.

internal/pkg/skuba/upgrade/node/versions.go
ereslibre: Is not strictly the same, `err` here could be an unknown error (e.g. could not connect to the apiserver).

internal/pkg/skuba/upgrade/node/versions.go
ereslibre: Is not strictly the same, `err` here could be an unknown error (e.g. could not connect to the apiserver).

internal/pkg/skuba/upgrade/node/versions.go
ereslibre: Is not strictly the same, `err` here could be an unknown error (e.g. could not connect to the apiserver).

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -60,6 +60,11 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 		return nil
 	}
 
+	// Check if the node is upgradeable (matches preconditions)
+	if err := nodeVersionInfoUpdate.NodeUpgradeableCheck(client, currentClusterVersion); err != nil {
+		return err
```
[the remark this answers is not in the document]
ereslibre: Done.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -60,6 +60,11 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 		return nil
 	}
 
+	// Check if the node is upgradeable (matches preconditions)
```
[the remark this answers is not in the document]
ereslibre: Done

ereslibre:
@jenting 

> The performance is not well, some functions would run twice when skuba node upgrade [plan|apply], lie kubeadm.GetCurrentClusterVersion .

I agree. We can improve performance on a follow up PR also for `UpdateStatus` that is performing many requests to the apiserver when it doesn't need that many.

kravciak:
All masters have to upgraded before workers?

Commit title: `Check preconditions on worker nodes before upgrading` - 53 chars, you are allowed 50 :)

ereslibre:
> Commit title: Check preconditions on worker nodes before upgrading - 53 chars, you are allowed 50 :)

I just cannot make any shorter while keeping it meaningful :), that check is not required so we can merge without that strict test passing :)

thehejik:
It works as described, now I'm unable to upgrade worker node before masters:
```
thehejik@dhcp225:~/git/scripts/cluster_tRp/cluster> skuba node upgrade apply -s -u sles -t 10.86.1.78 #(worker - masster not updated yet)
 This is an UNTAGGED version and NOT intended for production usage. 
Unable to apply node upgrade: caasp-worker-th-skuba-update-0 is not upgradeable until all control plane nodes are upgraded
```

But there is still one issue, when I upgrade first master and then I ask again for `skuba cluster upgrade plan` it says:
```
skuba cluster upgrade plan
Current Kubernetes cluster version: 1.16.2
Latest Kubernetes version: 1.16.2

Congratulations! You are already at the latest version available
```
But at the time ^ only first master was upgraded. @ereslibre do you think it deserves separate issue?

ereslibre:
@thehejik are you sure that's the output of the code with this PR applied? It should be slightly different. Something along the lines:

```
node %s cannot be upgraded yet. The following errors were detected:
  - Make sure all control plane nodes match the current cluster version: %s, and retry after upgrading them
```

(same error message on `plan` and `apply`).

Regarding your question: whenever you upgrade the first control plane node of the cluster, the "current cluster version" is bumped (this is, the Kubernetes version stored in the `kubeadm-config` ConfigMap). So this is expected. Please, file another issue if you believe `skuba cluster upgrade plan` should also take into consideration the real status of the nodes compared to the current cluster version. This would mean improving its output.

suseclee:
> @thehejik are you sure that's the output of the code with this PR applied? It should be slightly different. Something along the lines:
> 
> ```
> node %s cannot be upgraded yet. The following errors were detected:
>   - Make sure all control plane nodes match the current cluster version: %s, and retry after upgrading them
> ```
> 
> (same error message on `plan` and `apply`).
> 
> Regarding your question: whenever you upgrade the first control plane node of the cluster, the "current cluster version" is bumped (this is, the Kubernetes version stored in the `kubeadm-config` ConfigMap). So this is expected. Please, file another issue if you believe `skuba cluster upgrade plan` should also take into consideration the real status of the nodes compared to the current cluster version. This would mean improving its output.

I think this is a very good idea especially customers with very large number of nodes in a cluster. They want to check upgrade status.

thehejik:
> @thehejik are you sure that's the output of the code with this PR applied? It should be slightly different.

You are right, the output is different (I did it with older version), anyway it works as described with skuba from master, now the output is:
```
cluster> skuba node upgrade apply -s -u sles -t 10.86.1.78
Current Kubernetes cluster version: 1.16.2
Latest Kubernetes version: 1.16.2
Current Node version: 1.15.2


Unable to apply node upgrade: node caasp-worker-th-skuba-update-0 cannot be upgraded yet. The following errors were detected:
 - Make sure all control plane nodes match the current cluster version: 1.16.2, and retry after upgrading them
```

thehejik:
> I think this is a very good idea especially customers with very large number of nodes in a cluster. They might want to check the upgrade status for nodes when they are lost (who knows).

Issue opened https://github.com/SUSE/avant-garde/issues/1144

PR #878, opened 2019-12-04
Deprecate extensions apiGroup for PSP

drpaneas:
Replace the apiGroup 'extensions' with 'policy' as the
last part to make SUSE provided PSPs with Kubernetes 1.16.

## Why is this PR needed?

Fixes https://github.com/SUSE/avant-garde/issues/1122

## What does this PR do?

The PodSecurityPolicy in Kubernetes 1.15 used to work
in the `extentions/v1beta1` API group. In the new
version, this API group does not exist any more.
It has been deprecated and removed. The replacement is `policy/v1beta1`.

This PR replaces the apiGroup 'extensions' with 'policy' as the
last part to make SUSE provided PSPs with Kubernetes 1.16.

This PR makes sure that new kubernetes clusters bootstraped with skuba and 1.16 are working fine with PSP.

But, this PR does not check if this change will be affected during the maintenance update. The new configuration has to be applied during the maintenance update.

## Info for QA

All the info you need 

### Related info

Read my comment at: https://github.com/SUSE/avant-garde/issues/1122#issuecomment-561757161

```
$ kubectl auth can-i use psp/suse.caasp.psp.unprivileged --as=any-user
Warning: resource 'podsecuritypolicies' is not namespace scoped in group 'policy'
no
```

```
$ kubectl auth can-i use psp/suse.caasp.psp.unprivileged --as=any-user
Warning: resource 'podsecuritypolicies' is not namespace scoped in group 'policy'
yes
```

## Docs

If there is description talking about PSP and there is a copy-paste of the the definition of it, make sure it reflects the changes to the new API. 

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/upgrade/addon/versions_test.go
+++ b/internal/pkg/skuba/upgrade/addon/versions_test.go
@@ -83,7 +83,7 @@ AddonsVersion:
 					kubernetes.PSP:     nil,
 				},
 				Updated: kubernetes.AddonsVersion{
-					kubernetes.PSP: &kubernetes.AddonVersion{Version: "", ManifestVersion: 0},
+					kubernetes.PSP: &kubernetes.AddonVersion{Version: "", ManifestVersion: 1},
 				},
 			},
 			expectedErr: false,
@@ -124,7 +124,7 @@ func TestHasAddonUpdate(t *testing.T) {
 					kubernetes.Kured:   &kubernetes.AddonVersion{Version: "1.2.0", ManifestVersion: 0},
 					kubernetes.Dex:     &kubernetes.AddonVersion{Version: "2.16.0", ManifestVersion: 0},
 					kubernetes.Gangway: &kubernetes.AddonVersion{Version: "3.1.0", ManifestVersion: 0},
-					kubernetes.PSP:     &kubernetes.AddonVersion{Version: "1.0.0", ManifestVersion: 0},
+					kubernetes.PSP:     &kubernetes.AddonVersion{Version: "1.0.0", ManifestVersion: 1},
 				},
 				Updated: kubernetes.AddonsVersion{
 					kubernetes.Cilium: &kubernetes.AddonVersion{Version: "1.5.3", ManifestVersion: 1},
@@ -140,7 +140,7 @@ func TestHasAddonUpdate(t *testing.T) {
 					kubernetes.Kured:   &kubernetes.AddonVersion{Version: "1.2.0", ManifestVersion: 0},
 					kubernetes.Dex:     &kubernetes.AddonVersion{Version: "2.16.0", ManifestVersion: 0},
 					kubernetes.Gangway: &kubernetes.AddonVersion{Version: "3.1.0", ManifestVersion: 0},
-					kubernetes.PSP:     &kubernetes.AddonVersion{Version: "1.0.0", ManifestVersion: 0},
+					kubernetes.PSP:     &kubernetes.AddonVersion{Version: "1.0.0", ManifestVersion: 1},
 				},
 			},
 			expected: false,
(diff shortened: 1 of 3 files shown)


jordimassaguerpla:
@ereslibre I see in the description

> But, this PR does not check if this change will be affected during the maintenance update. The new configuration has to be applied during the maintenance update.

Do you know how can we fix this?

drpaneas:
> Thank you @drpaneas.
> 
> Can you bump the PSP manifest version on the `versions.go` file as well? Please bump both, [1.15](https://github.com/SUSE/skuba/blob/56bb93badff741003d3ee14427c9ca60a4e02af0/internal/pkg/skuba/kubernetes/versions.go#L118) and [1.16](https://github.com/SUSE/skuba/blob/56bb93badff741003d3ee14427c9ca60a4e02af0/internal/pkg/skuba/kubernetes/versions.go#L98).

I'm afraid I don't follow. What is expected to bump there? I saw similar _bumps_ like `cilium` and I saw there's an `annotation` called `addon.caasp.suse.com/manifest-version: "1.5.3-1`. Do you mean something like that?

ereslibre:
> I'm afraid I don't follow. What is expected to bump there? I saw similar _bumps_ like `cilium` and I saw there's an `annotation` called `addon.caasp.suse.com/manifest-version: "1.5.3-1`. Do you mean something like that?

The annotation will be created by skuba automatically, no need to worry about that. What needs to be done along with this change is to bump the manifest version on the two lines I linked on my comment, bumping the number from `1` to `2`, so `skuba addon upgrade plan/apply` knows that it has to reapply PSP's.

By doing so on both versions, and in conjuction with https://github.com/SUSE/skuba/pull/876, `skuba` will make sure that you `skuba addon upgrade apply` before you upgrade your first control plane, applying the new PSP (that I get should be compatible with 1.15, is it right?).

internal/pkg/skuba/upgrade/addon/versions_test.go
```diff
@@ -124,7 +124,7 @@ func TestHasAddonUpdate(t *testing.T) {
 					kubernetes.Kured:   &kubernetes.AddonVersion{Version: "1.2.0", ManifestVersion: 0},
 					kubernetes.Dex:     &kubernetes.AddonVersion{Version: "2.16.0", ManifestVersion: 0},
 					kubernetes.Gangway: &kubernetes.AddonVersion{Version: "3.1.0", ManifestVersion: 0},
-					kubernetes.PSP:     &kubernetes.AddonVersion{Version: "1.0.0", ManifestVersion: 0},
+					kubernetes.PSP:     &kubernetes.AddonVersion{Version: "1.0.0", ManifestVersion: 1},
```
Itxaka: somehow on my local testings, changing this and the one below still makes the tests pass, so this changes may not be necessary or we are not checking this or the tests are wrong. Only the change in line 86 seems to be needed.

ereslibre: Ideally these tests shouldn't be tied to the real versions in the `versions.go` file... If they are somehow interdependant we should make them independent.

jenting: Sorry I made this before :( It would be great if thais PR makes the tests independent to version.go file.

ereslibre: I think this patch is pretty important for getting all required pieces for 1.16, we can improve the tests later on a follow-up PR if that's fine.

jenting: PR for addressing this https://github.com/SUSE/skuba/pull/882 issue.

PR #883, opened 2019-12-06
Add option to configure datastore_cluster 

dirkmueller:
Using 3PAR over network ISCSI is not a great idea regarding IOPS
performance which matters for etcd.  Add ability to set a datastore_cluster
as well (optionally)

## Why is this PR needed?

It fixes the e2e CI failures on vmware because they are currently using
ISCSI (which is very slow) on 3PAR (which is extremely slow). 

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

## Docs

Documentation update is in https://github.com/SUSE/doc-caasp/pull/629

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/vmware/terraform.tfvars.json.ci.example
+++ b/ci/infra/vmware/terraform.tfvars.json.ci.example
@@ -1,6 +1,7 @@
 {
     "vsphere_datastore": "node51-2",
     "vsphere_datacenter": "NUREMBERG",
+    "vsphere_datastore_cluster": null,
     "vsphere_network": "VM Network",
     "vsphere_resource_pool": "CaaSP_CI",
     "template_name": "SLES15-SP1-GM-up191203-guestinfo_node51",
--- a/ci/infra/vmware/worker-instance.tf
+++ b/ci/infra/vmware/worker-instance.tf
@@ -66,7 +66,8 @@ resource "vsphere_virtual_machine" "worker" {
   firmware         = var.firmware
   scsi_type        = data.vsphere_virtual_machine.template.scsi_type
   resource_pool_id = data.vsphere_resource_pool.pool.id
-  datastore_id     = data.vsphere_datastore.datastore.id
+  datastore_id     = (var.vsphere_datastore == null ? null: data.vsphere_datastore.datastore[0].id)
+  datastore_cluster_id = (var.vsphere_datastore_cluster == null ? null : data.vsphere_datastore_cluster.datastore[0].id)
 
   clone {
     template_uuid = data.vsphere_virtual_machine.template.id
--- a/ci/infra/vmware/README.md
+++ b/ci/infra/vmware/README.md
@@ -54,6 +54,7 @@ Copy the `terraform.tfvars.example` to `terraform.tfvars` and provide reasonable
 
 `vsphere_datastore` - Provide the datastore to use in vSphere\
 `vsphere_datacenter` - Provide the datacenter to use in vSphere\
+`vsphere_datastore_cluster` - Provide the datastore cluster to use on the vSphere server\
 `vsphere_network` - Provide the network to use in vSphere - this network must be able to access the ntp servers and the nodes must be able to reach each other\
 `vsphere_resource_pool` - Provide the resource pool the machines will be running in\
 `template_name` - The template name the machines will be copied from\
--- a/ci/infra/vmware/lb-instance.tf
+++ b/ci/infra/vmware/lb-instance.tf
@@ -134,7 +134,8 @@ resource "vsphere_virtual_machine" "lb" {
   firmware         = var.firmware
   scsi_type        = data.vsphere_virtual_machine.template.scsi_type
   resource_pool_id = data.vsphere_resource_pool.pool.id
-  datastore_id     = data.vsphere_datastore.datastore.id
+  datastore_id     = (var.vsphere_datastore == null ? null: data.vsphere_datastore.datastore[0].id)
+  datastore_cluster_id = (var.vsphere_datastore_cluster == null ? null : data.vsphere_datastore_cluster.datastore[0].id)
 
   clone {
     template_uuid = data.vsphere_virtual_machine.template.id
--- a/ci/infra/vmware/master-instance.tf
+++ b/ci/infra/vmware/master-instance.tf
@@ -66,7 +66,8 @@ resource "vsphere_virtual_machine" "master" {
   firmware         = var.firmware
   scsi_type        = data.vsphere_virtual_machine.template.scsi_type
   resource_pool_id = data.vsphere_resource_pool.pool.id
-  datastore_id     = data.vsphere_datastore.datastore.id
+  datastore_id     = (var.vsphere_datastore == null ? null: data.vsphere_datastore.datastore[0].id)
+  datastore_cluster_id = (var.vsphere_datastore_cluster == null ? null : data.vsphere_datastore_cluster.datastore[0].id)
 
   clone {
     template_uuid = data.vsphere_virtual_machine.template.id
--- a/ci/infra/vmware/variables.tf
+++ b/ci/infra/vmware/variables.tf
@@ -5,6 +5,11 @@ variable "stack_name" {
 }
 
 variable "vsphere_datastore" {
+  default = null
+}
+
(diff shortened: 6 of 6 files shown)


dirkmueller:
ok, I see this is not so simple potentially - https://github.com/lcavajani/skuba/commit/2f4e58db951214cb7d4add1220baab4227795191

pablochacin:
The problem I see is that even when those changes are simple we still need to find a way to automate the way we introduce it without changing the terraform files.  We could simply prepare a patch and apply it before running the test.

thehejik:
> The problem I see is that even when those changes are simple we still need to find a way to automate the way we introduce it without changing the terraform files. We could simply prepare a patch and apply it before running the test.

Exactly, we are doing it on the fly within validation by using sed and echo but patch would work as well.

dirkmueller:
thansk for setting this to "needs change" without actually explaining what needs change. With VMware you have (at least) 3 options on how to allocate storage for a virtual machine:

* using VMFS, which afaik is the default
* using a clustered datastore (which is a internal datastore clustering inside vmware, similar to e.g. ceph in SUSE world)
* using a hardcoded datastore (which could be a particular local drive or a NAS drive). 

as far as I can see, limiting to a datastore seems to be the least versatile option to use for deployment, and in this case is outright harmful because it is an iSCS attached drive which is just too slow for the rest of the system's expectations. 

The right choice would be a) to make the ci default to a local disk and b) give the user options of deployment. 

Shipping terraform files that we're not using ourselves and instead letting QA sed the terraform files before validating the product seems against the principle of testing what we ship.

pablochacin:
> thansk for setting this to "needs change" without actually explaining what needs change.  

I think @thehejik was pretty clear, pointing to the actual changes that need to be made. 

> Shipping terraform files that we're not using ourselves and instead letting QA sed the terraform files before validating the product seems against the principle of testing what we ship.

Agree, but introducing this change as a temporary measure in order to have the tests running until we fix the problem with slow disks seems better than not testing anything at all, IMHO.

dirkmueller:
@pablochacin I was mentioning the patch already in previous comment. 

So I just submit the patch that was referenced and that we already decided against merging? Why?

Ar least it would make sense to make the storage backend selectable instead of preferring the least useful option. Or like this patch does, not set it at all.. which makes it chose some default.

jordimassaguerpla:
> @pablochacin I was mentioning the patch already in previous comment.
> 
> So I just submit the patch that was referenced and that we already decided against merging? Why?
> 
> Ar least it would make sense to make the storage backend selectable instead of preferring the least useful option. Or like this patch does, not set it at all.. which makes it chose some default.

IIUC our customers use our tf files as a template. Not choosing a default can make sense as long as we update the documentation to reflect this change.

If we don't use a default, will we be using the local storage in vmware? or do we need to change that?

dirkmueller:
@jordimassaguerpla yes, afaik it will use vmfs, which is defaulting to whatever has been configured by the admin (local datastores for example)

Anyway, I've updated the PR to include the ability to set the datastore_cluster in addition to the datastore, and changed the CI worker to set the datastore_cluster rather than datastore. I hope this is a sensible compromise between setting datastores (still not sure why we would want to do that but it has been the default so far) and datatore_cluster (which we need for CI).

jordimassaguerpla:
@dirkmueller I retriggered the tests and they are green now. Please see the requested changes to get this approved.Thanks.

dirkmueller:
@jordimassaguerpla  we need to do this after the TF 0.12 switch

jordimassaguerpla:
Hi!
Happy hacking

evrardjp:
Just curious, how does this relate to https://github.com/SUSE/skuba/pull/950/files ?

dirkmueller:
@evrardjp not at all, this pr is a month older. 

The original attempt was at making the terraform files usable so that you can configure either datastore or datastore_cluster (the latter I think being more relevant to customers, but YMMV). 

The other PR switches to a different cluster on better hardware which might be a quick fix to the current stability issues. It is incremental. 

Note that this PR is currently not in a mergeable state as terraform fails with that change and I'm slow with debugging this because I have to debug it via CI as I don't have credentials to test this outside CI. 

Please do not merge it until I got the job to pass or at least a way to fix the regression.

evrardjp:
Hey, if I merge I will break world, I am not looking forward to that :p 
My question was more about -- "Are you aware of this relatively new PR? Do we still need this?"

You answer tells me that datastore_cluster is more relevant for customers, which explains why it's not closed.

ci/infra/vmware/terraform.tfvars.json.ci.example
```diff
@@ -1,6 +1,7 @@
 {
     "vsphere_datastore": "3PAR",
     "vsphere_datacenter": "PROVO",
+    "vsphere_datastore_cluster": null,
```
Itxaka: I think deployment is failing because this is set to null. Should be set to `LOCAL-DISKS-CLUSTER` for our current state

ci/infra/vmware/worker-instance.tf
```diff
@@ -67,6 +67,7 @@ resource "vsphere_virtual_machine" "worker" {
   scsi_type        = data.vsphere_virtual_machine.template.scsi_type
   resource_pool_id = data.vsphere_resource_pool.pool.id
   datastore_id     = data.vsphere_datastore.datastore.id
+  datastore_cluster_id = data.vsphere_datastore_cluster.datastore.id
```
Itxaka: cant use both. If `datastore_id` and `datastore_cluster_id` is set at the same time, deployment fails:
`Error: "datastore_cluster_id": conflicts with datastore_id`

ci/infra/vmware/terraform.tfvars.json.ci.example
Itxaka: now that defautls are set to null we can override this for our CI and remove `vsphere_datastore` var from here :D

Itxaka:
FYI this is needed to use the datastore cluster in NUE as there is no shared storage or anything in there to use generic names https://github.com/SUSE/avant-garde/issues/1354

jordimassaguerpla:
Will this have any impact on our customers? Anything we should mention in the Release Notes?

Itxaka:
> Will this have any impact on our customers? Anything we should mention in the Release Notes?

Should not have an impact but we should mention that we can now choose to use a datastore_cluster **and** document that the options are exclusive, either you choose a datastore or a datastore_cluster.

jordimassaguerpla:
> > Will this have any impact on our customers? Anything we should mention in the Release Notes?
> 
> Should not have an impact but we should mention that we can now choose to use a datastore_cluster **and** document that the options are exclusive, either you choose a datastore or a datastore_cluster.

Ok. Then we will need also a PR to documentation and/or Release notes in order to merge this. It does not need to be a perfect PR, can be a WIP, but just so we not forget.

I'll trust you will do that **before** merging, so I am giving my LGTM

Itxaka:
> Ok. Then we will need also a PR to documentation and/or Release notes in order to merge this. It does not need to be a perfect PR, can be a WIP, but just so we not forget.

Rocke is borked so asking here. Do we have lables to add to PR for docs-required/release-notes-required? Should it be part of the card acceptance criteria? Both?

jordimassaguerpla:
> > Ok. Then we will need also a PR to documentation and/or Release notes in order to merge this. It does not need to be a perfect PR, can be a WIP, but just so we not forget.
> 
> Rocke is borked so asking here. Do we have lables to add to PR for docs-required/release-notes-required? Should it be part of the card acceptance criteria? Both?

you need an issue in github/SUSE/avant-garde. In the description of that issue, there is a section at the bottom for related issues/urls ... you can put there both, this PR and the PRs for the docs

dirkmueller:
The documentation has been updated already many months ago.

See the issue linked at the beginning of the PR

Itxaka:
>The documentation has been updated already many months ago.
>See the issue linked at the beginning of the PR

but that was reverted! https://github.com/SUSE/doc-caasp/pull/663

Plus my docs are cooler, they have a big IMPORTANT note in there about exclusivity of the options :p

PR #885, opened 2019-12-06
Rename testrunner logs directory

cmurphy:
## Why is this PR needed?

Without this change, the logs that testrunner collects from the platform
nodes are archived in the testrunner_logs directory, which is a
confusing place for these logs since the logs for testrunner itself are
not stored there. 

Fixes SUSE/avant-garde#714

## What does this PR do?

This change moves the default platform log directory
to platform_logs and updates the Jenkinsfiles and documentation for the
new log path.

n/a

## Info for QA

n/a

### Related info

n/a

Logs from CI are stored in testrunner_logs directory in the workspace.

Logs from CI are stored in the platform_logs directory in the workspace.

## Docs

n/a 

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/utils/constants.py
+++ b/ci/infra/testrunner/utils/constants.py
@@ -151,7 +151,7 @@ class BaseConfig:
         conf.workspace = os.path.expanduser(conf.workspace)
 
         if not conf.log_dir:
-            conf.log_dir = os.path.join(conf.workspace, 'testrunner_logs')
+            conf.log_dir = os.path.join(conf.workspace, 'platform_logs')
         elif not os.path.isabs(conf.log_dir):
             conf.log_dir = os.path.join(conf.workspace, conf.log_dir)
 
--- a/ci/infra/testrunner/README.md
+++ b/ci/infra/testrunner/README.md
@@ -110,7 +110,7 @@ username: "username"
 
 There are some arguments that are currently at the top level of the configuration, but are actually related to the platform:
 
-- log_dir: path to the directory where platform logs are collected. Defaults to `<workspace>/testrunner_logs` 
+- log_dir: path to the directory where platform logs are collected. Defaults to `<workspace>/platform_logs`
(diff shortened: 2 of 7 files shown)


cmurphy:
Logs are in the right place: https://ci.suse.de/blue/organizations/jenkins/caasp-jobs%2Fcaasp-v4-test/detail/PR-885/1/artifacts

ci/infra/testrunner/utils/constants.py
```diff
@@ -151,7 +151,7 @@ def finalize(conf):
         conf.workspace = os.path.expanduser(conf.workspace)
 
         if not conf.log_dir:
-            conf.log_dir = os.path.join(conf.workspace, 'testrunner_logs')
+            conf.log_dir = os.path.join(conf.workspace, 'platform_logs')
```
dirkmueller: I admit I don't fully understand the background of the story that led to this change but this feels slightly odd to change the path of the logdir of a python module stored inside 'testrunner' directory to default to 'platform'. 

should the whole infra/testrunner directory be renamed to platform as well?

dannysauer: @pablochacin touched on this in the linked issue (SUSE/avant-garde#714); the intended purpose is to differentiate the output of test execution from the output of testrunner itself.

cmurphy: > should the whole infra/testrunner directory be renamed to platform as well?

I don't think so. My understanding is that the platform logs here are like the service logs we would have collected with supportconfig in mkcloud, so calling these logs `testrunner_logs` is like calling the supportconfig tarballs `tempest_logs`, it doesn't make sense because the test runner has its own logs for the tests themselves (`tempest.log` and `testrunner.log`).

dirkmueller: Thanks!

PR #886, opened 2019-12-07
Fix unable to bootstrap by using FQDN

jenting:
##  Why is this PR needed?

Fix Unable to bootstrap first master node by using its FQDN hostname

Fixes https://github.com/SUSE/avant-garde/issues/1133

appear in the changelog.

## What does this PR do?

Previously, the user input target node IP or FQDN inserted into altNames.IPs. When the target is FQDN, this causes generates kubelet server certificate failed.

Add all host IP or FQDN into hosts slice, then check each entity in hosts slice is IP or FQDN, adding to corresponding altNames.

N/A

## Info for QA

### Related info

N/A

Bootstrap/Join the target node by using FQDN, deployed failed.

1. Bootstrap/Join the target node by using FQDN, deployed success.
2. Install metrics server success. Reference to https://github.com/SUSE/skuba/pull/832#issue-342218019
2. `kubectl top [pod|node]` success

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/kubelet.go
+++ b/internal/pkg/skuba/deployments/ssh/kubelet.go
@@ -85,8 +85,13 @@ func kubeletCreateAndUploadServerCert(t *Target, data interface{}) error {
 		}
 	}
 
-	alternateIPs := []net.IP{net.ParseIP(t.target.Target), net.IPv4(127, 0, 0, 1), net.IPv6loopback}
+	alternateIPs := []net.IP{net.IPv4(127, 0, 0, 1), net.IPv6loopback}
 	alternateDNS := []string{"localhost"}
+	if ip := net.ParseIP(t.target.Target); ip != nil {
+		alternateIPs = append(alternateIPs, ip)
+	} else {
+		alternateDNS = append(alternateDNS, t.target.Target)
+	}
 
 	altNames.IPs = append(altNames.IPs, alternateIPs...)
 	altNames.DNSNames = append(altNames.DNSNames, alternateDNS...)


internal/pkg/skuba/deployments/ssh/kubelet.go
```diff
@@ -66,33 +67,30 @@ func kubeletCreateAndUploadServerCert(t *Target, data interface{}) error {
...
+
+	altNames := certutil.AltNames{}
+	for _, host := range hosts {
+		if host == "" {
+			continue
+		}
+
+		if ip := net.ParseIP(host); ip != nil {
 			altNames.IPs = append(altNames.IPs, ip)
+		} else {
```
aojea: nit: Can we add a TODO to verify that it is a valid hostname?

jenting: Like this right? https://github.com/kubernetes/kubernetes/blob/56f60d1bcc81270fd9239ddb6b8ed83b1beb77a6/cmd/kubeadm/app/util/pkiutil/pki_helpers.go#L447-L451

jenting: Added
https://github.com/SUSE/skuba/blob/2eaf92b2c9139b0947e3fff23a12adfbead5e6bd/internal/pkg/skuba/util/cert.go#L52-L60

jenting:
> ```
> 	commonName := t.target.Nodename
> 	hosts := []string{t.target.Nodename, t.target.Target, "127.0.0.1", "localhost"}
> 	stdout, _, err := t.silentSsh("hostname", "-I")
> 	if err != nil {
> 		return err
> 	}
> 	hosts = append(hosts, strings.Split(stdout, " ")...)
> ```
This a bit hard now since ssh have not refactor to pass interface for unit test to mock remote host.
The rest test cases for SAN should be covered here. Correct me if I missed some test cases.
https://github.com/SUSE/skuba/blob/2eaf92b2c9139b0947e3fff23a12adfbead5e6bd/internal/pkg/skuba/util/cert_test.go#L31-L157

> ```
> 	altNames := certutil.AltNames{}
> 	for _, host := range hosts {
> 		if host == "" {
> 			continue
> 		}
> 
> 		if ip := net.ParseIP(host); ip != nil {
> 			altNames.IPs = append(altNames.IPs, ip)
> 		} else {
> 			altNames.DNSNames = append(altNames.DNSNames, host)
> 		}
> 	}
> ```

Create server cert with function `NewServerCertAndKey`.
https://github.com/SUSE/skuba/blob/2eaf92b2c9139b0947e3fff23a12adfbead5e6bd/internal/pkg/skuba/util/cert.go#L40-L79

dannysauer:
This is a *lot* of code to fix a single list element error.  Does there really need to be a full refactor of the code just to add a single "if the target is an IP add it to one list, else add it to the other list"?

dannysauer:
To be clear, this all looks like decent code.  But this kind of huge, multi-functional PR makes it hard to track down the source of an issue later on.  The FQDN bootstrap issue is fixed by altering the building of the IPs list, period - one file involved.  Renaming variables, shifting around includes, adding new logic in other functions...  That's all stuff that should really be in another separate "cleanup" PR.

innobead:
> This is a _lot_ of code to fix a single list element error. Does there really need to be a full refactor of the code just to add a single "if the target is an IP add it to one list, else add it to the other list"?

@jenting beside fixing the issue, he is also trying to add more logging and simplify some code w/ shared logic as well. IMHO, I do not see lots of code change here.

jenting:
> This is a _lot_ of code to fix a single list element error. Does there really need to be a full refactor of the code just to add a single "if the target is an IP add it to one list, else add it to the other list"?

For this case on _if the target is an IP add it to one list, else add it to the other list_, yes you are right.

I moved to generate a server certificate/key with function `NewServerCertAndKey` due to I forget we already have this API :roll_eyes:

And for DNS name-checking, I could remove it in this PR and note a TODO.

innobead:
> To be clear, this all looks like decent code. But this kind of huge, multi-functional PR makes it hard to track down the source of an issue later on. The FQDN bootstrap issue is fixed by altering the building of the IPs list, period - one file involved. Renaming variables, shifting around includes, adding new logic in other functions... That's all stuff that should really be in another separate "cleanup" PR.

Got your point after clear explanation :) Fair point.

dannysauer:
(also, sorry if I'm coming across as more grumpy than normal - it's 3AM here, and my filter might not be effective - I'm actually trying to be constructive. :))

jenting:
@dannysauer Thanks for your late review. I change the PR to _if the target is an IP add it to one list, else add it to the other list_. I will file another refactor PR in the future.

thehejik:
Please wait with the merge, I'm testing the PR

thehejik:
I bootstrapped a cluster on vmware where I used FQDN hostnames instead on IPs for all `skuba` command needed for bootstrap. 

It was successful with skuba from jenting's branch:
skuba version: untagged (v1.2.0-40-g8c0c897c) (development) 20191209 go1.12.6

PR #887, opened 2019-12-10
Add protection against transient states from ssh

manuelbuil:
Sometimes the deployment of terraform fails because of a "connection
refused". For example:

```
* Error running command 'ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null $user@$host sudo reboot || :
ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -oConnectionAttempts=60 $user@$host /usr/bin/true
': exit status 255. Output: Warning: Permanently added '10.17.2.0' (ECDSA) to the list of known hosts.
Connection to 10.17.2.0 closed by remote host.
ssh: connect to host 10.17.2.0 port 22: Connection refused
```

After investigating, it seems that after the reboot, there is a very 
small transient period in which the ssh server asks for password
instead of using the key. If Terraform tries to do the ssh command 
during that period, the deployment fails, even though everything is
fine. When trying manually with a while loop:

```
ssh: connect to host 10.17.3.0 port 22: No route to host
255
ssh: connect to host 10.17.3.0 port 22: No route to host
255
ssh: connect to host 10.17.3.0 port 22: Connection refused
255
ssh: connect to host 10.17.3.0 port 22: Connection refused
255
Warning: Permanently added '10.17.3.0' (ECDSA) to the list of known hosts.
0
Warning: Permanently added '10.17.3.0' (ECDSA) to the list of known hosts.
sign_and_send_pubkey: signing failed: communication with agent failed
Password:
```

After that, the ssh connection works always.

This patch protects ourselves against that because it checks the ssh
port first and then it tries the ssh connection. This way, the 
transient period happens while we are testing the ssh port.

## Why is this PR needed?

This PR fixes an issue we are seeing when using libvirt as the
provider of terraform. Explained in the message of the patch. This
issue was not reported in Jira or Github.

## What does this PR do?

Please refer to the message of the patch

No

## Info for QA

Run the deployment of terraform using libvirt as provider and around
30% of the time, you will see the problem I am describing.

### Related info
Nothing

Please check the message of the patch

Please check the message of the patch

## Docs
No docs are needed

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/libvirt/master-instance.tf
+++ b/ci/infra/libvirt/master-instance.tf
@@ -126,6 +126,7 @@ resource "null_resource" "master_reboot" {
     command = <<EOT
 ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null $user@$host sudo reboot || :
 # wait for ssh ready after reboot
+until nc -zv $host 22; do sleep 5; done
 ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -oConnectionAttempts=60 $user@$host /usr/bin/true
 EOT
   }
--- a/ci/infra/libvirt/worker-instance.tf
+++ b/ci/infra/libvirt/worker-instance.tf
@@ -126,6 +126,7 @@ resource "null_resource" "worker_reboot" {
     command = <<EOT
 ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null $user@$host sudo reboot || :
 # wait for ssh ready after reboot
+until nc -zv $host 22; do sleep 5; done
 ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -oConnectionAttempts=60 $user@$host /usr/bin/true
 EOT
   }


manuelbuil:
Trying again with an improved commit message

manuelbuil:
Early_morning_without_coffee mistake, I closed it but I wanted to just comment

pablochacin:
>  After the reboot there is a very small transient period in which the ssh server asks for password
instead of using the key.

I don't understand this comment. What I see is a connection refused. That basically means that the  node is up but `sshd` is not accepting connections.

manuelbuil:
> > After the reboot there is a very small transient period in which the ssh server asks for password
> > instead of using the key.
> 
> I don't understand this comment. What I see is a connection refused. That basically means that the node is up but `sshd` is not accepting connections.

There must be a state between sshd_not_accepting_connections and sshd_accepting_connections_and_checking_keys in which sshd asks for the password instead of checking the keys, even though the keys and the config are there. I am not sure if it is a problem of the ssh-agent or the sshd. Look at the output of my while loop (printing the RC in each step):

```
ssh: connect to host 10.17.3.0 port 22: No route to host
255
ssh: connect to host 10.17.3.0 port 22: No route to host
255
ssh: connect to host 10.17.3.0 port 22: Connection refused
255
ssh: connect to host 10.17.3.0 port 22: Connection refused
255
Warning: Permanently added '10.17.3.0' (ECDSA) to the list of known hosts.
0
Warning: Permanently added '10.17.3.0' (ECDSA) to the list of known hosts.
sign_and_send_pubkey: signing failed: communication with agent failed
Password: 
```
Note the message "sign_and_send_pubkey: signing failed: communication with agent failed". The keys were correct and if I try again without touching anything, it works:

```
Warning: Permanently added '10.17.3.0' (ECDSA) to the list of known hosts.
0
Warning: Permanently added '10.17.3.0' (ECDSA) to the list of known hosts.
0
Warning: Permanently added '10.17.3.0' (ECDSA) to the list of known hosts.
0
Warning: Permanently added '10.17.3.0' (ECDSA) to the list of known hosts.
```
If I login into the VM, I see:

`There was 1 failed login attempt since the last successful login.`

If I check the sshd messages in journalctl of the VM, I can see:

```
Dec 10 09:08:25 caasp-cluster-worker-0 sshd[1111]: pam_unix(sshd:auth): authentication failure; logname= uid=0 euid=0 tty=ssh ruser= rhost=10.17.0.1  user=sles
Dec 10 09:08:26 caasp-cluster-worker-0 sshd[1111]: Failed password for sles from 10.17.0.1 port 36214 ssh2
Dec 10 09:08:28 caasp-cluster-worker-0 sshd[1111]: Accepted keyboard-interactive/pam for sles from 10.17.0.1 port 36214 ssh2
Dec 10 09:08:28 caasp-cluster-worker-0 sshd[1111]: fatal: PAM: pam_setcred(): Permission denied

```
Why is it doing password authentication instead of publickey authentication? This is what happens when things work:

```
Dec 10 09:21:41 caasp-cluster-worker-0 sshd[1165]: Accepted publickey for sles from 10.17.0.1 port 60524 ssh2: RSA SHA256:KTmTAsyP+BBpGnxnPfS5i0U+NkKFEyJ++3aOrczBmI4
Dec 10 09:21:41 caasp-cluster-worker-0 sshd[1165]: pam_unix(sshd:session): session opened for user sles by (uid=0)
```
The weird thing is that it only happens temporarily and always when the sshd is changing its state from not accepting ssh connections to accepting ssh connections, that's why I concluded that it is a transient state/period

jcaamano:
It looks to me that the password authentication is a fallback from a failed key authentication caused by
`sign_and_send_pubkey: signing failed: communication with agent failed`
which seems to be a local problem to the client when attempting to communicate with the local agent to sign with the private key and also seems to be unrelated to the state of the server.

Also this is different error to the `Connection refused` one originally reported on the terraform attempt. Did you check if the 60 connection attempts `-oConnectionAttempts=60` are really attempted in that case? The attempts are one per second so maybe it is a matter making that number bigger?

ci/infra/libvirt/master-instance.tf
```diff
@@ -126,6 +126,7 @@ resource "null_resource" "master_reboot" {
     command = <<EOT
 ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null $user@$host sudo reboot || :
 # wait for ssh ready after reboot
+until nc -zv $host 22; do sleep 5; done
```
jcaamano: Perhaps do 1 second attempts as ssh connection attempts do.

manuelbuil:
> authentication is a fallback from a failed key authentication caused by
> `sign_and_send_pubkey: signing failed: communication with agent failed`
> which seems to be a local problem to the client when attempting to communicate with the local agent to sign with the private key and also seems



> It looks to me that the password authentication is a fallback from a failed key authentication caused by
> `sign_and_send_pubkey: signing failed: communication with agent failed`
> which seems to be a local problem to the client when attempting to communicate with the local agent to sign with the private key and also seems to be unrelated to the state of the server.
> 
> Also this is different error to the `Connection refused` one originally reported on the terraform attempt. Did you check if the 60 connection attempts `-oConnectionAttempts=60` are really attempted in that case? The attempts are one per second so maybe it is a matter making that number bigger?

Yes, I think the source of the problem is the `sign_and_send_pubkey: signing failed: communication with agent failed` but I think this is related to the state of the sshd because I only saw it when sshd was being restarted (e.g. in the load balancer, this also happens sometimes when the cloud-init agent is changing the config of the sshd server). Increasing the -oConnectionAttempts=60 did not work for me, what I saw is that as soon as the authentication fails, Terraform fails. I'm now wondering if perhaps what happens is that the server is waiting for the password to be written but Terraform writes nothing and thus it fails after a timeout. I'll test again adding a dummy password in the terraform config. In the meanwhile, I also asked in stackoverflow about the ssh-agent error: https://stackoverflow.com/questions/59265089/temporary-problem-with-ssh-agent-right-after-sshd-restarts

ci/infra/libvirt/master-instance.tf
```diff
@@ -126,6 +126,7 @@ resource "null_resource" "master_reboot" {
...
 ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -oConnectionAttempts=60 $user@$host /usr/bin/true
```
dirkmueller: so if the problem is that we want to avoid continuing while passwordless login does not yet succeed, then we should do this here rather than hacking around with nc. you can simply add

```
-o PasswordAuthentication=no -o ChallengeResponseAuthentication=no 
```

ci/infra/libvirt/worker-instance.tf
```diff
@@ -126,6 +126,7 @@ resource "null_resource" "worker_reboot" {
     command = <<EOT
 ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null $user@$host sudo reboot || :
 # wait for ssh ready after reboot
+until nc -zv $host 22; do sleep 5; done
 ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -oConnectionAttempts=60 $user@$host /usr/bin/true
```
dirkmueller: same here

dirkmueller:
Please note that https://github.com/SUSE/avant-garde/issues/1147 talks about removing the code you're patching.

manuelbuil:
> Please note that [SUSE/avant-garde#1147](https://github.com/SUSE/avant-garde/issues/1147) talks about removing the code you're patching.



> Please note that [SUSE/avant-garde#1147](https://github.com/SUSE/avant-garde/issues/1147) talks about removing the code you're patching.

Is this avant-garde project a dummy project to track roadmap tasks? And yes, when that issue gets solved, it will remove the code I am patching. However, I am just adding one liner, so I still think it is fine to merge as the effort is minimum. Nevertheless, what are the best practices here? what would you recommend? Thx!

ci/infra/libvirt/master-instance.tf
manuelbuil: The problem is that it should use the keys to authenticate via the ssh-agent instead of the password. However, right after the sshd server starts after the booting, there is a very small period of time in which ssh-agent fails to communicate the key. If the ssh request is done over that period of time, then Terraform concludes that the task has failed because it was not able to authenticate.

Using a tool (in this case netcat) to check that the ssh port is up and running before trying to do ssh seems to me like best engineering practices and no hack. TBH, perhaps I missed it in the documentation but I am surprised that Terraform does not include any syntax to check if ports are up and running, to me that is basic functionality. Anyway, not doing any ssh until we are sure that the port is up and running avoids the problem I have just described, that's why I consider it as a good solution

dirkmueller: I understand, so ssh will not return the exit code of "/usr/bin/true" (which is 0) but some other error code which indicates "ssh has failed to run the command". and that leads to terraform treating that as a failure. 

Your command will first check whether the port is listening to requests, and then try the request. if I understand the problem properly, then the race is still there, its just much smaller ,because the "restart" might not be finished after the nc succeeded but the ssh attempt afterwards is started. thats why I said "hacking around" (I didn"t say your code is a hack btw, I meant it in the sense of "workaround"). 


we could simply retry the ssh command until it exits with an exit code of 0 (or give up after like 3-5 attempts. that would close the race if I understand your problem description correctly, because then we can be confident that further ssh attempts will succeed.

dirkmueller: Another thing is that there might be some sleeping missing waiting for the node shutdown (because we're just executing a reboot command) so chances are there that we're still hitting the instance before the reboot, rather than waiting until reboot finished with the new ssh being up.

manuelbuil: Yes, that's a good point. @jcaamano was looking at it yesterday

manuelbuil: > I understand, so ssh will not return the exit code of "/usr/bin/true" (which is 0) but some other error code which indicates "ssh has failed to run the command". and that leads to terraform treating that as a failure.
> 
> Your command will first check whether the port is listening to requests, and then try the request. if I understand the problem properly, then the race is still there, its just much smaller ,because the "restart" might not be finished after the nc succeeded but the ssh attempt afterwards is started. thats why I said "hacking around" (I didn"t say your code is a hack btw, I meant it in the sense of "workaround").
> 
> we could simply retry the ssh command until it exits with an exit code of 0 (or give up after like 3-5 attempts. that would close the race if I understand your problem description correctly, because then we can be confident that further ssh attempts will succeed.
> 
> that latter approach is for example what the OpenStack Cloud team used to wait for VMs to be finished booting in order to ssh them. I think it might be worthwhile and its definitely proven in practice.

Not exactly. The command which is executed is:

`ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -oConnectionAttempts=60 $user@$host /usr/bin/true`

That one tries 60 times. That is fine as long as the receiving error is, no IP:
`ssh: connect to host 10.17.3.0 port 22: No route to host`

Or port 22 not up:
`ssh: connect to host 10.17.3.0 port 22: Connection refused`

But it is not fine and terraform fails when there is an authentication error due to the ssh-agent problem `sign_and_send_pubkey: signing failed: communication with agent failed`. Terraform fails regardless of how many times it has tried before, it could be the first time it tries and it will not try another 59 times, it will simply fail.

You are right that there is a race condition (the results in the ssh-agent problem) and this solution is not removing it but reducing the likelihood of it. I actually haven't seen it again since I started using this patch but there is probably still a chance. Therefore, we can consider other options but it cannot be by using ssh because we will be hitting the ssh-agent problem again. One alternative that would probably work is to disable ssh-agent in that terraform task but as ssh-agent is used across our Terraform deployment, I expect a lot of negative reviews.

Having said so (sorry for extending myself so much ;) ), I still think that checking the ssh port is up and running before doing ssh is something that should be always done

dirkmueller: I think I understand. Terraform fails when there is a comnabd exiting with exit code nonzero. But you can capture that and sleep while the exit code is nonzero like (pseudo code ahead)


```
while ! ssh ... ; do
  sleep 1
  if retriedtoooften; then
     exit 1
  fi
done
```

PR #891, opened 2019-12-10
cluster upgrade plan: remove misleading message

ereslibre:
## Why is this PR needed?

Remove this misleading message from `cluster upgrade plan`. We can
improve this situation by properly composing a list of addons based
on:

- Current addons version (for the current cluster version)
  - This is essentially our `skuba-config` ConfigMap contents

- Desired addons version (for the current cluster version)
  - This is essentially mapping the current versions and projecting
    them to their desired versions in the versions.go file, for the
    current cluster version.

- Desired addons version (for the next cluster version)
  - As the previous step, but projecting to the next cluster version.

By performing this logic, our `cluster upgrade plan` would be more
complete, so that we can report:

- Whether a `skuba addon upgrade apply` is required before trying to
  upgrade the platform (current addons version does not match desired
  addons version for the current cluster version)

- Versions that will be bumped when upgrading the platform

- Whether a `skuba addon upgrade apply` is required after upgrading
  the platform (desired addons version for the current cluster version
  does not match desired addons version for the next cluster version)

/cc @SUSE/caasp-qa 

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/pkg/skuba/actions/cluster/upgrade/plan.go
+++ b/pkg/skuba/actions/cluster/upgrade/plan.go
@@ -81,8 +81,6 @@ func Plan(client clientset.Interface) error {
 	}
 	fmt.Println()
 	if addon.HasAddonUpdate(updatedAddons) {
-		fmt.Println("After you have completed the platform upgrade, there are new addon versions available:")
-		fmt.Println()
 		fmt.Printf("Addon upgrades from %s to %s:\n", currentVersion, nextClusterVersion.String())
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
Removing from 4.1.0 given this https://github.com/SUSE/skuba/commit/0cba02dc11b958de1ccd1812b070785a92df0efb is not in caasp-release-4.1.0 branch

PR #892, opened 2019-12-10
Be explicit about not using ssh-agent

manuelbuil:
Cloud-init is taking care of setting the correct keys and ssh config
and thus it should not use key authentication. Our config is 
correct because it provides the password; however, if nothing is set,
then the default is agent = true, which does not make sense.

For example:

```
null_resource.master_wait_cloudinit[0]: Provisioning with
'remote-exec'...
null_resource.master_wait_cloudinit[0] (remote-exec): Connecting to
remote host via SSH...
null_resource.master_wait_cloudinit[0] (remote-exec):   Host: 10.17.2.0
null_resource.master_wait_cloudinit[0] (remote-exec):   User: sles
null_resource.master_wait_cloudinit[0] (remote-exec):   Password: true
null_resource.master_wait_cloudinit[0] (remote-exec):   Private key:
false
null_resource.master_wait_cloudinit[0] (remote-exec):   Certificate:
false
null_resource.master_wait_cloudinit[0] (remote-exec):   SSH Agent: true
null_resource.master_wait_cloudinit[0] (remote-exec):   Checking Host Key:
false
```

As a result, sometimes the terraform deploy fails at cloud-init step
when changing from password to key authentication in the same terraform
task (cloud-init)

## Why is this PR needed?
Yes, please read commit message

## What does this PR do?
Please read commit message

No

## Info for QA
This PR fixes a bug which appears from time to time when deploying terraform + libvirt

Deploying skuba with terraform + libvirt as provider. It sometimes fail in the cloud_init step with a ssh problem

The deployment never fails in the cloud_init step

## Docs
No docs needed

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/libvirt/lb-instances.tf
+++ b/ci/infra/libvirt/lb-instances.tf
@@ -110,6 +110,7 @@ resource "null_resource" "lb_wait_cloudinit" {
     user     = "${var.username}"
     password = "${var.password}"
     type     = "ssh"
+    agent    = false
   }
 
   provisioner "remote-exec" {
--- a/ci/infra/libvirt/master-instance.tf
+++ b/ci/infra/libvirt/master-instance.tf
@@ -104,6 +104,7 @@ resource "null_resource" "master_wait_cloudinit" {
     user     = "${var.username}"
     password = "${var.password}"
     type     = "ssh"
+    agent    = false
   }
 
   provisioner "remote-exec" {
--- a/ci/infra/libvirt/worker-instance.tf
+++ b/ci/infra/libvirt/worker-instance.tf
@@ -104,6 +104,7 @@ resource "null_resource" "worker_wait_cloudinit" {
     user     = "${var.username}"
     password = "${var.password}"
     type     = "ssh"
+    agent    = false
   }
 
   provisioner "remote-exec" {


jcaamano:
We opened a relevant terraform issue https://github.com/hashicorp/terraform/issues/23662.

For whatever reason agent forwarding session setup is failing, and that causes a deadlock. In theory the fix should allow further connection attempts to be made and I have not been able to reproduce the problem with it.

But anyway, whatever is causes the issue may happen t at a later point in time when the connection is already established, for which no reattempt will be made, provisioner will fail and resource be tainted. If cloud-init is causing established ssh connections to drop, then a more adequate workaround could be to replace remote-exec with local-exec script that is a bit smarter than the remote-exec provisioner.

lcavajani:
@manuelbuil you should not rely on using a username/password to authenticate to the nodes. Username/password authentication will be removed ASAP in favor of `ssh-agent` auth only. We normally only allow this type of authentication because it's the only one supported in `skuba`, this way we can ensure early in the deployment that it works.

FI, it's still around because `libvirt` was not in the backlog and it was somehow best effort, we will work on it to make it 1:1 with the other platforms.

e.g in openstack https://github.com/SUSE/skuba/blob/master/ci/infra/openstack/master-instance.tf#L92-L96

manuelbuil:
> @manuelbuil you should not rely on using a username/password to authenticate to the nodes. Username/password authentication will be removed ASAP in favor of `ssh-agent` auth only. We normally only allow this type of authentication because it's the only one supported in `skuba`, this way we can ensure early in the deployment that it works.
> 
> FI, it's still around because `libvirt` was not in the backlog and it was somehow best effort, we will work on it to make it 1:1 with the other platforms.
> 
> e.g in openstack https://github.com/SUSE/skuba/blob/master/ci/infra/openstack/master-instance.tf#L92-L96

Thanks for the information. I agree that ssh key authentication is much better. The reason why we disable ssh-agent in the cloud-init step is because we detected that sometimes, when cloud-init is configuring the ssh server (keys and password), the connection with ssh-agent breaks and Terraform does not retry again and thus marks the deployment as failure. As jcaamno commented, he has opened an issue in Terraform upstream which points at the problem ==> https://github.com/hashicorp/terraform/issues/23662. 

For the time being, this workaround makes our deployments always successful. I must say that I have never tried removing the password as your example with openstack shows. In that case, I guess we must make sure that a connection timeout is not triggered because it will not be able to connect until cloud-init configures the ssh part. I'll give it a try

manuelbuil:
@lcavajani I tried your suggestion and it also worked, so instead of adding agent = false. This PR gets replaced by https://github.com/SUSE/skuba/pull/902 and thus I close this one

PR #897, opened 2019-12-11
Reduce number of calls on version check

Itxaka:
On running node upgrade plan, there is several calls to the api to
check the existance of several pods.

We can minimize this by just calling it once and then checking over
the list for the pods that we want to check for. This should reduce
the number of calls per control plane to 1 instead of 4

## Why is this PR needed?

This is the first step into introducing a retry feature on this checks by reducing the calls, thus making the retry feature less expensive.

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubernetes/node_version.go
+++ b/internal/pkg/skuba/kubernetes/node_version.go
@@ -179,29 +179,39 @@ func nodeVersioningInfo(client clientset.Interface, nodeName string) (NodeVersio
 
 	// find out the container image tags, depending on the role of the node
 	if IsControlPlane(nodeObject) {
-		apiServerTag, err := getPodContainerImageTag(client, metav1.NamespaceSystem, fmt.Sprintf("%s-%s", "kube-apiserver", nodeName))
+		// list all the pods
+		allPods, err := client.CoreV1().Pods(metav1.NamespaceSystem).List(metav1.ListOptions{})
+		// check for error
 		if err != nil {
-			return NodeVersionInfo{}, errors.Wrap(err, "could not retrieve apiserver pod")
+			return NodeVersionInfo{}, errors.Wrap(err, "could not retrieve pods")
 		}
-		controllerManagerTag, err := getPodContainerImageTag(client, metav1.NamespaceSystem, fmt.Sprintf("%s-%s", "kube-controller-manager", nodeName))
+		// check for empty pod list
+		if len(allPods.Items) == 0 {
+			return NodeVersionInfo{}, errors.New("list of pods is empty")
+		}
+		// check that the needed pods exist
+		apiserverPod, err := getPodFromPodList(allPods, fmt.Sprintf("kube-apiserver-%s", nodeName))
 		if err != nil {
-			return NodeVersionInfo{}, errors.Wrap(err, "could not retrieve controller-manager pod")
+			return NodeVersionInfo{}, errors.Wrap(err, "could not retrieve api server pod")
 		}
-		schedulerTag, err := getPodContainerImageTag(client, metav1.NamespaceSystem, fmt.Sprintf("%s-%s", "kube-scheduler", nodeName))
+		controllerManagerPod, err := getPodFromPodList(allPods, fmt.Sprintf("kube-controller-manager-%s", nodeName))
+		if err != nil {
+			return NodeVersionInfo{}, errors.Wrap(err, "could not retrieve controller manager pod")
+		}
+		schedulerPod, err := getPodFromPodList(allPods, fmt.Sprintf("kube-scheduler-%s", nodeName))
 		if err != nil {
 			return NodeVersionInfo{}, errors.Wrap(err, "could not retrieve scheduler pod")
 		}
-		etcdTag, err := getPodContainerImageTag(client, metav1.NamespaceSystem, fmt.Sprintf("%s-%s", "etcd", nodeName))
+		etcdPod, err := getPodFromPodList(allPods, fmt.Sprintf("etcd-%s", nodeName))
 		if err != nil {
 			return NodeVersionInfo{}, errors.Wrap(err, "could not retrieve etcd pod")
 		}
 
-		nodeVersions.APIServerVersion = version.MustParseSemantic(apiServerTag)
-		nodeVersions.ControllerManagerVersion = version.MustParseSemantic(controllerManagerTag)
-		nodeVersions.SchedulerVersion = version.MustParseSemantic(schedulerTag)
-		nodeVersions.EtcdVersion = version.MustParseSemantic(etcdTag)
+		nodeVersions.APIServerVersion = version.MustParseSemantic(getPodContainerImageTagFromPodObject(apiserverPod))
+		nodeVersions.ControllerManagerVersion = version.MustParseSemantic(getPodContainerImageTagFromPodObject(controllerManagerPod))
+		nodeVersions.SchedulerVersion = version.MustParseSemantic(getPodContainerImageTagFromPodObject(schedulerPod))
+		nodeVersions.EtcdVersion = version.MustParseSemantic(getPodContainerImageTagFromPodObject(etcdPod))
 	}
-
 	return nodeVersions, nil
 }
 
--- a/internal/pkg/skuba/kubernetes/pod.go
+++ b/internal/pkg/skuba/kubernetes/pod.go
@@ -18,9 +18,11 @@
 package kubernetes
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 )
@@ -30,8 +32,23 @@ func getPodContainerImageTag(client clientset.Interface, namespace string, podNa
 	if err != nil {
 		return "", errors.Wrap(err, "could not retrieve pod object")
 	}
+	return getPodContainerImageTagFromPodObject(podObject), nil
+}
+
+// return the image version tag without calling the api, just with the Pod object
+func getPodContainerImageTagFromPodObject(podObject *v1.Pod) string {
 	containerImageWithName := podObject.Spec.Containers[0].Image
 	containerImageTag := strings.Split(containerImageWithName, ":")
 
-	return containerImageTag[len(containerImageTag)-1], nil
+	return containerImageTag[len(containerImageTag)-1]
+}
+
+// return a pod object matched from a podList
+func getPodFromPodList(list *v1.PodList, name string) (*v1.Pod, error) {
+	for _, pod := range list.Items {
+		if name == pod.Name {
+			return &pod, nil
+		}
+	}
+	return nil, fmt.Errorf("could not find pod %s in pod list", name)
 }
--- a/internal/pkg/skuba/kubernetes/pod_test.go
+++ b/internal/pkg/skuba/kubernetes/pod_test.go
@@ -97,3 +97,70 @@ func Test_getPodContainerImageTag(t *testing.T) {
 		})
 	}
 }
+
+func Test_getPodFromPodList(t *testing.T) {
+	podList := corev1.PodList{
+		TypeMeta: metav1.TypeMeta{},
+		ListMeta: metav1.ListMeta{},
+		Items:    make([]corev1.Pod, 2),
+	}
+	validPod := corev1.Pod{
+		TypeMeta:   metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{Name: "valid"},
+		Spec:       corev1.PodSpec{},
+		Status:     corev1.PodStatus{},
+	}
+	anotherPod := corev1.Pod{
+		TypeMeta:   metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{Name: "Another"},
+		Spec:       corev1.PodSpec{},
+		Status:     corev1.PodStatus{},
+	}
+
+	podList.Items[0] = validPod
+	podList.Items[1] = anotherPod
+
+	tests := []struct {
+		list         corev1.PodList
+		name         string
+		expect       *corev1.Pod
+		expectErrMsg string
+	}{
+		{
+			list:   podList,
+			name:   "valid",
+			expect: &validPod,
+		},
+		{
+			list:         podList,
+			name:         "invalid",
+			expectErrMsg: "could not find pod invalid in pod list",
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt // Parallel testing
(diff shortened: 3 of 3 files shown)


internal/pkg/skuba/kubernetes/pod.go
```diff
@@ -30,8 +32,23 @@ func getPodContainerImageTag(client clientset.Interface, namespace string, podNa
...
+}
+
+// return a pod object matched from a podList
+func getPodFromPodList(list []v1.Pod, name string) (v1.Pod, error) {
+	for _, pod := range list {
+		if name == pod.Name {
+			return pod, nil
+		}
+	}
+	return v1.Pod{}, errors.New(fmt.Sprintf("Could not find pod %s in pod list", name))
```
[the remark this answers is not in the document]
jenting: As [1], I would prefer is pod not exist, return `nil`.

internal/pkg/skuba/kubernetes/pod.go
```diff
@@ -30,8 +32,23 @@ func getPodContainerImageTag(client clientset.Interface, namespace string, podNa
...
+func getPodContainerImageTagFromPodObject(podObject v1.Pod) string {
 	containerImageWithName := podObject.Spec.Containers[0].Image
 	containerImageTag := strings.Split(containerImageWithName, ":")
 
-	return containerImageTag[len(containerImageTag)-1], nil
+	return containerImageTag[len(containerImageTag)-1]
+}
+
+// return a pod object matched from a podList
+func getPodFromPodList(list []v1.Pod, name string) (v1.Pod, error) {
```
jenting: As [1], I would prefer v1.Pod use pointer `*v1.Pod`.

internal/pkg/skuba/kubernetes/pod.go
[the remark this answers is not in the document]
Itxaka: > As [1], I would prefer is pod not exist, return nil.

So for this, my first approach was to return `nil`, but as the func returns a `v1.Pod, error` it was complaining that I was returning a `nil` instead of a expected `v1.Pod`

Is this expected and safe to ignore?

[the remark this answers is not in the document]
Itxaka: >  This line could be simply as fmt.Errorf("Could not find pod %s in pod list", name)

Nice, thanks!

internal/pkg/skuba/kubernetes/pod.go
```diff
@@ -30,8 +32,23 @@ func getPodContainerImageTag(client clientset.Interface, namespace string, podNa
 	if err != nil {
 		return "", errors.Wrap(err, "could not retrieve pod object")
 	}
+	return getPodContainerImageTagFromPodObject(*podObject), nil
+}
+
+// return the image version tag without calling the api, just with the Pod object
+func getPodContainerImageTagFromPodObject(podObject v1.Pod) string {
```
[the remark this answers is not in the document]
Itxaka: is there any advantage in using the pointer here? Avoid resolving the pointer to the object? Or is there something Im missing?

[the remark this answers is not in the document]
jenting: My first thought is no need to dereference when calling `getPodContainerImageTagFromPodObject`, so the `getPodContainerImageTagFromPodObject(*podObject)` could be change to 
`getPodContainerImageTagFromPodObject(podObject)`.

internal/pkg/skuba/kubernetes/pod.go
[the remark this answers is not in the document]
jenting: > > As [1], I would prefer is pod not exist, return nil.
> 
> So for this, my first approach was to return `nil`, but as the func returns a `v1.Pod, error` it was complaining that I was returning a `nil` instead of a expected `v1.Pod`

golint or golang-ci complaing?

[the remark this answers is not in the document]
Itxaka: `make lint`

```
GO111MODULE=on go vet ./...
# github.com/SUSE/skuba/internal/pkg/skuba/kubernetes
internal/pkg/skuba/kubernetes/pod.go:53:2: cannot use nil as type "k8s.io/api/core/v1".Pod in return argument
make: *** [Makefile:94: lint] Error 2
```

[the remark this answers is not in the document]
jenting: I saw the slice []v1.Pod is not a pointer.
Let's keep your original code, it seems the elegant way.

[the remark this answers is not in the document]
Itxaka: oh yeah, I see. If I change it to return `*v1.Pod` I can return nil :)

Itxaka:
@jenting updated! Lets see if its better now, thanks for the feedback!

internal/pkg/skuba/kubernetes/node_version.go
```diff
@@ -179,29 +178,39 @@ func nodeVersioningInfo(client clientset.Interface, nodeName string) (NodeVersio
...
+		allPods, err := client.CoreV1().Pods(metav1.NamespaceSystem).List(metav1.ListOptions{})
+		// check for error
 		if err != nil {
-			return NodeVersionInfo{}, errors.Wrap(err, "could not retrieve apiserver pod")
+			return NodeVersionInfo{}, errors.Wrap(err, "Could not retrieve pods")
 		}
-		controllerManagerTag, err := getPodContainerImageTag(client, metav1.NamespaceSystem, fmt.Sprintf("%s-%s", "kube-controller-manager", nodeName))
+		// check for empty pod list
+		if len(allPods.Items) == 0 {
+			return NodeVersionInfo{}, errors.Wrap(err, "List of pods is empty")
```
ereslibre: errors are typically started with lowercase :), applies to all errors on this PR

ereslibre: Besides this, `errors.Wrap` will return `nil` if `err` is `nil` (this is the case here), so you need to create a new error here instead of wrapping.

internal/pkg/skuba/kubernetes/pod.go
```diff
@@ -30,8 +32,23 @@ func getPodContainerImageTag(client clientset.Interface, namespace string, podNa
...
+func getPodContainerImageTagFromPodObject(podObject *v1.Pod) string {
 	containerImageWithName := podObject.Spec.Containers[0].Image
 	containerImageTag := strings.Split(containerImageWithName, ":")
 
-	return containerImageTag[len(containerImageTag)-1], nil
+	return containerImageTag[len(containerImageTag)-1]
+}
+
+// return a pod object matched from a podList
+func getPodFromPodList(list *v1.PodList, name string) (*v1.Pod, error) {
```
ereslibre: A minor improvement in terms of speed would be to construct a map of `map[string]*v1.Pod` (or even `map[string]string`) with the pod name as key (in the second case we would have the image as the value). That way we iterate only once, constructing this map, and read from the map.

aojea:
now I'm curious, how is the client-go cache working in these cases?
https://github.com/kubernetes/client-go/blob/a32a6f7a3052374fff2c57c5b4b52e64dcd55eed/listers/core/v1/pod.go#L49

Itxaka:
> now I'm curious, how is the client-go cache working in these cases?
https://github.com/kubernetes/client-go/blob/a32a6f7a3052374fff2c57c5b4b52e64dcd55eed/listers/core/v1/pod.go#L49

Im also curious. We seem to use this GET/LIST functions[0] that do not use the cache. Should we use those with caching instead?

Probably @ereslibre can answer this :D

[0] https://github.com/kubernetes/client-go/blob/a32a6f7a3052374fff2c57c5b4b52e64dcd55eed/kubernetes/typed/core/v1/pod.go#L83

internal/pkg/skuba/kubernetes/node_version.go
Itxaka: > errors are typically started with lowercase :), applies to all errors on this PR

Thats so weird. Fixed.

> Besides this, errors.Wrap will return nil if err is nil (this is the case here), so you need to create a new error here instead of wrapping.

Fixed!

internal/pkg/skuba/kubernetes/pod.go
Itxaka: I thought about that. Will it be faster? does adding that complexity and unreadability in code outweighs the speed gain? (I guess we should test that with max pods (250) to see the speed change.

internal/pkg/skuba/kubernetes/node_version.go
aojea: it's a golang thing https://www.reddit.com/r/golang/comments/241u95/best_practices_for_errors_in_go/

internal/pkg/skuba/kubernetes/pod.go
aojea: you have an example with the nodes info, they are returning a map
https://github.com/SUSE/skuba/pull/897/files#diff-9236f8f0e4ba056ee3aa9030db67f0a0R125-R142

find an element in a map O(1) vs O(N) using a for loop :)

Itxaka: that doesnt really answer my question of introducing complexity vs speed gain :p

Maybe is not that complex and it only seems to be because Im a go newbie, you know as a pythonista we say "its fast enough" :P

ereslibre:
> Im also curious. We seem to use this GET/LIST functions[0] that do not use the cache. Should we use those with caching instead?
>
> Probably @ereslibre can answer this :D

The error about a not found pod are not related to caching here IMO. We are not listing and then getting currently, but getting pod names provided we know beforehand their names. In this case (what this PR is fixing) is when we try to get the version of a component based on the pod name and it doesn't exist at that time (maybe it's getting recreated for example -- here the apiserver takes some time to notice that it got restarted).

Caching on skuba is a double edged sword IMO, because since it's a one-shot application, the cache will always be cold and usually we want up to date information when retrying and so on, so I believe in this case we don't want caching, but the most up-to-date information possible. If we had an endless loop querying the apiserver that's different, because you don't want to hammer the apiserver that bad, but this is not our case. So in my opinion we should try to make as few requests as possible, but when we make a "request" I think it's important that we get fresh information instead of cached information that happens transparently from us.

internal/pkg/skuba/kubernetes/pod.go
aojea: I think that both approaches should be fine for this case, just a bit of bikeshedding on Mondays doesn't hurt :)

ereslibre: I agree that it may not be worth at this time (maybe O(n) is faster with this small amount than lookups O(1) and insertions O(n*log(n))? However if the number of pods grow on `kube-system` we will probably hit where both complexities cross and a map is always more efficient after that crossing.

I'm fine with your current solution though.

Itxaka: To be fair, Im not against that, but I rather move this forward to then introduce the retry part ASAP as that fixes a bug, *then* rework this by adding the map-ing to squeeze more speed out of it

aojea:
overall LGTM, missing unit tests for the new function :smile:

internal/pkg/skuba/kubernetes/pod.go
```diff
@@ -30,8 +32,23 @@ func getPodContainerImageTag(client clientset.Interface, namespace string, podNa
...
+	for _, pod := range list.Items {
+		if name == pod.Name {
+			return &pod, nil
+		}
+	}
+	return nil, fmt.Errorf("Could not find pod %s in pod list", name)
```
ereslibre:
```suggestion
	return nil, fmt.Errorf("could not find pod %s in pod list", name)
```

internal/pkg/skuba/kubernetes/node_version.go
```diff
@@ -179,29 +178,39 @@ func nodeVersioningInfo(client clientset.Interface, nodeName string) (NodeVersio
...
-			return NodeVersionInfo{}, errors.Wrap(err, "could not retrieve apiserver pod")
+			return NodeVersionInfo{}, errors.Wrap(err, "could not retrieve pods")
 		}
-		controllerManagerTag, err := getPodContainerImageTag(client, metav1.NamespaceSystem, fmt.Sprintf("%s-%s", "kube-controller-manager", nodeName))
+		// check for empty pod list
+		if len(allPods.Items) == 0 {
+			return NodeVersionInfo{}, errors.New("list of pods is empty")
+		}
+		// check that the needed pods exist
+		apiserverPod, err := getPodFromPodList(allPods, "kube-apiserver-"+nodeName)
```
ereslibre: Could be a matter of taste, but I prefer to use `Sprintf` when tweaking strings:

```suggestion
		apiserverPod, err := getPodFromPodList(allPods, fmt.Sprintf("kube-apiserver-%s", nodeName))
```

This applies to the other ones as well, if you prefer to keep the current way of appending the string it's fine for me :), just a matter of taste :)

internal/pkg/skuba/kubernetes/pod.go
Itxaka: done!

internal/pkg/skuba/kubernetes/node_version.go
Itxaka: done!

Itxaka:
unit tests added @aojea :p

internal/pkg/skuba/kubernetes/pod.go
cmurphy: The default number of pods in kube-system looks like only 20:

```
$ kubectl --kubeconfig ./admin.conf get pods -n kube-system | wc -l
20
```
which is basically nothing. Is it likely to grow many more pods? Even if it was 100 I don't think that's so much to be worth comparing O(1) to O(N).

ereslibre: We don't know to what can grow (users can deploy whatever they prefer on this namespace), but I agree there's no real benefit in thinking about optimizing this at this point (if we have evidence that there's an issue we can take action, but it's perfectly fine for now IMO).

aojea:
> unit tests added @aojea :p

@itxaka check if is worth to use these helper functions to create a list with 1000 pods so we cover the scenario with a large number of pods https://github.com/Itxaka/skuba/pull/1/files

Itxaka:
> @Itxaka check if is worth to use these helper functions to create a list with 1000 pods so we cover the scenario with a large number of pods Itxaka/skuba/pull/1/files

Looks good! I wonder if benchmark stuff should go in a different set of tests or should be marked somehow, in the end its a different territory than a unit test...

Itxaka:
@aojea I merged this because I need to move forward with the retry stuff first but I wont forget about the 1000 podlist when I have some spare cycles!

PR #898, opened 2019-12-11
Introduce retry on nodeVersioningInfo

Itxaka:
## Why is this PR needed?

Introduces a retry with backoff while getting the pods version info
as to avoid failing as soon as we cannot find one pod.
    
It could happen that there is a temporal problem that prevents us
from continuing, so retrying is the way to go
    
Also introduces extra testing for nodeVersioningInfo to check that
we are raising the proper errors on failure

Fixes: https://github.com/SUSE/avant-garde/issues/1129

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubernetes/node_version.go
+++ b/internal/pkg/skuba/kubernetes/node_version.go
@@ -25,7 +25,9 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/version"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


internal/pkg/skuba/util/retry.go
```diff
@@ -0,0 +1,19 @@
...
+		if attempts--; attempts > 0 {
+			jitter := time.Duration(rand.Int63n(int64(sleep)))
+			sleep = sleep + jitter/2
+			time.Sleep(sleep)
+			return Retry(attempts, 2*sleep, fn)
+		}
+		return err
+	}
+	return nil
+}
```
ereslibre: We should use https://github.com/kubernetes/kubernetes/blob/b141a990c39da1295c8be521a9dea3317d67082e/staging/src/k8s.io/apimachinery/pkg/util/wait/wait.go#L290, or https://github.com/kubernetes/kubernetes/blob/b141a990c39da1295c8be521a9dea3317d67082e/staging/src/k8s.io/client-go/util/retry/util.go#L48 depending on the purpose.

internal/pkg/skuba/kubernetes/node_version.go
```diff
@@ -179,29 +180,47 @@ func nodeVersioningInfo(client clientset.Interface, nodeName string) (NodeVersio
...
-		controllerManagerTag, err := getPodContainerImageTag(client, metav1.NamespaceSystem, fmt.Sprintf("%s-%s", "kube-controller-manager", nodeName))
-		if err != nil {
-			return NodeVersionInfo{}, errors.Wrap(err, "could not retrieve controller-manager pod")
-		}
-		schedulerTag, err := getPodContainerImageTag(client, metav1.NamespaceSystem, fmt.Sprintf("%s-%s", "kube-scheduler", nodeName))
-		if err != nil {
-			return NodeVersionInfo{}, errors.Wrap(err, "could not retrieve scheduler pod")
-		}
-		etcdTag, err := getPodContainerImageTag(client, metav1.NamespaceSystem, fmt.Sprintf("%s-%s", "etcd", nodeName))
+		err := skubautil.Retry(5, time.Second, func() error {
```
ereslibre: Retrying the whole block will make all steps inside the block being retried. I think we should only retry on a per-call basis, so what we already retrieved is not retried if a later step fails.

Itxaka: but...we need to retry the api call top get a fresh status and that means that some of the pods can have a state change meanwhile. We cant guaranteee that after X retries the pods that we checked and __were__ ok are __now__ ok. 

Either all of them are ok at the time of the check, or we need to recheck them. Performance wise, shouldnt be that big of a problem as we already have the list, so Im not sure why we would avoid rechecking them all.

Plus...this means we can move later to check their actual status (Running for example) and in that case we __really__ need all of them to be on that state at the same time, not later.

I can avoid checking already checked ones, but I dont think its such a good idea, could lead to issues down the line where we check pod X and it was ok but after X retries due to pod Y, it was no longer ok and we would no think this was the issue.

internal/pkg/skuba/util/retry.go
Itxaka: done!

ereslibre:
https://github.com/SUSE/skuba/pull/911 is refactoring this code so it won't use a client to retrieve this information (only once).

I'm fine merging this (I'll get conflicts), or we can add the retry afterwards on the refactored code. WDYT @Itxaka?

PR #903, opened 2019-12-17
Benchmark tests

aojea:
## Why is this PR needed?

It allows running benchmark tests, that may be useful if we need to test our code at scale.

## What does this PR do?

Adds a new option `make test-bench` to the Makefile so it can runs benchmark tests, and runs it automatically when using `make test`
It also implements an example benchmarking the function `getPodFromList()`

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubernetes/pod_test.go
+++ b/internal/pkg/skuba/kubernetes/pod_test.go
@@ -18,6 +18,7 @@
 package kubernetes
 
 import (
+	"fmt"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


Itxaka:
Looks great!

Should we track the benchmarking in jenkins? That way we can see how it evolves over time and how changes affect the performance of the code :)

internal/pkg/skuba/kubernetes/pod_test.go
```diff
@@ -18,10 +18,12 @@
 package kubernetes
 
 import (
+	"fmt"
 	"reflect"
 	"testing"
 
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
c3y1huang: Duplicates?

aojea: good catch

VScode autocomplete :man_shrugging:

aojea: solved

PR #908, opened 2020-01-02
Pass verbose level to kubeadm

jenting:
## Why is this PR needed?

The skuba verbosity does not distribute to kubeadm, sometimes make it hard to troubleshoot kubeadm issue without pass the verbosity.

Fixes #

appear in the changelog.

## What does this PR do?

Pass skuba verbosity to kubeadm.

N/A

## Info for QA

### Related info

* upstream issues

Setup verbosity to 1, the log output of kubeadm part DOES NOT display klog.V(1) related log.

Setup verbosity to 1, the log output of kubeadm part DOES display klog.V(1) related log.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/cmd/skuba/flags/verbose.go
+++ b/cmd/skuba/flags/verbose.go
@@ -21,6 +21,7 @@ import (
 	"flag"
 
 	"github.com/spf13/pflag"
+	"k8s.io/klog"
 )
 
 const (
@@ -46,5 +47,7 @@ func RegisterVerboseFlag(local *pflag.FlagSet) {
 		pflagFlag.Name = verboseLevelFlagLong
 		pflagFlag.Usage = verboseLevelFlagUsage
 		local.AddFlag(pflagFlag)
+	} else {
+		klog.Fatalf("failed to find flag in global flagset (flag): %s", verboseLevelFlagShort)
 	}
 }
--- a/cmd/skuba/flags/verbose.go
+++ b/cmd/skuba/flags/verbose.go
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2020 SUSE LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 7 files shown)


cmd/skuba/main.go
```diff
@@ -74,15 +73,3 @@ func printVersionStatement() {
 		fmt.Printf("** This is a tagged version (%s), but is not built in release mode (mode: %q.) **\n", skubapkg.Tag, skubapkg.BuildType)
 	}
 }
-
-// register adds a flag to local that targets the Value associated with the Flag named globalName in flag.CommandLine.
```
jodavis-suse: Nit: the comment explaining what is happening with 'local' is not reflected in the new verbose.go file.

cmd/skuba/flags/verbose.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// RegisterVerboseFlag register verbose flag.
+func RegisterVerboseFlag(local *pflag.FlagSet) {
+	if f := flag.CommandLine.Lookup(verboseLevelFlagShort); f != nil {
+		pflagFlag := pflag.PFlagFromGoFlag(f)
+		pflagFlag.Name = verboseLevelFlagLong
+		pflagFlag.Usage = verboseLevelFlagUsage
+		local.AddFlag(pflagFlag)
+	}
```
Klaven: should we log fatal here if this fails? I'm trying to think of why this might fail... in general I would like to see a failure here I think maybe, but might not be a big deal? :thinking:

jodavis-suse: I can't think of a reason it would fail here.  The parsing handles most problems and will spit out the help message if something doesn't match (like "-v somestring").  Any other problem would be an internal coding error (like trying to register the flag twice) which are detected by our testing.

jenting: Add klog back.

Now the `v` flag register by klog package. Skuba changes the name and usage output string (different to klog).

The situation _Lookup_ failed is the `v` flag is not register by klog (we replace the klog package to other logging package). For easier debugging for the engineer, I added it back.

PR #912, opened 2020-01-14
libvirt: Fix ip in testrunner

Itxaka:
commit 084b421e048c5a17f47c65534e66c090a4092295 introduced a change in
the output for the libvirt deployments which now included the node
names.

Unfortunately testrunner was not updated to use that new format which
results in the impossibility of bootstraping the master node as we
return a dictionary instead of the master0 IP/FQDN.

This patch fixes that by returning the IP as before

## Why is this PR needed?

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/platforms/libvirt.py
+++ b/ci/infra/testrunner/platforms/libvirt.py
@@ -17,3 +17,15 @@ class Libvirt(Terraform):
     def _cleanup_platform(self):
         self.destroy()
 
+    def get_lb_ipaddr(self):
+        self._load_tfstate()
+        return self.state["modules"][0]["outputs"]["ip_load_balancer"]["value"]["{}-lb".format(self.conf.terraform.stack_name)]
+
+    def get_nodes_ipaddrs(self, role):
+        self._load_tfstate()
+
+        if role not in ("master", "worker"):
+            raise ValueError("Invalid role: {}".format(role))
+        role_key = "ip_" + role + "s"
+
+        return list(self.state["modules"][0]["outputs"][role_key]["value"].values())


pablochacin:
Wondering if instead of fixing this, revert the commit that introduced the change in the format, mostly as it makes the output from libvirt different from the output from other providers, and I honestly don't see the advantage. IMHO consistency should prevail over ad-hoc "optimizations" unless there's a good reason.

Itxaka:
> Wondering if instead of fixing this, revert the commit that introduced the change in the format, mostly as it makes the output from libvirt different from the output from other providers, and I honestly don't see the advantage. IMHO consistency should prevail over ad-hoc "optimizations" unless there's a good reason.

Sounds good to me, but wanted to check with @suseclee first to see why was this introduced, but meanwhile others can move forward if they are using libvirt by applying this patch :D

Im in favour of reverting the other patch instead, as at a glance it doesnt introduce any useful features.

evrardjp:
I am biased against not reverting as it means that I will have to update my tf patch :rofl: 
I would prefer reverting too.

pablochacin:
> I am biased against not reverting as it means that I will have to update my tf patch 
> I would prefer reverting too.

I found this justification for approving the PR rather shallow. I think the implications in the code base should have a higher priority than preferences. IMHO

evrardjp:
> I found this justification for approving the PR rather shallow

The comment above was a joke.

> I think the implications in the code base should have a higher priority than preferences. IMHO

I agree with that. Although I believe that the patch that was incompletely applied can be either reverted, or the pattern it sets can be used everywhere else _later_. In the meantime, there is no reason to **block** something actually fixing the current state, which means we should merge it.

suseclee:
> Wondering if instead of fixing this, revert the commit that introduced the change in the format, mostly as it makes the output from libvirt different from the output from other providers, and I honestly don't see the advantage. IMHO consistency should prevail over ad-hoc "optimizations" unless there's a good reason.

I would rather to fix outputs for all other platforms.

suseclee:
> > Wondering if instead of fixing this, revert the commit that introduced the change in the format, mostly as it makes the output from libvirt different from the output from other providers, and I honestly don't see the advantage. IMHO consistency should prevail over ad-hoc "optimizations" unless there's a good reason.
> 
> Sounds good to me, but wanted to check with @suseclee first to see why was this introduced, but meanwhile others can move forward if they are using libvirt by applying this patch :D
> 
> Im in favour of reverting the other patch instead, as at a glance it doesnt introduce any useful features.

I would prefer to use this output for all other platform that hostname matches ips.

pablochacin:
> I would rather to fix outputs for all other platforms.

In that case, I suggest to revert the initial commit and create a PR that updates the output for all platforms and in testrunner at once, so we have a consistent change. In this case, @Itxaka's  change should be geeneric for the terraform class (not specific for libvirt).

Itxaka:
@pablochacin so do we have an agreement here? Follow up patch to bring the other platform outputs to the same output as the libvirt? otherwise, please reject the patch and I will send the revert PR instead. Thanks!

pablochacin:
@Itxaka. I'm not convinced of the need for the changes, but as there seems to be an agreement on moving them forward, I agree on the proposed way of action. It would be great if we create an issue for the followup ticket and link to this one.

Itxaka:
issue created, merging this

Issue #904, opened 2019-12-18
Call zypper with `--userdata`


kkaempf:
All calls to `zypper` should pass a `--userdata` value.

https://github.com/SUSE/doc-caasp/issues/644#issuecomment-566900165 raised the question of "how can support detect that a customer did this [call `zypper` on the node manually] "wrong"?"

`zypper` has an option to pass a `transaction id`:
```
> zypper -h
  ...
    --userdata <STRING>     User defined transaction id used in history and plugins.
  ...
```
which will appear in the `zypper.log`. `skuba` could set this id to indicated "zypper was called from skuba" in the zypper.log.

jodavis-suse:
Interesting idea.  Would we want to use a generic and common "--userdata skuba" for all calls to zypper made by skuba, or would there be additional information that could be valuable as well?

Klaven:
just a comment. I like the idea of doing something like `--userdata skuba-[uid]` not sure what all your thoughts are on it. might be overkill/not needed or useful. was thinking if we had better logging it might help tie things together.

jodavis-suse:
As we talked about this, it would also make sense to have separate identifiers for skuba and skuba-update and any other modules within the code base.  There are also a fair number of templates and helper scripts under the ci/ which can be reviewed too, though instances for build and testing can be skipped.

dannysauer:
Unless we're going to have something associated with logs of the skuba / skuba-update execution itself (which the transaction ID could then be correlated against), I don't see much point in making this any sort of dynamic string. It only serves to makes "grep for all the things skuba did" and similar slightly more difficult, at that point. :)

If it's going to be something that can be tied back to other logs, on the other hand, it seems like the transaction ID should really be a "session ID" generated every time the zypper-calling command runs rather than just as part of the zypper command.

jodavis-suse:
I think I'll go for the 'simplest possible' approach for this and use static strings.  Expect a new PR commit soon.

jodavis-suse:
@dannysauer Both commits for this issue have now merged.  Is there a release this should get labeled for before I close the issue?

dannysauer:
I added the 4.1.2 label to both PRs and this ticket.  I think that's all which is needed.

Before marking this closed, should there perhaps also be a note somewhere about how this data shows up in the zypper logs?  Right now, it's just tribal knowledge that skuba-triggered zypper executions contain this additional information.  Might be worth opening a separate ticket (probably in avant-garde) to document this behavior change.

[...]

PR #913, opened 2020-01-15
skuba-update: include --userdata in calls to zypper

jodavis-suse:
To help differentiate calls to zypper from Skuba from calls made
by a user administrator, include a --userdata flag in the command.

## Why is this PR needed?

https://github.com/SUSE/skuba/issues/904

## What does this PR do?

This change does a minor refactor to the run_zypper_command in skuba-update.py
A further change will later be made to kubernetes.go to match.

A simple static userdata identifier is sufficient for now (log messages already contain
a timestamp to make them unique and come from a unique machine).

## Info for QA

This change has virtually no impact to a user, as it does not change any functional behavior.
The one marker for this change working is that when skuba-update is run, any commands
to zypper will now include the --userdata flag, so when zypper writes to its log at 
/var/log/zypper.log, lines similar to this will appear:

`2020-01-15 18:52:28 <1> test-worker-0(7476) [zypper] main.cc(main):98 ===== 'zypper' '--userdata' 'skuba-update' 'ref' =====`

`2020-01-15 18:52:28 <1> test-worker-0(7476) [zconfig] ZConfig.cc(setUserData):908 Set user data string to 'skuba-update'`

This userdata may also be visible in the system journal when the zypper command is logged.

## Docs

No known docs impact.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/skuba-update/skuba_update/skuba_update.py
+++ b/skuba-update/skuba_update/skuba_update.py
@@ -75,7 +75,7 @@ def main():
     if os.geteuid() != 0:
         raise Exception('root privileges are required to run this tool')
 
-    run_zypper_command(['zypper', 'ref', '-s'])
+    run_zypper_command(['ref', '-s'])
     if not args.annotate_only:
         code = update()
         restart_services()
@@ -142,7 +142,7 @@ def annotate_updates_available(node_name):
     """
 
     patch_xml = run_zypper_command(
-        ['zypper', '--non-interactive', '--xmlout', 'list-patches'],
+        ['--non-interactive', '--xmlout', 'list-patches'],
         needsOutput=True
     ).output
     updates = get_update_list(patch_xml)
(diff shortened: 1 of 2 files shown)


skuba-update/skuba_update/skuba_update.py
```diff
@@ -289,19 +290,22 @@ def run_zypper_command(command, needsOutput=False):
     Run the given zypper command. The command is expected to be a tuple which
     also contains the 'zypper' string. It returns the exit code from zypper.
```
jodavis-suse: Missed changing this before making the PR.  Refactored code no longer requires including 'zypper', which makes the calls a little shorter.

skuba-update/skuba_update/skuba_update.py
```diff
@@ -289,19 +290,22 @@ def run_zypper_command(command, needsOutput=False):
...
     """
+    # updateProcessId = 'skuba-update-{0}'.format(os.getpid())
+    uniqueId = 'skuba-update-{0}'.format(uuid.uuid4())
+    if command[0] != 'zypper':
```
jodavis-suse: Will drop this if statement before final commit.

dannysauer: Why random rather than a hostname+timestamp `uuid1()`?  Any particular reason?

jodavis-suse:
Got in a bit of a fight with github - will need to squash the commits when this is merged.

PR #918, opened 2020-01-17
Isolate libvirt deployment disk size by role

c3y1huang:
## Why is this PR needed?

Currently libvirt deployment restrict to same disk size for all instance role. This might cause over resource reservation on host, as different cluster role have different requirement for disk size.

## What does this PR do?

This isolates libvirt deployment disk size by role of lb, master, worker.

Unable to configure libvirt deployment disk size according to role.

Able to configure libvirt deployment disk size according to role.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

Signed-off-by: Chin-Ya Huang <chin-ya.huang@suse.com>

### DIFF
--- a/ci/infra/libvirt/lb-instances.tf
+++ b/ci/infra/libvirt/lb-instances.tf
@@ -63,7 +63,7 @@ data "template_file" "lb_cloud_init_userdata" {
 resource "libvirt_volume" "lb" {
   name           = "${var.stack_name}-lb-volume"
   pool           = "${var.pool}"
-  size           = "${var.disk_size}"
+  size           = "${var.lb_disk_size}"
   base_volume_id = "${libvirt_volume.img.id}"
 }
 
--- a/ci/infra/libvirt/master-instance.tf
+++ b/ci/infra/libvirt/master-instance.tf
@@ -53,7 +53,7 @@ data "template_file" "master-cloud-init" {
 resource "libvirt_volume" "master" {
   name           = "${var.stack_name}-master-volume-${count.index}"
   pool           = "${var.pool}"
-  size           = "${var.disk_size}"
+  size           = "${var.master_disk_size}"
   base_volume_id = "${libvirt_volume.img.id}"
   count          = "${var.masters}"
 }
--- a/ci/infra/libvirt/variables.tf
+++ b/ci/infra/libvirt/variables.tf
@@ -72,11 +72,6 @@ variable "rmt_server_name" {
   description = "SUSE Repository Mirroring Server Name"
 }
 
-variable "disk_size" {
-  default     = "25769803776"
-  description = "Disk size (in bytes)"
-}
-
 variable "dns_domain" {
   type        = "string"
   default     = "caasp.local"
@@ -110,6 +105,11 @@ variable "lb_vcpu" {
   description = "Amount of virtual CPUs for a load balancer node"
 }
 
+variable "lb_disk_size" {
+  default     = "25769803776"
+  description = "Disk size (in bytes)"
+}
+
 variable "lb_repositories" {
   type = "map"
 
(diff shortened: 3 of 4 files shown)


c3y1huang:
> This changes the user interface and user experience, do we really want to do this during the upgrade of 0.12? I am not in favor, but I can be convinced otherwise.

IMHO this would give user more control over the disk resource used for CaaSP cluster. 
Workers would need more disk space than loadbalancer. Imagine user want to have 1T on workers, the provision will then provision same disk for loadbalancer and masters.

evrardjp:
Yeah I understood what you are trying to achieve. But I am pondering if a change in the user interface should happen now or not.

c3y1huang:
> Yeah I understood what you are trying to achieve. But I am pondering if a change in the user interface should happen now or not.

I don't have strong opinion on that. When do you suggest this should happen?

evrardjp:
We discussed this in our x team standup.

We would prefer:
- Postpone this until the tf 0.12 upgrade is merged, tested by QA, and released (to have releases focus on the minimum things). This could currently indeed delay the delivery of the tf 0.12 upgrade.
- Discussing about the value of this PR as a _libvirt only_ change: If this change is interesting, I suppose it should best be done for all the providers, and therefore change the user interface for all the providers in a consistent way.
- Have a QA sign-off on this, as it effectively changes the user interface (whether it's not a libvirt-only change, or if QA is using this for their testing)

c3y1huang:
> We discussed this in our x team standup.
> 
> We would prefer:
> 
> * Postpone this until the tf 0.12 upgrade is merged, tested by QA, and released (to have releases focus on the minimum things). This could currently indeed delay the delivery of the tf 0.12 upgrade.
> * Discussing about the value of this PR as a _libvirt only_ change: If this change is interesting, I suppose it should best be done for all the providers, and therefore change the user interface for all the providers in a consistent way.
> * Have a QA sign-off on this, as it effectively changes the user interface (whether it's not a libvirt-only change, or if QA is using this for their testing)

Fair suggestion, thanks. Let’s postpone this.

c3y1huang:
Ref: https://github.com/SUSE/avant-garde/issues/1114

c3y1huang:
> * Postpone this until the tf 0.12 upgrade is merged, tested by QA, and released (to have releases focus on the minimum things). This could currently indeed delay the delivery of the tf 0.12 upgrade.
> * Discussing about the value of this PR as a _libvirt only_ change: If this change is interesting, I suppose it should best be done for all the providers, and therefore change the user interface for all the providers in a consistent way.
> * Have a QA sign-off on this, as it effectively changes the user interface (whether it's not a libvirt-only change, or if QA is using this for their testing)

On the second look, I see currently there is only `libvirt` has not yet applied to the role-base-disk-size.
```
Following is the current implementation of different infa providers:
* AWS: already implemented with: `var.master_size`, `var.worker_size`.
* Bare-metal: autoyast does not control resource provision.
* Libvirt: Not implemented, currently use `var.disk_size`.
* Openstack: already implemented with `var.master_size`, `var.worker_size`.
* Vmware: already implemented with: `var.lb_disk_size`, `var.master_disk_size`, `var.worker_disk_size`.
```
I would say we can still postpone this after 0.12 and to have QA signed off on this. But I don't really see the purpose of discussing because all other provider has already adapt to it. What do you think?

evrardjp:
As said earlier, it seems fair to do this kind of change. I just think it's a wrong timing.

I would still prefer to wait for 0.12, as it impacts delivery time (QA would have to re-test again).

I will ask QA at our standup. It makes sense that every user of the vmware platform be informed of this, and for me that's QA and CAP.

c3y1huang:
> As said earlier, it seems fair to do this kind of change. I just think it's a wrong timing.
> 
> For me, it makes sense that every user of the vmware platform be informed of this, and for me that's QA and CAP.
> 
> I would still prefer to wait for 0.12, as it impacts delivery time (QA would have to re-test again).
> 
> I will ask QA at our standup.

c3y1huang:
ohoh, I mistakenly closed this. Will open another one.

PR #921, opened 2020-01-17
Improve log level conventions

jodavis-suse:
Referencing the upstream Kubernetes Instrumentation
Logging Conventions, move some of the log messages
in Skuba to align with described levels.

[0] https://github.com/kubernetes/community/blob/master/contributors
/devel/sig-instrumentation/logging.md

## Why is this PR needed?

https://github.com/SUSE/avant-garde/issues/1132

## What does this PR do?

The upstream Kubernetes Instrumentation SIG has published a convention for setting logging levels.  This PR attempts to get skuba code in line with those conventions (with values in the 0-5 range) by reviewing the existing klog.V() calls (which currently are all klog.V(1)) and setting levels for the messages that align with the convention.

There should be no significant change to user experience, though some messages that are currently logged with a 1 will now require a higher log level to trigger.

## Info for QA

Running skuba with -v [2-5] may not produce more log output than just -v 1, depending on the command and code path invoked.

Effectlively, all Info messages from skuba internal code were logged with a 1 level.

Running skuba with -v [2-5] may not produce more log output than just -v 1, depending on the command and code path invoked.

## Docs

No documentation impact due to this change - current documentation about log levels is still correct.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/ssh.go
+++ b/internal/pkg/skuba/deployments/ssh/ssh.go
@@ -165,7 +165,7 @@ func (t *Target) internalSshWithStdin(silent bool, stdin string, command string,
 		finalCommand = fmt.Sprintf("sudo sh -c '%s'", finalCommand)
 	}
 	if !silent {
-		klog.V(1).Infof("running command: %q", finalCommand)
+		klog.V(2).Infof("running command: %q", finalCommand)
 	}
 	if err := session.Start(finalCommand); err != nil {
 		return "", "", err
@@ -188,7 +188,7 @@ func readerStreamer(reader io.Reader, outputChan chan<- string, description stri
 	for scanner.Scan() {
 		result.Write([]byte(scanner.Text()))
 		if description == "stdout" && !silent {
-			klog.V(1).Infof("%s", scanner.Text())
+			klog.V(2).Infof("%s", scanner.Text())
 		} else if description == "stderr" {
 			klog.Errorf("%s", scanner.Text())
 		}
--- a/internal/pkg/skuba/kured/kured.go
+++ b/internal/pkg/skuba/kured/kured.go
@@ -44,7 +44,7 @@ func Lock(client clientset.Interface) error {
 	if err != nil {
 		return errors.Wrap(err, "unable to patch daemonset with kured locking annotation")
 	}
-	klog.V(1).Info("successfully annotated daemonset with kured locking annotation")
+	klog.V(2).Info("successfully annotated daemonset with kured locking annotation")
 	return nil
 }
 
@@ -56,6 +56,6 @@ func Unlock(client clientset.Interface) error {
 	if err != nil {
 		return errors.Wrap(err, "unable to patch daemonset with kured unlocking annotation")
 	}
-	klog.V(1).Info("successfully removed kured locking annotation")
+	klog.V(2).Info("successfully removed kured locking annotation")
 	return nil
 }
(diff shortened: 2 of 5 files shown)


internal/pkg/skuba/deployments/ssh/ssh.go
```diff
@@ -165,7 +165,7 @@ func (t *Target) internalSshWithStdin(silent bool, stdin string, command string,
 		finalCommand = fmt.Sprintf("sudo sh -c '%s'", finalCommand)
 	}
 	if !silent {
-		klog.V(1).Infof("running command: %q", finalCommand)
+		klog.V(2).Infof("running command: %q", finalCommand)
```
jodavis-suse: Actually, these log messages in ssh.go make me nervous, as I don't see any sort of security check in place that would prevent any sensitive information used in the ssh command from being logged.  I am hoping others with more experience in skuba can comment.

internal/pkg/skuba/deployments/ssh/ssh.go
```diff
@@ -188,7 +188,7 @@ func readerStreamer(reader io.Reader, outputChan chan<- string, description stri
 	for scanner.Scan() {
 		result.Write([]byte(scanner.Text()))
 		if description == "stdout" && !silent {
-			klog.V(1).Infof("%s", scanner.Text())
+			klog.V(2).Infof("%s", scanner.Text())
```
jodavis-suse: Again, slight security paranoia flag here as we are logging all the output from the ssh command.

internal/pkg/skuba/kured/kured.go
```diff
@@ -44,7 +44,7 @@ func Lock(client clientset.Interface) error {
 	if err != nil {
 		return errors.Wrap(err, "unable to patch daemonset with kured locking annotation")
 	}
-	klog.V(1).Info("successfully annotated daemonset with kured locking annotation")
+	klog.V(2).Info("successfully annotated daemonset with kured locking annotation")
```
jodavis-suse: I wasn't clear on the frequency or importance of the log messages in kured.go.  If they are rarely logged and of high importance, then leaving them at V(1) may be appropriate.

internal/pkg/skuba/deployments/ssh/ssh.go
suseclee: I believe, most of logs comes from this line. Good to have differentiate log level.
I was confusing v(2) and V(3). I read doc again. klog.V(2) - Useful steady state information about the service and important log messages that may correlate to significant changes in the system. I think this satisfies V(2).

jodavis-suse:
Rebased to resolve a merge conflict.  Hope that resolves any check failures too.

internal/pkg/skuba/deployments/ssh/ssh.go
Klaven: there is a "silent" ssh. this is the one we use for secure stuff that does not log anything.

internal/pkg/skuba/kured/kured.go
Klaven: IMO 2 is fine here.

internal/pkg/skuba/deployments/ssh/ssh.go
jodavis-suse: Good point, and good to know where we place the trust that the command is not run incorrectly.

PR #925, opened 2020-01-22
PoC: Rewrite kubectl functionality with client-go

cmurphy:
Instead of shelling out to kubectl, parse each resource in the addon
manifest and apply them using client-go. Use server-side apply[1] by
setting a field manager, forcing conflicts, and using the
"application/apply-patch+yaml" patch type.

[1] https://kubernetes.io/docs/reference/using-api/api-concepts/#server-side-apply

## Why is this PR *not* needed?

This was a useful exercise but I do not think it should merge. It reimplements a chunk of kubectl for no real benefit. Server-side apply does not get us very much, its usefulness is for keeping resources in a consistent state when multiple clients are working on them. Trying to write an apply function from client-go primitives instead of just relying on kubectl has a high risk of introducing new bugs.

Fixes https://github.com/SUSE/avant-garde/issues/854

### DIFF
--- a/internal/pkg/skuba/addons/addons.go
+++ b/internal/pkg/skuba/addons/addons.go
@@ -20,16 +20,25 @@ package addons
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"sort"
 	"text/template"
 
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/version"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/dynamic"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


cmurphy:
@ereslibre (as the creator of https://github.com/SUSE/avant-garde/issues/854) after working through this proof of concept I think it's not worthwhile to pursue this refactor for the reasons given in the PR description. What do you think? I'd be willing to discuss it more.

aojea:
well, it seems to remove the `kubectl` dependency on the cluster nodes, right?

internal/pkg/skuba/addons/addons.go
```diff
@@ -252,11 +261,61 @@ func (addon Addon) Apply(client clientset.Interface, addonConfiguration AddonCon
...
+		if err != nil {
+			return err
+		}
+		dynamicClient, err := dynamic.NewForConfig(clientCfg)
+		if err != nil {
+			return err
+		}
+		_, err = dynamicClient.Resource(mapping.Resource).
+			Namespace(namespace).
+			Patch(name, "application/apply-patch+yaml", data, *options)
```
ereslibre:
```suggestion
			Patch(name, types.ApplyPatchType, data, *options)
```

where package `types` is `k8s.io/apimachinery/pkg/types`.

cmurphy:
> Thank you for checking @cmurphy!, yes indeed, this looks a bit cumbersome, maybe it's safer to keep relying on `kubectl` for now and revisit this at some later time? WDYT?

@ereslibre I'm not sure this needs to be revisited, I think it could just be closed. I don't think client-go is ever going to have enough functionality to make it easy to parse and apply a manifest on its own.

ereslibre:
Thank you @cmurphy, let's close it then.

Issue #919, opened 2020-01-17
Disable `firewalld`


ereslibre:
Skuba should disable `firewalld` when bootstrapping/joining a node. Currently, this is done by `cloud-init` in certain deployments with `terraform`.

Being this a deployment trait, `skuba` should make sure that this service is disabled, so it doesn't cause any unexpected issues.

evrardjp:
Not sure what's the intent of this issue, as it lacks explanations on _why_ we want to do this.

As firewalld can handle bridges and NAT, I don't think disabling the service is the right idea.
Instead we should probably change rules to allow everything in runtime, and restart the firewalld service when we are done, to re-use the rules that are stored in the permanent configuration.

https://firewalld.org/documentation/configuration/runtime-versus-permanent.html

ereslibre:
We are disabling `firewalld` with terraform for `libvirt` and `vmware`, and the JeOS image used for `openstack` don't have `firewalld` at all.

This imposes a hidden dependency on `terraform`, assuming from `skuba` side that `terraform` will disable this service. However, customers can use whatever they prefer to create the infrastructure, they can rely on `terraform` if they don't have anything else to create the infra.

In this case, `skuba` should make sure that `firewalld` is disabled so we get rid of this hidden dependency.

evrardjp:
But what's the problem that was the reason of disabling firewalld with libvirt and vmware in the first place?

I am worried about disabling firewalld for two reasons:
- either we then have two different code paths for customers and our CI for this, which isn't great
- or we disable firewalld in a customer environment, which doesn't sound so secure.

I understand that the hidden dependency shouldn't be there. But I also am not sure whether disabling the firewall altogether is the right way.

ereslibre:
> But what's the problem that was the reason of disabling firewalld with libvirt and vmware in the first place?

It is preventing the apiserver from being reachable. Also, it makes all platforms even, because JeOS (openstack image) does not have `firewalld`.

> I am worried about disabling firewalld for two reasons:

This is the case already, `firewalld` either is not installed at all (JeOS), or disabled (libvirt, vmware).

This issue just proposes to remove this logic from terraform and implement it on `skuba`, so if a user deploys the infrastructure with their own tooling, `skuba` will take care of making sure that everything is as expected.

I understand your concern, but enabling firewalld and making sure that all rules exist and are as expected would be a different task and we should treat it as a new item in our backlog, since it's a new security feature.

evrardjp:
[...]

PR #927, opened 2020-01-22
Disable firewalld

chentex:
## Why is this PR needed?

Skuba should disable `firewalld` when bootstrapping/joining a node. Currently, this is done by `cloud-init` in certain deployments with `terraform`.

Being this a deployment trait, `skuba` should make sure that this service is disabled, so it doesn't cause any unexpected issues.

Fixes: https://github.com/SUSE/skuba/issues/919

## What does this PR do?

- adds a startup step to check that `firewald` is disabled, when bootstraping and joining nodes.

## Info for QA

When checking for `firewalld` service it should state that it is disabled.

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/firewalld.go
+++ b/internal/pkg/skuba/deployments/ssh/firewalld.go
@@ -26,7 +26,7 @@ func init() {
 }
 
 func firewalldDisable(t *Target, data interface{}) error {
-	_, _, err := t.ssh("systemctl", "list-units", "--type=service", "firewalld", "|", "grep", "firewalld")
+	_, _, err := t.ssh("systemctl", "cat", "firewalld")
 	if err == nil {
 		_, _, err := t.ssh("systemctl", "disable", "--now", "firewalld")
 		return err
--- a/internal/pkg/skuba/deployments/ssh/firewalld.go
+++ b/internal/pkg/skuba/deployments/ssh/firewalld.go
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2019 SUSE LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package ssh
+
+func init() {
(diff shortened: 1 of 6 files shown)


internal/pkg/skuba/deployments/ssh/firewalld.go
```diff
@@ -0,0 +1,38 @@
...
+	stateMap["firewalld.disable"] = firewalldDisable
+}
+
+func firewalldDisable(t *Target, data interface{}) error {
+	_, _, err := t.ssh("systemctl", "list-units", "--type=service", "firewalld", "|", "grep", "firewalld")
+	if err == nil {
+		_, _, err := t.ssh("systemctl", "disable", "--now", "firewalld")
+		return err
+	}
+	if err != nil {
```
jenting: This if is not needed.

internal/pkg/skuba/deployments/ssh/firewalld.go
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2019 SUSE LLC.
```
jenting: 2020

internal/pkg/skuba/deployments/ssh/firewalld.go
chentex: good catch!

chentex:
> Would the firewalld re-enable after bootstrap/join?

Quoting @ereslibre 

> It is preventing the `apiserver` from being reachable.

So I am assuming that if we re-enable it again this issue will surface again.

Maybe Rafa can correct mew if I;m wrong, and we should re-enable it after.

internal/pkg/skuba/deployments/ssh/firewalld.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"k8s.io/klog"
+)
+
+func init() {
+	stateMap["firewalld.disable"] = firewalldDisable
+}
+
+func firewalldDisable(t *Target, data interface{}) error {
+	_, _, err := t.ssh("systemctl", "list-units", "--type=service", "firewalld", "|", "grep", "firewalld")
```
ereslibre: Instead of grepping here you could do:

```suggestion
	_, _, err := t.ssh("systemctl", "cat", "firewalld")
```

chentex: That is a neat trick.

chentex: Done with the changes.

aojea:
@jordimassaguerpla @chentex can we add a release note mentioning this change?
I think that the description of the PR is pretty accurate 
/cc @evrardjp 


`

jordimassaguerpla:
Sure. Please open a PR to https://github.com/SUSE/caasp-release-notes

PR #928, opened 2020-01-22
Show node's role in 'skuba cluster status'

suseclee:
## Why is this PR needed?

Does it fix an issue? addresses a business case?skuba cluster status should show if a node is a master or worker (Daimler and T-Systems requested this).

## What does this PR do?

Sample output:
```
 skuba cluster status
NAME       ROLE     OS-IMAGE                              KERNEL-VERSION           KUBELET-VERSION   CONTAINER-RUNTIME   HAS-UPDATES   HAS-DISRUPTIVE-UPDATES   CAASP-RELEASE-VERSION
master-0   master   SUSE Linux Enterprise Server 15 SP1   4.12.14-197.29-default   v1.16.2           cri-o://1.16.1      yes           no                       <none>
master-1   master   SUSE Linux Enterprise Server 15 SP1   4.12.14-197.29-default   v1.16.2           cri-o://1.16.1      <none>        <none>                   <none>
worker-0   worker   SUSE Linux Enterprise Server 15 SP1   4.12.14-197.29-default   v1.16.2           cri-o://1.16.1      yes           no                       <none>
worker-1   worker   SUSE Linux Enterprise Server 15 SP1   4.12.14-197.29-default   v1.16.2           cri-o://1.16.1      yes           no                       <none>
```

This is the indication of master node from `kubectl get node -o json`.
"labels": {
                    "beta.kubernetes.io/arch": "amd64",
                    "beta.kubernetes.io/os": "linux",
                    "kubernetes.io/arch": "amd64",
                    "kubernetes.io/hostname": "master-0",
                    "kubernetes.io/os": "linux",
                    "node-role.kubernetes.io/master": ""
                },
In the case of worker nodes, there is no label.

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/pkg/skuba/actions/cluster/status/status.go
+++ b/pkg/skuba/actions/cluster/status/status.go
@@ -19,6 +19,7 @@ package cluster
 
 import (
 	"os"
+	"strings"
 
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -34,7 +35,24 @@ func Status(client clientset.Interface) error {
 		return errors.Wrap(err, "could not retrieve node list")
 	}
 
-	outputFormat := "custom-columns=NAME:.metadata.name,OS-IMAGE:.status.nodeInfo.osImage,KERNEL-VERSION:.status.nodeInfo.kernelVersion,KUBELET-VERSION:.status.nodeInfo.kubeletVersion,CONTAINER-RUNTIME:.status.nodeInfo.containerRuntimeVersion,HAS-UPDATES:.metadata.annotations.caasp\\.suse\\.com/has-updates,HAS-DISRUPTIVE-UPDATES:.metadata.annotations.caasp\\.suse\\.com/has-disruptive-updates,CAASP-RELEASE-VERSION:.metadata.annotations.caasp\\.suse\\.com/caasp-release-version"
+	for _, node := range nodeList.Items {
+		for label := range node.Labels {
+			if strings.Contains(label, "node-role.kubernetes.io") && len(strings.Split(label, "/")) > 0 {
+				node.Labels["caasp-role.kubernetes.io"] = strings.Split(label, "/")[1]
+			}
+		}
+	}
+
+	outputFormat := "custom-columns=" +
+		"NAME:.metadata.name," +
+		"ROLE:.metadata.labels.caasp-role\\.kubernetes\\.io," +
+		"OS-IMAGE:.status.nodeInfo.osImage," +
+		"KERNEL-VERSION:.status.nodeInfo.kernelVersion," +
+		"KUBELET-VERSION:.status.nodeInfo.kubeletVersion," +
+		"CONTAINER-RUNTIME:.status.nodeInfo.containerRuntimeVersion," +
+		"HAS-UPDATES:.metadata.annotations.caasp\\.suse\\.com/has-updates," +
+		"HAS-DISRUPTIVE-UPDATES:.metadata.annotations.caasp\\.suse\\.com/has-disruptive-updates," +
+		"CAASP-RELEASE-VERSION:.metadata.annotations.caasp\\.suse\\.com/caasp-release-version"
 
 	printFlags := kubectlget.NewGetPrintFlags()
 	printFlags.OutputFormat = &outputFormat


pkg/skuba/actions/cluster/status/status.go
```diff
@@ -34,7 +34,24 @@ func Status(client clientset.Interface) error {
 		return errors.Wrap(err, "could not retrieve node list")
 	}
 
-	outputFormat := "custom-columns=NAME:.metadata.name,OS-IMAGE:.status.nodeInfo.osImage,KERNEL-VERSION:.status.nodeInfo.kernelVersion,KUBELET-VERSION:.status.nodeInfo.kubeletVersion,CONTAINER-RUNTIME:.status.nodeInfo.containerRuntimeVersion,HAS-UPDATES:.metadata.annotations.caasp\\.suse\\.com/has-updates,HAS-DISRUPTIVE-UPDATES:.metadata.annotations.caasp\\.suse\\.com/has-disruptive-updates,CAASP-RELEASE-VERSION:.metadata.annotations.caasp\\.suse\\.com/caasp-release-version"
+	for _, item := range nodeList.Items {
+		if _, ok := item.Labels["node-role.kubernetes.io/master"]; ok {
```
aojea: with this logic it assumes that whatever node that is not labeled as `node-role.kubernetes.io/master"` is a worker, is this a safe assumption? can be nodes that are not labeled?

suseclee: Good question.
The answer would be:
```
kubectl get nodes
NAME       STATUS   ROLES    AGE   VERSION
master-0   Ready    master   18h   v1.16.2
worker-0   Ready    <none>   18h   v1.16.2
worker-1   Ready    <none>   18h   v1.16.2
```
You can see what get role here with `kubectl get nodes -o json`

Klaven: :thinking: If I understand https://github.com/kubernetes/community/pull/911 right, then this should be a find assumption and one we can update later. I'm not worried that this will have much of a lasting hard to fix problem. if there becomes another node role we want to care about, we can update it then.

dannysauer:
> Personally the entire community uses `control-plane` and not `master` anymore.

Without getting into a discussion of the merits of words, I agree that getting rid of `slave` but keeping `master` seems like a half measure in terms of making sure no one is offended.

However, considering how important consistency is, making this change here means we have a bunch of documentation that should be updated, and also means we should change the parameter passed to role (or at least accept both monikers).  Are we ready to invest the resources to suppress the `master` term?  I'm inclined to suggest doing that change to the whole codebase in one separate PR.

pkg/skuba/actions/cluster/status/status.go
dannysauer: The answer here depends on whether we want to reflect "what did skuba say" or "what is the node role".

Skuba only allows two roles during bootstrap, but that's misleading, IMO.  Nodes can have arbitrary roles.  The `master` role is treated specially when scheduling jobs, but a customer site can put an arbitrary value in the node using the label format `node-role.kubernetes.io/<role>`.  The "right" behavior if we're trying to stay close to upstream is to use the kubectl output format as referenced in https://github.com/SUSE/skuba/pull/928#discussion_r370202922.  We should therefore emit `<none>` if no role is assigned, otherwise emit the contents of that label.  Documentation can explain that worker nodes default to `<none>` unless a site-specific role is assigned to that node.

Klaven:
> > Personally the entire community uses `control-plane` and not `master` anymore.
> 
> Without getting into a discussion of the merits of words, I agree that getting rid of `slave` but keeping `master` seems like a half measure in terms of making sure no one is offended.
> 
> However, considering how important consistency is, making this change here means we have a bunch of documentation that should be updated, and also means we should change the parameter passed to role (or at least accept both monikers). Are we ready to invest the resources to suppress the `master` term? I'm inclined to suggest doing that change to the whole codebase in one separate PR.

so the role that is assigned as a label will stay `master` there was a long talk on it. the only thing upstream changed was how they talked about it... aka, the control plane is a documentation/conversation. label == master. I am on board for leaving this `master` as it is consistent with upstream `kubectl` as well as what the label actually is. I do think our documentation should change to match upstream documentation.

so in the end. i'm fine with `master` and `<none>` just like `kubectl`

suseclee:
Th eoutput sameple will be:
```
skuba cluster status
NAME       ROLE     OS-IMAGE                              KERNEL-VERSION           KUBELET-VERSION   CONTAINER-RUNTIME   HAS-UPDATES   HAS-DISRUPTIVE-UPDATES   CAASP-RELEASE-VERSION
master-0   master   SUSE Linux Enterprise Server 15 SP1   4.12.14-197.29-default   v1.16.2           cri-o://1.16.1      yes           no                       <none>
worker-0   <none>   SUSE Linux Enterprise Server 15 SP1   4.12.14-197.29-default   v1.16.2           cri-o://1.16.1      yes           no                       <none>
worker-1   <none>   SUSE Linux Enterprise Server 15 SP1   4.12.14-197.29-default   v1.16.2           cri-o://1.16.1      yes           no                       <none>
worker-2   <none>   SUSE Linux Enterprise Server 15 SP1   4.12.14-197.29-default   v1.16.2           cri-o://1.16.1      yes           no                       <none>

```

pkg/skuba/actions/cluster/status/status.go
```diff
@@ -34,7 +35,25 @@ func Status(client clientset.Interface) error {
 		return errors.Wrap(err, "could not retrieve node list")
 	}
 
-	outputFormat := "custom-columns=NAME:.metadata.name,OS-IMAGE:.status.nodeInfo.osImage,KERNEL-VERSION:.status.nodeInfo.kernelVersion,KUBELET-VERSION:.status.nodeInfo.kubeletVersion,CONTAINER-RUNTIME:.status.nodeInfo.containerRuntimeVersion,HAS-UPDATES:.metadata.annotations.caasp\\.suse\\.com/has-updates,HAS-DISRUPTIVE-UPDATES:.metadata.annotations.caasp\\.suse\\.com/has-disruptive-updates,CAASP-RELEASE-VERSION:.metadata.annotations.caasp\\.suse\\.com/caasp-release-version"
+	for _, node := range nodeList.Items {
+		for label, value := range node.Labels {
+			_ = value
```
suseclee: lint error for `for label, _ := range node.Labels {`. I added  `_ = value`.

Klaven: I don't think I understand what is going on here.... but this cant be the solution.

pkg/skuba/actions/cluster/status/status.go
```diff
@@ -34,7 +35,25 @@ func Status(client clientset.Interface) error {
...
+			if strings.Contains(label, "node-role.kubernetes.io") && len(strings.Split(label, "/")) > 0 {
+				node.Labels["caasp-role.kubernetes.io"] = strings.Split(label, "/")[1]
```
Klaven: :thinking: hmmmmmmmm will this give the same output as kubectl? aka, `<none>` for "worker" nodes or do we need an `else` statement?

Klaven: never mind. i'm a dumb dumb.

suseclee: if you set  custom label, "kubectl label nodes worker-1 node-role.kubernetes.io/worker=", you will get 
```
skuba cluster status
NAME       ROLE     OS-IMAGE                              KERNEL-VERSION           KUBELET-VERSION   CONTAINER-RUNTIME   HAS-UPDATES   HAS-DISRUPTIVE-UPDATES   CAASP-RELEASE-VERSION
master-0   master   SUSE Linux Enterprise Server 15 SP1   4.12.14-197.29-default   v1.16.2           cri-o://1.16.1      yes           no                       <none>
worker-0   worker   SUSE Linux Enterprise Server 15 SP1   4.12.14-197.29-default   v1.16.2           cri-o://1.16.1      yes           no                       <none>
worker-1   <none>   SUSE Linux Enterprise Server 15 SP1   4.12.14-197.29-default   v1.16.2           cri-o://1.16.1      yes           no                       <none>
worker-2   <none>   SUSE Linux Enterprise Server 15 SP1   4.12.14-197.29-default   v1.16.2           cri-o://1.16.1      yes           no                       <none>

kubectl get nodes
NAME       STATUS                     ROLES    AGE     VERSION
master-0   Ready                      master   4d20h   v1.16.2
worker-0   Ready                      worker   4d20h   v1.16.2
worker-1   Ready                      <none>   4d20h   v1.16.2
worker-2   Ready                      <none>   2d16h   v1.16.2
```

suseclee: >  hmmmmmmmm will this give the same output as kubectl? aka, `<none>` for "worker" nodes or do we need an `else` statement?

If there is no label of caasp-role.kubernetes.io, the role will be shown as None. However, what ever roles are assigned, it will be shown like `kubectl get node`.

pkg/skuba/actions/cluster/status/status.go
cmurphy: Should be `for label := range nodeList.Items {` without the second value

```
$ golint pkg/skuba/actions/cluster/status/status.go 
pkg/skuba/actions/cluster/status/status.go:39:14: should omit 2nd value from range; this loop is equivalent to `for label := range ...`
```

dannysauer: The outer loop iterates over `range nodeList.Items`. How would that also work on the inner loop?

dannysauer: Oh, did you mean `for label := range node.Labels {`, @cmurphy? Does Go just silently discard later elements if the LHS only has one variable name provided?

dannysauer: Apparently.  https://stackoverflow.com/a/49193047/65589

Ugh.  Go is horrible. :D

cmurphy: Sorry yes I meant `range node.Labels`

suseclee: Thanks Calleen

jordimassaguerpla:
Do we have a bsc number for this?

suseclee:
> Do we have a bsc number for this?

Sorry for being late. We had a holiday on Monday. The number is #1140530: https://bugzilla.suse.com/show_bug.cgi?id=1140530

jordimassaguerpla:
Thanks. I see this is refered in https://github.com/SUSE/avant-garde/issues/545 which is already closed and in staging. All matches now. :+1:

PR #936, opened 2020-01-28
Added ablilty to set env var for pip verbosity

tdaines42:
## Why is this PR needed?

Fixes https://github.com/SUSE/avant-garde/issues/901

## What does this PR do?

Adds the ability to set an env var to control the verbosity of the pip install the testrunner shell script does.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/testrunner
+++ b/ci/infra/testrunner/testrunner
@@ -7,6 +7,7 @@ echo "Starting $0 script"
 SDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 VENVDIR=${WORKSPACE:-~}/py3venv
 export VENVDIR=${VENVDIR}
+PIP_VERBOSE=${PIP_VERBOSE:-"false"}
 
 setup_python_env() {
   type python3 >/dev/null 2>&1 || sudo zypper install --no-confirm python3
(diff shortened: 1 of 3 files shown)


ci/infra/testrunner/testrunner
```diff
@@ -18,7 +19,12 @@ setup_python_env() {
   source ${VENVDIR}/bin/activate
   # Not calling to `pip3` directly because its shebang can exceed 128 characters
   # and this operation will fail. Check: https://github.com/pypa/pip/issues/1773
-  python -m pip install -r "${SDIR}/requirements.txt" &>/dev/null
+  if [[ $PIP_VERBOSE == "true" ]]; then
+    python -m pip install -r "${SDIR}/requirements.txt"
+  else
+    python -m pip install -r "${SDIR}/requirements.txt" &>/dev/null
```
dannysauer: I'm not personally a huge fan of this bash-specific redirection, but if we can always assume this will be bash (as opposed to `/bin/sh`), it's fine. :)

tdaines42:
> Looks like it's working. Might be nice to have it documented in the README though.

Hmm, yeah I looked at adding it but the readme seems to only be about testrunner proper.

PR #954, opened 2020-02-13
Terraform OpenStack: create deployment script

flavio:
## Why is this PR needed?

We have users who want to automate the deployment of a CaaS Platform cluster on top of the infrastructure that is created by terraform on OpenStack.

They basically want a simple bash script that can be invoked inside of a CI pipeline.

## What does this PR do?

Previous to this commit the terraform deployment for OpenStack wrote a deployment snippet to the stdout during the `apply` phase.
With this commit the quick deployment snippet is instead written to a file.

That allows users (or CI systems) to quickly provision a CaaS Platform cluster on top of the freshly created infrastructure.

**Note well:** as prior to this commit, the snippet is created only when the terraform `cpi_enable` variable is set to `true`.

Nothing special.

## Info for QA

Nothing really special to be done.

## Docs

Maybe we could extend the documentation to inform the users about this new script. However the output of terraform explicitly mention it.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/openstack/generate-cpi-conf.sh
+++ b/ci/infra/openstack/generate-cpi-conf.sh
@@ -1,9 +1,11 @@
 #!/bin/bash
 #shellcheck disable=SC2145,SC2016
 log()   { (>&1 echo -e "$@") ; }
-cmd()   { log "[ CMD ] $@" ; }
 info()  { log "[ INFO ] $@" ; }
 error() { (>&2 echo -e "[ ERROR ] $@") ;}
+append_deployment_cmd()   {
+  echo "$@" >> "${SKUBA_DEPLOYMENT_SCRIPT}"
+}
 
 if [ -z "${OS_AUTH_URL}" ] || [ -z "${OS_USERNAME}" ] || \
    [ -z "${OS_PASSWORD}" ] || [ -z "${OS_PROJECT_ID}" ] || \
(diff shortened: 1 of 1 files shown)


SheilaghM:
Would this help
https://confluence.suse.com/display/~mcounts/Skuba+--config+RFC
You might also want to take a look at Ty's Diving Bell project

Klaven:
Thanks for mentioning my old RFC, at this point I honestly like @tdaines42 diving bell better then my original rfc.

it can be found here https://github.com/tdaines42/diving-bell

It is a cool separation of concerns. My RFC included this functionality as part of skuba, but with diving bell you can get the separation of concerns nicely defined.

Personally if it was up to me we would officially adopt diving bell and close this PR.

My major complaint with diving bell is the fact that it has to call skuba via the command line because skuba lacks an API. we should fix this to improve diving bell.

Anyhow. I think it would work really well in this application, and would let test runner worry only about running tests, not setting up infrastructure as well.

flavio:
I like the project created by Tyler, I think this could be useful and I think we should discuss about that. However this would take some time and right now we need a short term solution that addresses a concrete request.

This PR takes what we already have in place (writing the instructions to the stdout) and puts them into a simple script. The change is minimal and we can happily drop it once we're ready to ship a better solution.

dannysauer:
I know this is intended to be temporary, but temporary things often become permanent.  So I'm reviewing this like it's permanent. ;)

flavio:
@dannysauer thanks for the review!

I'll close this PR because we never got feedback from the user who initially requested it. They probably have created their own automation script.

I'll re-open this PR if needed

PR #960, opened 2020-02-19
Disable firewall in vmware+libvirt load balancers (SUSE/avant-garde#1337)

Itxaka:
We moved the firewalld service to be managed by skuba, but
unfortunately on libvirt and vmware the LB is a machine itself
and skuba doesnt run anything on there, so the firewalld service
could still be enabled and running

This patch disables firewalld on cloud-init for vmware and
libvirt

## Why is this PR needed?

Fixes https://github.com/SUSE/avant-garde/issues/1337

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/libvirt/cloud-init/lb.tpl
+++ b/ci/infra/libvirt/cloud-init/lb.tpl
@@ -49,6 +49,8 @@ runcmd:
   # With a new machine-id generated the journald daemon will work and can be restarted
   # Without a new machine-id it should be in a failed state
   - [ systemctl, restart, systemd-journald ]
+  # LB firewall is not disabled via skuba, so we need to manually disable it
+  - [ systemctl, disable, firewalld, --now ]
 
 bootcmd:
   - ip link set dev eth0 mtu 1400
--- a/ci/infra/vmware/cloud-init/lb.tpl
+++ b/ci/infra/vmware/cloud-init/lb.tpl
@@ -42,6 +42,9 @@ runcmd:
   # With a new machine-id generated the journald daemon will work and can be restarted
   # Without a new machine-id it should be in a failed state
   - [ systemctl, restart, systemd-journald ]
+  # LB firewall is not disabled via skuba, so we need to manually disable it
+  - [ systemctl, disable, firewalld, --now ]
+
 
 bootcmd:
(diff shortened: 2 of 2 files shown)


Itxaka:
> This was done in purpose, please have a look at: #927, skuba is taking care of this, so we untangle how infra is deployed from the skuba deployment.

How is that posible if you cant reach the LB from the deployer node because the firewall is up? We cant even bootstrap a node properly (hint, on kubelet start it tries to reach the apiserver but it cant becuase the firewall on the lb is blocking the calls)

The only reason this seems to work on openstack its because the openstack images do not come with firewalld anymore. On vmware it never worked (basically because we are not bootstrapping a cluster in that job, its a fake job that does nothing)

So either skuba its not doing it properly or there is an extra issue here. Let me have a deeper look at this

Itxaka:
> Normally we have this disabled by default in the images, which image do you use ?

@lcavajani default one used by CI, `SLES15-SP1-GM-up190905-guestinfo`

ereslibre:
Hm, that's certainly interesting. We are using SSH to connect to the new machine, and at the first stages we are disabling `firewalld` using `skuba`, so as long as `firewalld` allows connections on the 22 port (that I believe it should), there should not be any problem... am I missing something?

Itxaka:
> Hm, that's certainly interesting. We are using SSH to connect to the new machine, and at the first stages we are disabling firewalld using skuba, so as long as firewalld allows connections on the 22 port (that I believe it should), there should not be any problem... am I missing something?

We are both missing something I think.

Skuba runs on the master.

Load balancer is a different machine.

Ergo its impossible for skuba to disable the firewalld in the LB if it runs against the master. 

:laughing: :laughing:

ereslibre:
Maybe we can perform this at the `lb.tpl`, since it's specific to the LB then.

Itxaka:
> so it means we also need to re-enable this for libvirt as well

Yep!

> Maybe we can perform this at the lb.tpl, since it's specific to the LB then.

Agreed!

PR #964, opened 2020-02-20
WIP: DON'T MERGE: use bridge for libvirt

thehejik:
## Why is this PR needed?

This PR is not meant to be merged, it was created just for sharing purposes. 

Terraform libvirt deployment is by default configured to use NAT networking model managed by libvirt but it might be not useful for CI/validation jobs because then the guest machines are not reachable from outside (without port-forwarding set) but only from a hypervisor host.

Also this sample deployment was configured (see `terraform.tfvars.example.th`) to use remote libvirt host so now it's possible to call `terraform apply` from different machine (eg. from jenkins worker).

#### To make it work you'll need perform several steps:
* checkout this PR on worker machine
* rename `terraform.tfvars.example.th` to `terraform.tfvars`
* modify `libvirt_uri = "qemu+ssh://thehejik@zeus.qa.suse.cz/system"` URI pointing to your KVM in `terraform.tfvars` file
* either add your SCC key into `registration.auto.tfvars` or add repos into `terraform.tfvars`
* on worker machine as a user which will run terraform, run: `ssh-copy-id <myhypervisor.suse.cz>` (the user have to exists on hypervisor machine as well)
* configure libvirt daemon on your KVM hypervisor machine:
```
cat > /etc/libvirt/libvirtd.conf << EOF 
listen_tcp = 1
auth_unix_ro = "none"
auth_unix_rw = "none"
auth_tcp = "none"
auth_tls = "none"
EOF
```
* restart libvirtd `systemctl restart libvirtd`
* configure bridge on hypervisor machine:
```
cat >  /etc/sysconfig/network/ifcfg-br0 << EOF
BOOTPROTO='dhcp'
BRIDGE='yes'
BRIDGE_FORWARDDELAY='0'
BRIDGE_PORTS='eth0'
BRIDGE_STP='off'
STARTMODE='auto'
ZONE=public
EOF
```
* configure enslaved interface used by bridge (interface connected to uplink):
```
cat > /etc/sysconfig/network/ifcfg-eth0 << EOF
BOOTPROTO='none'
STARTMODE='auto'
EOF
```
* restart wicked `systemctl restart wicked`
* done, call `terraform apply` on worker machine

## What does this PR do?

It bootstraps SLES15-SP1 VMs on libvirt/KVM hypervisor used for CaaSP v4 deployment by skuba. The VMs are preconfigured to use bridge network model instead of default NAT.


thehejik:
\cc @pablochacin

thehejik:
BTW I just realized that NAT mode is not so suitable for users as well - then the management host has to be deployed within the libvirt network to have direct access to the nodes.

OR

With management host located outside of the libvirt network the deployment of the nodes would be a mess:
```
# skuba cluster init cluster --control-plane kvm_host_public_ip:26443
# skuba node bootstrap --user sles --sudo --target kvm_host_public_ip:20022 caasp-master-0
# skuba node join -s -u sles -t kvm_host_public_ip:20023 -r master caasp-master-1
...
```
also it would need to configure port-forwarding on hypervisor. 

@innobead @evrardjp would it be possible to change terraform libvirt to bridge mode instead of NAT according to this PR?

aojea:
if you disable NAT mode then the VMs will not able to download anything from outside, because the packets will have the VM address as the source, in order to work you have to route the internal subnet, that doesn't seem a good idea to me for a CI system.

pablochacin:
Just wanted to note that IIRC we are NOT delivering libvirt related terraform files to customers. This configuration is only for testing  -- mostly CI, but also locally-- and in that sense, keeping the NAT would be maybe better.

I'm testing an alternative to make VMs in a NAT network visible from the CI, replicating the setup we have for VMware cluster.

thehejik:
> if you disable NAT mode then the VMs will not able to download anything from outside, because the packets will have the VM address as the source, in order to work you have to route the internal subnet, that doesn't seem a good idea to me for a CI system.

I'm replacing NAT by bridge so VMs will lease IP/route/DNS directly from company's network
DHCPd. When using bridge there is no internal subnet, the nodes will use "public" SUSE IPs directly. 

The very same approach was used in CaaSP v3 and bridge is also used in CaaSP v4 VMware terraform deployment.

The NAT makes it similar to OpenStack deployment but libvirt doesn't offer floating IPs.

pablochacin:
> The NAT makes it similar to OpenStack deployment but libvirt doesn't offer floating IPs.

That's true. I'm approaching this by creating static routes from CI workers to the VMs, as we did with VMware. This is the same approach I used locally between my laptop and workstation. So it can be used also for testing locally without having to modify the terraform.

innobead:
> > if you disable NAT mode then the VMs will not able to download anything from outside, because the packets will have the VM address as the source, in order to work you have to route the internal subnet, that doesn't seem a good idea to me for a CI system.
> 
> I'm replacing NAT by bridge so VMs will lease IP/route/DNS directly from company's network
> DHCPd. When using bridge there is no internal subnet, the nodes will use "public" SUSE IPs directly.
> 
> The very same approach was used in CaaSP v3 and bridge is also used in CaaSP v4 VMware terraform deployment (which is used by CI as well).
> 
> The NAT makes it similar to OpenStack deployment but libvirt doesn't offer floating IPs.

Thanks @thehejik for this PR. 

Even this PR comes from a CI requirement, but actually raised a good point for rethinking what case we really want to support for KVM infra deployment? especially we are about to release KVM infra offering. 

Like you said, the current KVM terraform scripts is more about local deployment which means management & provision/hypervisor on the same host that is totally different from other TF infra offerings including VMware, OpenStack and AWS. 

I think this approach will make KVM offering similar as others and being able to deploy from a management host different from the provision/hypervisor node. I am not against this change actually but just a little concern about the configurations of the provision/hypervisor node becoming a little tedious. 

cc @jordimassaguerpla @Klaven @cmurphy  related people involving in KVM offering to provide feedbacks.

jordimassaguerpla:
We are not releasing terraform libvirt to our customers. We assume they have a kvm cluster. Thus, this is only for CI.

pablochacin:
This PR was used for discussing this alternative, but the actual implementation of KVM networking in CI will be different.

PR #965, opened 2020-02-24
Better node count for tests

Itxaka:
Currently we rely on the output of skuba to know the number of
masters and workers. This is a very flimsy way of getting those
numbers as the output can change and its not really ready for
batch processing.

Instead, lets use kubectl to get those, which gives us a more
stable way of doing so thanks to selectors and the possibility
of getting the exact output we want.

This patch moves from using skuba output to guess the number of
masters to using kubectl to get the exact number of masters. The
new function can be used to get the workers as well.

## Why is this PR needed?

Fixes https://github.com/SUSE/avant-garde/issues/1349

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/kubectl/kubectl.py
+++ b/ci/infra/testrunner/kubectl/kubectl.py
@@ -20,3 +20,19 @@ class Kubectl:
             return self.utils.runshellcommand(shell_cmd, stdin=stdin)
         except Exception as ex:
             raise Exception("Error executing cmd {}".format(shell_cmd)) from ex
+
+    def get_node_names_by_role(self, role):
+        """Returns a list of node names for a given role
+        Uses selectors to get the nodes. Master nodes have the node-role.kubernetes.io/master="" label, while other
+        nodes (workers) dont even have the label.
+        """
+
+        if role not in ("master", "worker"):
+            raise ValueError("Invalid role {}".format(role))
+
+        roles = {
+            "master": "==",
+            "worker": "!="
+        }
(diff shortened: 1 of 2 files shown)


Itxaka:
This seems to solve the stability issues I have found with the `test_remove_master` test which seems to fail due to wrongly getting the master numbers

pablochacin:
In the context of this test, it makes sense to introduce this change. However, it leaves unsolved the root cause of the problem.

One possibility is that at the moment the test is executed, not all masters have joined the cluster. May it be possible the test is executed after bootstrapping the cluster but before the join-nodes is execute?

Itxaka:
>  May it be possible the test is executed after bootstrapping the cluster but before the join-nodes is execute?

join-nodes runs before executing any e2e tests so it not should be possible for that to happen.

Unfortunately the e2e tests output is useless as its dumping all the json responses from k8s into the output its really difficult to be able to read anything in the logs.

pablochacin:
I'm in favor of approving this PR and open an issue regarding  `skuba.num_of_nodes("master")` not returning the correct number of nodes. 

Regarding the verbosity of the output, this is due to the way the status of the pods is checked in some tests: instead of retrieving only the status, the whole pod in retrieved. This could be improved and use a similar approach than used in [check_nodes_ready](https://github.com/SUSE/skuba/blob/master/ci/infra/testrunner/tests/utils.py#L6-L14)

Itxaka:
@pablochacin I created the following card for this issues: https://github.com/SUSE/avant-garde/issues/1349 can you re-review?

PR #973, opened 2020-02-26
WIP: ssh to all nodes in a role + ramfs for etcd

Itxaka:
## Why is this PR needed?

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
+++ b/ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
@@ -108,6 +108,7 @@ pipeline {
 
         stage('Cluster Bootstrap') {
             steps {
+                sh(script: 'make -f skuba/ci/Makefile fix_etcd', label: 'ramfs for etcd dir')
                 sh(script: 'make -f skuba/ci/Makefile bootstrap', label: 'Bootstrap')
             }
         }
--- a/ci/Makefile
+++ b/ci/Makefile
@@ -140,3 +140,7 @@ test_remove_master:
 .PHONY: test_remove_worker
 test_remove_worker:
 	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} test --verbose --suite test_workers.py --test test_remove_worker
+
+.PHONY: fix_etcd
+fix_etcd:
+	${TEST_RUNNER} -v vars.yaml -p ${PLATFORM} ssh_all -r master -c "sudo mkdir -p /var/lib/etcd && sudo mount -t tmpfs -o size=1048M tmpfs /var/lib/etcd"
--- a/ci/infra/testrunner/testrunner.py
+++ b/ci/infra/testrunner/testrunner.py
@@ -99,6 +99,14 @@ def ssh(options):
         role=options.role, nr=options.node, cmd=" ".join(options.cmd))
 
 
+def ssh_all(options):
+    plat = platforms.get_platform(options.conf, options.platform)
+    ipaddress = plat.get_nodes_ipaddrs(options.role)
+
+    for node in ipaddress:
+        plat.utils.ssh_run(node, " ".join(options.cmd))
+
+
 def main():
     help_str = """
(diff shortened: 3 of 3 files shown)


ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
```diff
@@ -108,6 +108,7 @@ pipeline {
 
         stage('Cluster Bootstrap') {
             steps {
+                sh(script: 'skuba/ci/infra/testrunner/testrunner --platform ${PLATFORM} ssh_all -r master "sudo mount -t tmpfs -o size=1048M tmpfs /var/lib/etcd"', label: 'ramfs for etcd dirs')
```
aojea: 1Gb of ram for etcd is not excessive?

Itxaka: arent you sick? :P

Its just WIP, I dont even know if this will work. Will redo it later with a more proper size *if* this work and its a decent stopgap for our current issues with etcd

jenting:
I :heart: this improvement.

davidcassany:
I am surprised that the node reboot tests passed... that makes me wonder about the test sanity... I understand the point on ETCD is having persistent status. Thus it should not be possible to reuse a node that rebooted, unless you consider it as new fresh one.

I don't think our node reboot test should pass with this setup.

aojea:
> I am surprised that the node reboot tests passed... that makes me wonder about the test sanity... I understand the point on ETCD is having persistent status. Thus it should not be possible to reuse a node that rebooted, unless you consider it as new fresh one.
> 
> I don't think our node reboot test should pass with this setup.

where is the test code?
what nodes are rebooted?
If all the etcd nodes are rebooted at the same time you are completely right, the test shouldn't pass, the state is lost ... or we have an incredible feature we weren't aware of ;)

Itxaka:
> I don't think our node reboot test should pass with this setup.

why not? We only reboot one node, and it will come up as follower. It will probably catch up with the rest of the nodes, if etcd even has replication to slaves like mysql/postgres*. It will come up, start etcd and will probably recreate the dir from scratch with normal storage, but by that time, we are not hitting the API that much as to hit the same error so it probably wont happen again.

Still, the fix its as easy as instead of temp mounting the dir, writing the same line to fstab and mount -o remount all so it lives trougth the reboot :D


* Dont know how etcd replication to new followers work so this may be wrong

pablochacin:
As @davidcassany mentioned, I don't like this to be used in any pipeline as it introduces a source of fragility in the tests. This was the reason this idea was tried and rejected in the past.

Itxaka:
> As @davidcassany mentioned, I don't like this to be used in any pipeline as it introduces a source of fragility in the tests. This was the reason this idea was tried and rejected in the past.

How is this a source of fragility when its fixing a flaky issue upstream??? In any case its fixing a fragility issue which is out of our hands? its a hack or a temp fix, or whatever you wanna call it, but source of fragility I would not call it ever :P

pablochacin:
> > As @davidcassany mentioned, I don't like this to be used in any pipeline as it introduces a source of fragility in the tests. This was the reason this idea was tried and rejected in the past.
> 
> How is this a source of fragility when its fixing a flaky issue upstream??? In any case its fixing a fragility issue which is out of our hands? its a hack or a temp fix, or whatever you wanna call it, but source of fragility I would not call it ever :P

Not sure I follow your arguments. Introducing a "fix" that can cause failures due to unexpected behavior (etcd state disappearing at the middle of a test). 

As how to solve the issue, there's a clear solution in our hards: better hardware for testing. And this is a know situation for management and AFIK, something that is under consideration.

Itxaka:
talked and agreed on the meeting, closing!

PR #977, opened 2020-02-28
Fix errors for skuba-update tests

suseclee:
skuba-update test runs when skuba-update folder has any code change.
My PR still has failures because skuba-update tests had errors. 

1. Error when six is installed in a container.
CI skuba-update test has an error because skuba-update-unit-build:latest
has six 1.11.0 installed.

2. shell checks all the unnecessary sh files from root in a container env.
tox.init generates errors  on `bash -c 'find . -name "*.sh" -exec shellcheck  -e SC1090,SC1117 \{\} \+'` because of all sh scripts from /.  

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/skuba-update/tox.ini
+++ b/skuba-update/tox.ini
@@ -69,4 +69,4 @@ setenv =
 commands =
     flake8 --statistics -j auto --count {toxinidir}/skuba_update
     flake8 --statistics -j auto --count {toxinidir}/test
-    bash -c 'find . -name "*.sh" -exec shellcheck -e SC1090,SC1117 \{\} \+'
+    find {toxinidir}/test -name "*.sh" -exec shellcheck -e SC1090 \{\} +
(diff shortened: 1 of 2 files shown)


skuba-update/tox.ini
```diff
@@ -69,4 +69,4 @@ setenv =
 commands =
     flake8 --statistics -j auto --count {toxinidir}/skuba_update
     flake8 --statistics -j auto --count {toxinidir}/test
-    bash -c 'find . -name "*.sh" -exec shellcheck -e SC1090,SC1117 \{\} \+'
+    bash -c 'find {toxinidir}/test -name "*.sh" -exec shellcheck -e SC1090,SC1117 \{\} \+'
```
dannysauer: Any idea why this is wrapped in a `bash -c` instead of just directly running find?

jodavis-suse: I admit, I'm not familiar with 'shellcheck' as a command, so this whole line was a little confusing to me.

suseclee: Fixed both with `find {toxinidir}/test -name "*.sh" -exec shellcheck -e SC1090 \{\} +`
Thanks for your reviews.

PR #981, opened 2020-03-03
fix issue for set workload revisionHistoryLimit [bsc#1162651]

cclhsu:
fix issue items for set workload revisionHistoryLimit [bsc#1162651]

Signed-off-by: cclhsu <clark.hsu@suse.com>

## Why is this PR needed?

Reduce revisionHistoryLimit to as decribed in https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#revision-history-limit.  Per discussion, [#1162651](https://bugzilla.suse.com/show_bug.cgi?id=1162651), change our revisionHistoryLimit of our workloads from 10 to 3.
 
```
# kubectl -n kube-system get deployments,replicasets,daemonset -o jsonpath='{range .items[*]}{@.kind}{" "}{@.metadata.name}{" "}{@.spec.revisionHistoryLimit}{"\n"}'

Deployment cilium-operator 10
Deployment oidc-dex 10
Deployment oidc-gangway 10
DaemonSet cilium 10
DaemonSet kured 10
```

Fixes [#1162651](https://bugzilla.suse.com/show_bug.cgi?id=1162651)

## What does this PR do?

```
# kubectl -n kube-system get deployments,replicasets,daemonset -o jsonpath='{range .items[*]}{@.kind}{" "}{@.metadata.name}{" "}{@.spec.revisionHistoryLimit}{"\n"}'

Deployment cilium-operator 3
Deployment oidc-dex 3
Deployment oidc-gangway 3
DaemonSet cilium 3
DaemonSet kured 3
```

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubernetes/versions.go
+++ b/internal/pkg/skuba/kubernetes/versions.go
@@ -92,8 +92,8 @@ var (
 				Tooling:   &ContainerImageTag{Name: "skuba-tooling", Tag: "0.1.0"},
 			},
 			AddonsVersion: AddonsVersion{
-				Cilium:        &AddonVersion{"1.5.3", 1},
-				Kured:         &AddonVersion{"1.3.0", 3},
+				Cilium:        &AddonVersion{"1.5.3", 2},
+				Kured:         &AddonVersion{"1.3.0", 4},
 				Dex:           &AddonVersion{"2.16.0", 5},
 				Gangway:       &AddonVersion{"3.1.0-rev4", 4},
 				MetricsServer: &AddonVersion{"0.3.6", 0},
@@ -113,8 +113,8 @@ var (
 				Tooling:   &ContainerImageTag{Name: "skuba-tooling", Tag: "0.1.0"},
 			},
 			AddonsVersion: AddonsVersion{
-				Cilium:  &AddonVersion{"1.5.3", 1},
-				Kured:   &AddonVersion{"1.2.0-rev4", 1},
+				Cilium:  &AddonVersion{"1.5.3", 2},
+				Kured:   &AddonVersion{"1.2.0-rev4", 2},
 				Dex:     &AddonVersion{"2.16.0", 5},
 				Gangway: &AddonVersion{"3.1.0-rev4", 4},
 				PSP:     &AddonVersion{"", 1},
--- a/internal/pkg/skuba/addons/cilium.go
+++ b/internal/pkg/skuba/addons/cilium.go
@@ -112,6 +112,7 @@ metadata:
   name: cilium
   namespace: kube-system
 spec:
+  revisionHistoryLimit: 3
   updateStrategy:
     type: "RollingUpdate"
     rollingUpdate:
@@ -324,6 +325,7 @@ metadata:
   name: cilium-operator
   namespace: kube-system
 spec:
+  revisionHistoryLimit: 3
   replicas: 1
   selector:
     matchLabels:
--- a/internal/pkg/skuba/addons/kured.go
+++ b/internal/pkg/skuba/addons/kured.go
@@ -132,6 +132,7 @@ spec:
   selector:
     matchLabels:
       name: kured
+  revisionHistoryLimit: 3
   updateStrategy:
     type: RollingUpdate
   template:


c3y1huang:
Think missing manifest version bump. 

And could we add some test-case along side this?

internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -111,10 +111,10 @@ var (
...
 			},
 			AddonsVersion: AddonsVersion{
-				Cilium:  &AddonVersion{"1.5.3", 1},
-				Kured:   &AddonVersion{"1.2.0-rev4", 1},
-				Dex:     &AddonVersion{"2.16.0", 4},
-				Gangway: &AddonVersion{"3.1.0-rev4", 3},
+				Cilium:  &AddonVersion{"1.5.3", 2},
+				Kured:   &AddonVersion{"1.2.0-rev4", 2},
+				Dex:     &AddonVersion{"2.16.0", 5},
+				Gangway: &AddonVersion{"3.1.0-rev4", 4},
```
jenting: IMO, I don't think we need to bump K8s 1.15.2 cluster addons manifest version.

internal/pkg/skuba/addons/dex.go
```diff
@@ -164,6 +164,7 @@ metadata:
   labels:
     app: oidc-dex
 spec:
+  revisionHistoryLimit: 3
```
jenting: Already in https://github.com/SUSE/skuba/blob/c8998e8/internal/pkg/skuba/addons/dex.go#L181

internal/pkg/skuba/addons/gangway.go
```diff
@@ -125,6 +125,7 @@ metadata:
   labels:
     app: oidc-gangway
 spec:
+  revisionHistoryLimit: 3
```
jenting: Already in https://github.com/SUSE/skuba/blob/c8998e8/internal/pkg/skuba/addons/gangway.go#L139

internal/pkg/skuba/kubernetes/versions.go
c3y1huang: I couldn't think of a reason customer installed with K8s 1.15.2 shouldn't also benefit from this fix.

jenting: From the product perspective.

AFAIK, we don't have to take 4.0.x once 4.1.x released. The customer has to upgrade from 4.0.x to 4.1.x first.

On the other hand, I saw kured bump container image version only on `1.16.2`. Intended or missing?

c3y1huang: > From the product perspective.
> 
> AFAIK, we don't have to take 4.0.x once 4.1.x released. The customer has to upgrade from 4.0.x to 4.1.x first.

Is upgrade to K8s 1.16.2 is enforced if customer have cluster with 1.15.2? Or could customer upgrade addon only?

> 
> On the other hand, I saw kured bump container image version only on `1.16.2`. Intended or missing?

jenting: I think the customer could upgrade addon only~

innobead: Yep, it looks like if would like to benefit for the current cluster w/o cluster upgrade, we should bump addon version for what cluster version we would like to support.

So it should be fine for this change.

https://github.com/SUSE/skuba/blob/c5895c7bfc9fd5fdaa10e6997bb52364b7f69e43/pkg/skuba/actions/addon/upgrade/plan.go#L51

c3y1huang:
There is still a on going [discussion](https://bugzilla.suse.com/show_bug.cgi?id=1162651#c5) about what revision should be set for cillium* .

jordimassaguerpla:
This is blocking packaging for 4.2.0. Any ETA on this? Does it really need to be in 4.2.0? If so, why?

c3y1huang:
@cclhsu since this is a bug fix. Could you add test case along side this?

innobead:
Good reminder! Chinya :)

PR #983, opened 2020-03-05
Include libvirt platform as part of the PRs tests

davidcassany:
## Why is this PR needed?

This PR is needed to start using libvirt platform to tests skuba PRs in addition to the current VMWare tests.

Part of SUSE/avant-garde#1227

## What does this PR do?

This commit sets to test skuba PRs using testrunner for libvirt and
vmware platforms. Moreover it cleans up some of the previous old and
unused VMWare job settings.

## Info for QA

None, it is CI PR.

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/libvirt/README.md
+++ b/ci/infra/libvirt/README.md
@@ -35,6 +35,8 @@ Copy the `terraform.tfvars.example` to `terraform.tfvars` and provide reasonable
 
 ## Variables
 
+`libvirt_uri` - URL of the libvirt host
+`libvirt_keyfile` - The key file used for the connection to a remote libvirt server
 `image_uri` - URL of the image to use
 `stack_name` - Identifier to make all your resources unique and avoid clashes with other users of this terraform project
 `authorized_keys` - A list of ssh public keys that will be installed on all nodes
--- a/ci/infra/libvirt/terraform.tfvars.json.ci.example
+++ b/ci/infra/libvirt/terraform.tfvars.json.ci.example
@@ -1,14 +1,14 @@
 {
     "libvirt_uri": "qemu:///system",
+    "libvirt_keyfile": "",
     "pool": "default",
-    "image_uri": "SLES15-SP1-JeOS.x86_64-15.1-OpenStack-Cloud-QU1.qcow2",
+    "image_uri": "",
     "stack_name": "testing",
     "network_cidr": "10.17.0.0/22",
     "dns_domain": "caasp.local",
     "masters": 1,
     "workers": 2,
     "username": "sles",
-    "password": "linux",
     "repositories": {
         "caasp_devel": "http://download.suse.de/ibs/Devel:/CaaSP:/4.0/SLE_15_SP1/",
         "suse_ca": "http://download.suse.de/ibs/SUSE:/CA/SLE_15_SP1/",
--- a/ci/infra/testrunner/platforms/libvirt.py
+++ b/ci/infra/testrunner/platforms/libvirt.py
@@ -9,6 +9,11 @@ from utils import Format
 class Libvirt(Terraform):
     def __init__(self, conf):
         super().__init__(conf, 'libvirt')
+        self.platform_new_vars = {
+            "libvirt_uri": self.conf.libvirt.uri,
+            "libvirt_keyfile": self.conf.libvirt.keyfile,
+            "image_uri": self.conf.libvirt.image_uri
+        }
 
     def _env_setup_cmd(self):
         return ":"
--- a/ci/infra/testrunner/platforms/terraform.py
+++ b/ci/infra/testrunner/platforms/terraform.py
@@ -151,6 +151,8 @@ class Terraform(Platform):
             "authorized_keys": [self.utils.authorized_keys()]
         }
 
+        new_vars.update(self.platform_new_vars)
+
         for k, v in new_vars.items():
             if tfvars.get(k) is not None:
                 if isinstance(v, list):
--- a/ci/jenkins/jobs.yaml
+++ b/ci/jenkins/jobs.yaml
@@ -3,10 +3,9 @@
     repo-name: skuba
     repo-owner: SUSE
     repo-credentials: github-token
-    platform: openstack
     jobs:
         - '{name}/test'
-        - '{name}/test-vmware'
+        - '{name}/test-libvirt'
         - '{name}/update-unit'
         - '{name}/update-acceptance'
         - '{name}/code-lint'
--- a/ci/jenkins/pipelines/prs/skuba-test-libvirt.Jenkinsfile
+++ b/ci/jenkins/pipelines/prs/skuba-test-libvirt.Jenkinsfile
@@ -1,0 +1,146 @@
+/**
+ * This pipeline verifies on a Github PR:
+ *   - skuba unit and e2e tests
+ *   - Basic skuba deployment, bootstrapping, and adding nodes to a cluster
+ */
+
+pipeline {
+    agent { node { label 'caasp-team-private-integration-nue && kvm' } }
+
+    environment {
+        SKUBA_BINPATH = '/home/jenkins/go/bin/skuba'
+        GITHUB_TOKEN = credentials('github-token')
+        PLATFORM = 'libvirt'
+        TERRAFORM_STACK_NAME = "${JOB_NAME.replaceAll("/","-")}-${BUILD_NUMBER}".take(70)
+        PR_CONTEXT = 'jenkins/skuba-test-libvirt'
+        PR_MANAGER = 'ci/jenkins/pipelines/prs/helpers/pr-manager'
+        REQUESTS_CA_BUNDLE = '/var/lib/ca-certificates/ca-bundle.pem'
+        PIP_VERBOSE = 'true'
+        LIBVIRT_URI = 'qemu+ssh://jenkins@node52.cloud.suse.de/system'
+        LIBVIRT_KEYFILE = credentials('libvirt-keyfile')
+        LIBVIRT_IMAGE_URI = 'http://download.suse.de/install/SLE-15-SP1-JeOS-QU2/SLES15-SP1-JeOS.x86_64-15.1-OpenStack-Cloud-QU2.qcow2'
+    }
+
+    stages {
+        stage('Collaborator Check') { steps { script {
+            if (env.BRANCH_NAME.startsWith('PR')) {
+                def membersResponse = httpRequest(
+                    url: "https://api.github.com/repos/SUSE/skuba/collaborators/${CHANGE_AUTHOR}",
+                    authentication: 'github-token',
+                    validResponseCodes: "204:404")
+
+                if (membersResponse.status == 204) {
+                    echo "Test execution for collaborator ${CHANGE_AUTHOR} allowed"
+
+                } else {
+                    def allowExecution = false
+
+                    try {
+                        timeout(time: 5, unit: 'MINUTES') {
+                            allowExecution = input(id: 'userInput', message: "Change author is not a SUSE member: ${CHANGE_AUTHOR}", parameters: [
+                                booleanParam(name: 'allowExecution', defaultValue: false, description: 'Run tests anyway?')
+                            ])
+                        }
+                    } catch(err) {
+                        def user = err.getCauses()[0].getUser()
+                        if('SYSTEM' == user.toString()) {
+                            echo "Timeout while waiting for input"
+                        } else {
+                            allowExecution = false
+                            echo "Unhandled error:\n${err}"
+                        }
+                    }
+
+                    if (!allowExecution) {
+                        echo "Test execution for unknown user (${CHANGE_AUTHOR}) disallowed"
+                        error(message: "Test execution for unknown user (${CHANGE_AUTHOR}) disallowed")
+                        return;
+                    }
+                }
+            }
+        } } }
+
+        stage('Setting GitHub in-progress status') { steps {
+            sh(script: "${PR_MANAGER} update-pr-status ${GIT_COMMIT} ${PR_CONTEXT} 'pending'", label: "Sending pending status")
+        } }
+
+        stage('Git Clone') { steps {
+            deleteDir()
+            checkout([$class: 'GitSCM',
+                      branches: [[name: "*/${BRANCH_NAME}"], [name: '*/master']],
+                      doGenerateSubmoduleConfigurations: false,
+                      extensions: [[$class: 'LocalBranch'],
+                                   [$class: 'WipeWorkspace'],
+                                   [$class: 'RelativeTargetDirectory', relativeTargetDir: 'skuba']],
+                      submoduleCfg: [],
+                      userRemoteConfigs: [[refspec: '+refs/pull/*/head:refs/remotes/origin/PR-*',
+                                           credentialsId: 'github-token',
+                                           url: 'https://github.com/SUSE/skuba']]])
+
+            dir("${WORKSPACE}/skuba") {
+                sh(script: "git checkout ${BRANCH_NAME}", label: "Checkout PR Branch")
+            }
+        }}
+
+        stage('Run skuba unit tests') { steps {
+            dir("skuba") {
+              sh(script: 'make test-unit', label: 'make test-unit')
+            }
+        } }
+
+        stage('Getting Ready For Cluster Deployment') { steps {
+            sh(script: 'make -f skuba/ci/Makefile pre_deployment', label: 'Pre Deployment')
+            sh(script: 'make -f skuba/ci/Makefile pr_checks', label: 'PR Checks')
+            sh(script: "pushd skuba; make -f Makefile install; popd", label: 'Build Skuba')
+        } }
+
+        stage('Deploy cluster') {
+            steps {
+                sh(script: 'make -f skuba/ci/Makefile create_environment', label: 'Provision')
+                sh(script: 'make -f skuba/ci/Makefile bootstrap', label: 'Bootstrap')
+                sh(script: 'make -f skuba/ci/Makefile join_nodes', label: 'Join Nodes')
+            }
+        }
+
+        stage('Run e2e tests') { steps {
+            sh(script: "make -f skuba/ci/Makefile test_pr", label: "test_pr")
+        } }
+
+    }
+    post {
+        always {
+            archiveArtifacts(artifacts: "skuba/ci/infra/${PLATFORM}/terraform.tfstate", allowEmptyArchive: true)
+            archiveArtifacts(artifacts: "skuba/ci/infra/${PLATFORM}/terraform.tfvars.json", allowEmptyArchive: true)
+            archiveArtifacts(artifacts: 'testrunner.log', allowEmptyArchive: true)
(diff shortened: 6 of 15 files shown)


ci/infra/libvirt/terraform.tfvars.json.ci.example
```diff
@@ -1,9 +1,9 @@
 {
-    "libvirt_uri": "qemu:///system",
+    "libvirt_uri": "qemu+ssh://jenkins@node52.cloud.suse.de/system",
```
davidcassany: I am wondering if this should be part of configuration options in `vars.yml` or part of an ENV file

ci/infra/libvirt/terraform.tfvars.json.ci.example
```diff
@@ -1,9 +1,9 @@
...
     "pool": "default",
-    "image_uri": "SLES15-SP1-JeOS.x86_64-15.1-OpenStack-Cloud-QU1.qcow2",
+    "image_uri": "http://download.suse.de/ibs/SUSE:/SLE-15-SP1:/GA/images/SLES15-SP1-JeOS.x86_64-15.1-OpenStack-Cloud-Build35.30.qcow2",
     "stack_name": "testing",
-    "network_cidr": "10.17.0.0/22",
+    "network_cidr": "10.164.80.0/20",
```
davidcassany: Same as above, shouldn't it be configurable form a vars.yml and/or env file?

ci/infra/libvirt/terraform.tfvars.json.ci.example
```diff
@@ -1,9 +1,9 @@
 {
-    "libvirt_uri": "qemu:///system",
+    "libvirt_uri": "qemu+ssh://jenkins@node52.cloud.suse.de/system",
     "pool": "default",
-    "image_uri": "SLES15-SP1-JeOS.x86_64-15.1-OpenStack-Cloud-QU1.qcow2",
+    "image_uri": "http://download.suse.de/ibs/SUSE:/SLE-15-SP1:/GA/images/SLES15-SP1-JeOS.x86_64-15.1-OpenStack-Cloud-Build35.30.qcow2",
```
davidcassany: We need to agree on which image to use and the URI should be static, this one is depending on the build.

ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
```diff
@@ -11,9 +11,8 @@ pipeline {
         SKUBA_BINPATH = '/home/jenkins/go/bin/skuba'
         VMWARE_ENV_FILE = credentials('vmware-env')
         GITHUB_TOKEN = credentials('github-token')
-        PLATFORM = 'vmware'
         TERRAFORM_STACK_NAME = "${BUILD_NUMBER}-${JOB_NAME.replaceAll("/","-")}".take(70)
-        PR_CONTEXT = 'jenkins/skuba-test'
+        PR_CONTEXT = 'jenkins/skuba-test-${PLATFORM}'
```
davidcassany: Note this is likely to break PRs requirements in skuba github repo as there will not be anymore a `skuba-test`, which is a required one at the moment, not sure how and where is this configured, I guess on github UI.

ci/infra/libvirt/terraform.tfvars.json.ci.example
Itxaka: I would say vars, because that gives us automatic ENV vars so its easy to override and easy to track (both in file and easily overridable)

The problem I see here is, we can only manage a kvm host at once? So to deploy on kvm we need a huge machine to be able to deploy caasp in just one? How does that work for customers wanting to deploy in several servers? They have to keep changing the kvm host? How does networking work with that?

ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
Itxaka: github-ui -> protected branches I think :)

ci/infra/libvirt/terraform.tfvars.json.ci.example
davidcassany: `terraform-provider-libvirt` is just not supported at all, thus this is out of the scope of customers, this is only for internal use. Terraform plus libvirt tandem is just not delivered to customers, AFAIK we just deliver documentation about how to use CaaSP on top of libvirt (something close to bare metal procedure), not any sort of automation.

Itxaka: Perfect!

ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
davidcassany: This is no longer the case, I changed the PR to be really non intrusive just creating a new job with new PR context and do not touch anything else.

davidcassany:
Note that we should have a follow up card for improving this configuration. At least for the two areas:

1. Use an updated terraform-provider-libvirt that solves network restrictions
2. Use a generic name for the server, not `node52`

ci/infra/libvirt/terraform.tfvars.json.ci.example
```diff
@@ -1,14 +1,13 @@
 {
-    "libvirt_uri": "qemu:///system",
+    "libvirt_uri": "qemu+ssh://jenkins@node52.cloud.suse.de/system",
     "pool": "default",
-    "image_uri": "SLES15-SP1-JeOS.x86_64-15.1-OpenStack-Cloud-QU1.qcow2",
+    "image_uri": "http://download.suse.de/install/SLE-15-SP1-JeOS-QU2/SLES15-SP1-JeOS.x86_64-15.1-OpenStack-Cloud-QU2.qcow2",
     "stack_name": "testing",
-    "network_cidr": "10.17.0.0/22",
```
pablochacin: why are you changing this? this must match the range defined in the ip route to the kwm host

ci/infra/libvirt/terraform.tfvars.json.ci.example
```diff
@@ -1,14 +1,13 @@
 {
-    "libvirt_uri": "qemu:///system",
+    "libvirt_uri": "qemu+ssh://jenkins@node52.cloud.suse.de/system",
```
pablochacin: How are the credentials been passed to libvirt?

ci/infra/libvirt/terraform.tfvars.json.ci.example
davidcassany: I know, that's why I used the values that appeared in the docs for the jump host in https://github.com/SUSE/caasp-automation/tree/master/ecp-nbg-jumphost

ci/jenkins/pipelines/prs/skuba-test-libvirt.Jenkinsfile
```diff
@@ -0,0 +1,166 @@
+/**
+ * This pipeline verifies on a Github PR:
+ *   - skuba unit and e2e tests
+ *   - Basic skuba deployment, bootstrapping, and adding nodes to a cluster
+ */
+
+pipeline {
+    agent { node { label 'caasp-team-private-integration-nue && kvm' } }
+
+    environment {
```
pablochacin: We need to add the credentials for the `jenkins` user used to acces the KVM host. Those are not defined in the workers in general (I copied them manually to the test worker)

ci/infra/libvirt/terraform.tfvars.json.ci.example
davidcassany: I assume the jenkins user for on CI workers to use the jenkins key pair.

ci/jenkins/pipelines/prs/skuba-test-libvirt.Jenkinsfile
davidcassany: I am wondering what it needs, I guess it just requires the ssh key and I am not sure how to tell libvirt provider which one to use. Gonna review if there is anything in the docs, not sure we can configure the ssh key for the server, it might use the once associated by default on the current user.

ci/infra/libvirt/terraform.tfvars.json.ci.example
davidcassany: OK after some investigations I think the way to pass the custom ssh key to the libvirt URI is by passing is as a parameter of the keyfile within the URI, so something like: `qemu+ssh://jenkins@node52.cloud.suse.de/system?keyfile=<file>`. I find it weird adding a file reference of the client host for a remote URI, but it is the only way I found to use something else than `~/.ssh/id_rsa`.

ci/jenkins/pipelines/prs/skuba-test-libvirt.Jenkinsfile
pablochacin: I forgot to mention this: I was using this uri for the qemu connection in the test worker `qemu+ssh://jenkins@node52.cloud.suse.de/system` and it was working, BUT now it is not. I need to check. I remember at some point I setup the ssh-agent, but now even if I do, it is not working.

More strangely, if I test the connection with `virsh` and add the key to `ssh-agent` it works
```
$> eval (ssh-agent)
$> ssh-add /home/jenkins/.ssh/jenkins
$>virsh -c qemu+ssh://jenkins@node52.cloud.suse.de/system
```

ci/infra/libvirt/terraform.tfvars.json.ci.example
pablochacin: I forgot to mention this: I was using this uri for the qemu connection in the test worker `qemu+ssh://jenkins@node52.cloud.suse.de/system` and it was working, BUT now it is not. I need to check. I remember at some point I setup the ssh-agent, but now even if I do, it is not working.

More strangely, if I test the connection with `virsh` and add the key to `ssh-agent` it works
```
$> eval (ssh-agent)
$> ssh-add /home/jenkins/.ssh/jenkins
$>virsh -c qemu+ssh://jenkins@node52.cloud.suse.de/system
```

ci/jenkins/pipelines/prs/skuba-test-libvirt.Jenkinsfile
pablochacin: We can add this key-pair to jenkins as credentials and use in the same way we use other credentials.

ci/infra/libvirt/terraform.tfvars.json.ci.example
davidcassany: `virsh -c qemu+ssh://jenkins@node52.cloud.suse.de/system` 
This works for me if I properly set the keys or I use in URI parameter `keyfile` to point to the specific private key.
Now I seeing if I can configure a credentials in Jenkins to use a URI similar to  `qemu+ssh://jenkins@node52.cloud.suse.de/system?keyfile=${jenkins_key}` so we do not depend on certain specific underlaying configuration on the worker.

pablochacin: This is weird. Now it works with the uri as you describe. Earlier I tried it and terraform kept asking me for a password (I suppose it was not finding the ssh key). Better to keep :eyes: on this.

ci/infra/libvirt/terraform.tfvars.json.ci.example
pablochacin: I think that is the network for the VMware VMs. The kvm host is used to provide the route. We should update that document to reflect that now there are two VM networks: one for VMware and other for libivirt.

davidcassany:
@pablochacin @Itxaka 

My plan regarding this PR is doing to following on Monday:

1. Unless something there isn't something terrible merge it on Monday
2. It is possible that it will not work, however it will not break any current stuff, as it is a new job
3. Run a follow up PR in which we will be able to actually modify and update and validate the code within the PR itself as the job will already be there and running.

This way we can verify and test the final setup within the PR process.

ci/jenkins/pipelines/prs/skuba-test-libvirt.Jenkinsfile
```diff
@@ -0,0 +1,169 @@
...
+        SKUBA_BINPATH = '/home/jenkins/go/bin/skuba'
+        GITHUB_TOKEN = credentials('github-token')
+        PLATFORM = 'libvirt'
+        TERRAFORM_STACK_NAME = "${BUILD_NUMBER}-${JOB_NAME.replaceAll("/","-")}".take(70)
+        PR_CONTEXT = 'jenkins/skuba-test-libvirt'
+        PR_MANAGER = 'ci/jenkins/pipelines/prs/helpers/pr-manager'
+        REQUESTS_CA_BUNDLE = '/var/lib/ca-certificates/ca-bundle.pem'
+        PIP_VERBOSE = 'true'
+        withCredentials(bindings: [sshUserPrivateKey(credentialsId: 'libvirt_credentials', keyFileVariable: 'KEY_FILE')]) {
```
davidcassany: I have doubts this will work, an alternative could be just passing as an environment value the location of the key file, however this is dependent on the worker node definition, I'd like to avoid that.

Probably the `withCredentials` block needs to encompass all testrunner calls...

ci/jenkins/jobs.yaml
```diff
@@ -3,10 +3,9 @@
     repo-name: skuba
     repo-owner: SUSE
     repo-credentials: github-token
-    platform: openstack
     jobs:
         - '{name}/test'
-        - '{name}/test-vmware'
```
pablochacin: I thought we were adding the libvirt, not replacing the vmware (at this stage)

davidcassany: The `test` now runs on vmware (previous openstack one) and the `test-vmware` was disabled and now reenabled again but renamed. In fact I did not reenable, just deleted and created a new file, but the `git` after delete and add understood it as a rename :P

pablochacin:
> My plan regarding this PR is doing to following on Monday:
> 
> 1. Unless something there isn't something terrible merge it on Monday
> 2. Play de waka mole game on the new job :P
> 3. It is possible that this new job doesn't work, however it will not break any current stuff, as it is a new job
> 4. Run a follow up PR in which we will be able to actually modify and update and validate the code within the PR itself as the job will already be there and running.
> 
> This way we can verify and test the final setup within the PR process.

@davidcassany if you only want to test this pipeline, what you could do is to change the pr-test job to use this pipeline. That way this PR will test it but no other PR will be affected, this until we can test this works.

davidcassany:
> @davidcassany if you only want to test this pipeline, what you could do is to change the pr-test job to use this pipeline. That way this PR will test it but no other PR will be affected, this until we can test this works.

Good point this is another option I did not consider. The only problem is that once validated I will have to revert it from pr-test and you'll have to trust me it works :)

pablochacin:
> Good point this is another option I did not consider. The only problem is that once validated I will have to revert it from pr-test and you'll have to trust me it works :)

We trust you, bro!  

But what I suggest is for you to test. I think we are not sure the jenkins credentials will work, for example. One you have tested, you can revert this change and submit the final version like this one.

ci/jenkins/pipelines/prs/skuba-test-libvirt.Jenkinsfile
Itxaka: we need to check if we have the https://plugins.jenkins.io/ssh-agent/ plugin in jenkins so we can use the `sshagent` call in the pipeline. That should set the proper keys to use, if Im understanding this correctly.

Unfortunately with that plugin we would need to wrap every step :/

davidcassany: > Unfortunately with that plugin we would need to wrap every step :/

Yes I already tested wrapping every step and it seams to work from a syntax perspective (I assume we have the plugin), the environment variable holding the ssh file exists. However i am wondering if this is enough for our use case, since with this setup it is likely that the file reference is changing on each step and probably terraform keeps the URI as part of the tfstate (I still need to verify that), so in that case we require a constant SSH key file for every step.

The simplest solution is having the file installed in the worker and just make reference of it, but this implies that our pipeline is somehow coupled with the worker configuration and setup. I am currently testing this approach [here](https://ci.suse.de/job/caasp-jobs/job/v4/job/pr/job/test/view/change-requests/job/PR-983/) and my biggest concern it that I am facing some non consistent failures during the deployment.

pablochacin: IIRC the credentials are the same for all the pipeline. At least, this is true for the credentials we use for Openstack and VMware. I don't see why this should be different with ssh keys.

ci/infra/testrunner/platforms/libvirt.py
```diff
@@ -16,3 +16,19 @@ def _env_setup_cmd(self):
...
         self.destroy()
+
+    def _update_platform_tfvars(self, tfvars):
+        """Update the tfvars that are specific to the platform"""
+        new_vars = {
+            "libvirt_uri": self.conf.libvirt.uri,
+            "libvirt_keyfile": self.conf.libvirt.keyfile
+        }
+
+        for k, v in new_vars.items():
```
Itxaka: should this be the same in a generic method and the new_vars be a class attribute? That way, less code :D

ci/infra/testrunner/platforms/terraform.py
```diff
@@ -151,6 +151,8 @@ def _update_tfvars(self, tfvars):
             "authorized_keys": [self.utils.authorized_keys()]
         }
 
+        new_vars.update(self.platform_new_vars)
```
Itxaka: on on-libvirt platforms this fails because there is no platform_new_vars so we need at least an empty dict as part of this class that other classes can override

davidcassany: :man_facepalming: OMG you are right, fixed now :)

ci/infra/libvirt/terraform.tfvars.json.ci.example
```diff
@@ -1,14 +1,14 @@
 {
     "libvirt_uri": "qemu:///system",
+    "libvirt_keyfile": "",
```
Itxaka: Do we need to add docs for this new var?

davidcassany: docs? what are you having in mind? We do not have any libvirt variable documented AFAIK, however probably we should add it in `terraform.tfvars.example` file. 

Probably I can improve the README for libvirt to mention at least the variables configurable in `vars.yml`. Is something like that what are you imagining?

davidcassany: What about now?

Itxaka: :+1:

ci/infra/libvirt/README.md
```diff
@@ -35,6 +35,8 @@ Copy the `terraform.tfvars.example` to `terraform.tfvars` and provide reasonable
 
 ## Variables
 
+`libvirt_uri` - URL of the libvirt host
```
pablochacin: I think we should add also a section in the testrunner's `README.md` for this libvirt specific variables.

davidcassany: Yes doing it right now :+1:

PR #988, opened 2020-03-06
CI: simplify Makefile e2e test targets

Itxaka:
## Why is this PR needed?

Check commit info more an extended description

Fixes https://github.com/SUSE/avant-garde/issues/1283

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


ci/infra/testrunner/tests/test_addon_upgrade_apply.py
```diff
@@ -0,0 +1,21 @@
+import yaml
```
davidcassany: I would not divide addons tests in separate suites, those tests are not disruptive, neither having different environment considerations, they can be all part of the same suite.

Itxaka:
> I believe we should only split the ones that require it.

If we dont separate them, then the simplify of the test targets does not work anymore and the old complexity is moved from the test targets into the jenkins job definition as to execute a test inside a test suite. We agreed to do this as to simplify the CI part, then move into tackling the test, if they are still gonna be there or not, if we should regroup or not, etc....

ci/infra/testrunner/tests/utils.py
```diff
@@ -81,3 +123,79 @@ def _handle_timeout(signum, frame):
...
+
+
+def get_skuba_configuration_dict(kubectl):
+    skubaConf_yml = kubectl.run_kubectl(
+        "get configmap skuba-config --namespace kube-system -o jsonpath='{.data.SkubaConfiguration}'"
+    )
+    return yaml.load(skubaConf_yml, Loader=yaml.FullLoader)
+
+
+def decrease_one_addon_manifest(addons_dict, skip=None):
```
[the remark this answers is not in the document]
Itxaka: I think it does, it just a shared util used in different tests. I could move it to a file called lib instead, but in the end in a shared, isolated function. Its specific? Yeah it is, but it could be reused in the future in other tests so having it centralized makes sense to me

Itxaka:
Maybe I didnt understand it correctly...

re-reading it... @davidcassany do you mean keeping non-destructive tests separated and run them as separated suites but leaving the other tests together and run them under one job instead?

Maybe Im not even explaining myself correctly, not feeling that good today...

Itxaka:
This should be merged after https://github.com/SUSE/skuba/pull/990 as it will require a rebase due to the name change

ci/jenkins/jobs.yaml
```diff
@@ -34,20 +34,20 @@
...
-        - 'test_hard_reboot'
+        - 'test_node_reboot'
         - 'test_upgrade_plan_all_fine'
         - 'test_upgrade_plan_from_previous'
         - 'test_upgrade_plan_from_previous_with_upgraded_control_plane'
         - 'test_upgrade_apply_all_fine'
         - 'test_upgrade_apply_from_previous'
         - 'test_upgrade_apply_user_lock'
         - 'test_cilium'
-        - 'test_deployments'
```
pablochacin: Curious: why make this change? The test is named test_nginx_deployment, but if I understand correctly, the suite will be executed, so the name of the suite can be still the same.

Itxaka: just to make clear what it contains I guess? The suite being called `deployments` makes me feel like it tests several things while its just testing deploying nginx.

Could probably be more generic if we expand the test to deploy other types of items to test different types of deployments (sidecards in the future? pvcs?)

PR #998, opened 2020-03-13
CI: Silence pip actions on PR jobs

Itxaka:
There is no reason to fill the logs with debug logs from pip
which are totally useless for us. This removes the verbose
option from pip installs to clean the log

## Why is this PR needed?

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
+++ b/ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
@@ -16,7 +16,6 @@ pipeline {
         PR_CONTEXT = 'jenkins/skuba-test'
         PR_MANAGER = 'ci/jenkins/pipelines/prs/helpers/pr-manager'
         REQUESTS_CA_BUNDLE = '/var/lib/ca-certificates/ca-bundle.pem'
-        PIP_VERBOSE = 'true'
     }
 
     stages {


ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
```diff
@@ -16,7 +16,6 @@ pipeline {
         PR_CONTEXT = 'jenkins/skuba-test'
         PR_MANAGER = 'ci/jenkins/pipelines/prs/helpers/pr-manager'
         REQUESTS_CA_BUNDLE = '/var/lib/ca-certificates/ca-bundle.pem'
-        PIP_VERBOSE = 'true'
```
evrardjp: I think this is fine. I am not sure we need to see installation of pip requirements in the job.

Itxaka:
> I'd like just to suggest a nit change. I'd prefer having PIP_VERBOSE = 'false' rather than nothing, mostly because this way it is explicit by anyone looking at the job that this env variable is available and taken into account.

So the reason I didnt do that its because by default is false and other pipelines do not even mention the var...making it more in sync with the others.

PR #1000, opened 2020-03-17
CI: wait for sonobuoy to finish tests 

Itxaka:
Instead of running the tests and checking every 2 minutes we
can instead just launch the command and make it wait for the
test finish.

This makes our wrapper more clean by removing uneeded lines
and avoids having a complex recheck or wait when running the
tests. Timeouts are provided via jenkins so it should not
exceed the normal times in case of the command hanging

## Why is this PR needed?

Sonobuoy tests are failing

Fixes https://github.com/SUSE/avant-garde/issues/1367

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/tasks/sonobuoy_e2e.py
+++ b/ci/tasks/sonobuoy_e2e.py
@@ -30,42 +30,29 @@ class SonobuoyE2eTests:
         logger.info(delete)
 
     def collect_results(self, retries, sonobuoy_args):
-        error = None
         attempts = 0
 
-        if 'Sonobuoy has completed' in self._get_status():
-            results = None
+        results = None
 
-            while retries > attempts:
-                try:
-                    attempts += 1
-                    logger.info(f'Attempting to retrieve the results {attempts}/{retries}')
-                    results = self._sonobuoy('retrieve results' + ' '.join(sonobuoy_args))
+        while retries > attempts:
+            try:
+                attempts += 1
+                logger.info(f'Attempting to retrieve the results {attempts}/{retries}')
+                results = self._sonobuoy('retrieve results' + ' '.join(sonobuoy_args))
 
-                except SonobuoyE2eTestsError:
-                    if retries == attempts:
-                        error = 'Could not retrieve sonobuoy results'
-                        break
-                    time.sleep(self.default_sleep)
-                else:
-                    break
+            except SonobuoyE2eTestsError:
+                if retries == attempts:
+                    raise SonobuoyE2eTestsError('Could not retrieve sonobuoy results')
+                time.sleep(self.default_sleep)
 
-            self._extract_results(results.strip())
-        else:
-            error = 'Sonobuoy e2e tests failed'
+        self._extract_results(results.strip())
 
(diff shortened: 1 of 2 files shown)


kravciak:
Maybe you could just use `--wait` instead of retrying, see https://gitlab.suse.de/mkravec/scripts/-/blob/master/tests/sonobuoy.sh#L5

Itxaka:
thanks !@kravciak  I was actually trying to reproduce the issue from jenkins and wanted several retries in there to understand why it was failing (temporal issues, wrong command) althougth I dont think a retry will hurt for those cases it seems like the actual issue is that the containers are removed when the test ends? So the status check fails because there are no pods!

In that case, I think the wait option is the proper one, instead of just checking to see if its finished, lets just wait for it to be done and use the step timeout on jenkins to avoid it getting out of hand. Otherwise its gonna be a terrible mess of coded to check if it finished,

ci/tasks/sonobuoy_e2e.py
```diff
@@ -106,45 +91,7 @@ def _sonobuoy(self, sonobuoy_args):
...
-                last_check = run_time
-                if 'Sonobuoy is still running' not in self._get_status():
-                    break
-
-            time.sleep(self.default_sleep)
-            run_time = int(time.time() - start_time)
-
-        if 'Sonobuoy is still running' in self._get_status():
-            raise SonobuoyE2eTestsError('Sonobuoy e2e tests ran out of time')
+        self._sonobuoy('run ' + ' '.join(sonobuoy_args) + '--wait')
```
kravciak: Can you add `--mode=certified-conformance` arg? Otherwise it's running in non-disruptive mode, which I think we don't need.
https://gitlab.suse.de/mkravec/scripts/-/blob/master/tests/sonobuoy.sh#L5

Itxaka: out of scope for this card to be fair. Im all for it if it brings an improvement but it should be either discussed and a card created or a bug opened.

I think this is more like a bug innit? should run in certified confirmance but its not doing so.

ci/tasks/sonobuoy_e2e.py
```diff
@@ -30,43 +30,33 @@ def cleanup_cluster(self, sonobuoy_args):
...
+
+            except SonobuoyE2eTestsError:
+                if retries == attempts:
+                    error = 'Could not retrieve sonobuoy results'
+                    break
+                time.sleep(self.default_sleep)
+            else:
+                break
 
-        if error is not None:
```
pablochacin: why are you removing this check? Does it make sense to try to extract results if there was an error collecting them?

Itxaka: Umm, the error part is not used anymore but I see that we are no longer raising an error if we could not retrieve the results.

let me rework this

PR #997, opened 2020-03-13
Require go version 1.13

cmurphy:
Epic: https://github.com/SUSE/avant-garde/issues/1416

Kubernetes 1.17 requires go version 1.13[1]. This change updates the
makefile version check and the package spec file template to require go
1.13 and updates the module list with `go mod tidy`.

[1] https://kubernetes.io/docs/setup/release/notes/#dependencies

## Why is this PR needed?

Does it fix an issue? addresses a business case?

add a description and link to the issue if one exists.

Fixes https://github.com/SUSE/avant-garde/issues/1576

**Reminder**: Add the "fixes bsc#XXXX" to the title of the commit so that it will
appear in the changelog.

## What does this PR do?

please include a brief "management" technical overview (details are in the code)

## Anything else a reviewer needs to know?

Special test cases, manual steps, links to resources or anything else that could be helpful to the reviewer.

## Info for QA

This is info for QA so that they can validate this. This is **mandatory** if this PR fixes a bug.
If this is a new feature, a good description in "What does this PR do" may be enough.

### Related info

Info that can be relevant for QA:
* link to other PRs that should be merged together
* link to packages that should be released together
* upstream issues

### Status **BEFORE** applying the patch

How can we reproduce the issue? How can we see this issue? Please provide the steps and the prove
this issue is not fixed.

### Status **AFTER** applying the patch

How can we validate this issue is fixed? Please provide the steps and the prove this issue is fixed.

## Docs

If docs need to be updated, please add a link to a PR to https://github.com/SUSE/doc-caasp.
At the time of creating the issue, this PR can be work in progress (set its title to [WIP]),
but the documentation needs to be finalized before the PR can be merged. 

# Merge restrictions

(Please do not edit this)

We are in *v4-maintenance phase*, so we will restrict what can be merged to prevent unexpected surprises:

    What can be merged (merge criteria):
        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine
        there is a PR for updating documentation (or a statement that this is not needed)

PR #1003, opened 2020-03-17
Add code ownership for ci to SUSE/caasp-platform/automation group

jordimassaguerpla:
Following the decision of having this group own

github/SUSE/caasp-automation

I would like to have the same ownership in the ci directory and then we
make it a requirement to have an approval of this group.

This is to prevent merging code into the CI without the CI maintainers
knowledge.

### DIFF
--- a/ci/CODEOWNERS
+++ b/ci/CODEOWNERS
@@ -1,0 +1,2 @@
+@SUSE/caasp-platform-automation
+


jordimassaguerpla:
@kkaempf : I would like to do this change to prevent issues like the one we have yesterday https://github.com/SUSE/skuba/pull/997#issuecomment-600007936 .

This change was installing go1.13 to all the ci workers. It is better if a member of the automation-groups is a reviewer so it can see this.

evrardjp:
I like the idea.

jordimassaguerpla:
> I like the idea. This can be extended if necessary.

Can I get your approval :) ?

aojea:
we should start to use CODEOWNERS more
:clap:

PR #1008, opened 2020-03-19
CI: Use configurable branch for sonobuoy job

Itxaka:
Looks like we were hardcoding the branch value on the job template
for sonobuoy, but what we really want is to have the default
branch parameter be master and allow setting the branch to pick
the jenkinsfile from set to that value dynamically so we
can test changes in the jenkinsfile from PRs.

## Why is this PR needed?

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/jenkins/templates/conformance-template.yaml
+++ b/ci/jenkins/templates/conformance-template.yaml
@@ -3,7 +3,6 @@
     project-type: pipeline
     number-to-keep: 30
     days-to-keep: 30
-    branch: master
     wrappers:
       - timeout:
           timeout: 180
@@ -13,7 +12,7 @@
     parameters:
         - string:
             name: BRANCH
-            default: '{branch}'
+            default: master
             description: The branch to checkout
         - string:
               name: PLATFORM
@@ -25,7 +24,7 @@
                 url: 'https://github.com/{repo-owner}/{repo-name}.git'
                 credentials-id: '{repo-credentials}'
                 branches:
-                    - '{branch}'
+                    - $branch
                 browser: auto
                 suppress-automatic-scm-triggering: true
                 basedir: skuba
--- a/ci/jenkins/templates/conformance-template.yaml
+++ b/ci/jenkins/templates/conformance-template.yaml
@@ -24,8 +24,13 @@
                 url: 'https://github.com/{repo-owner}/{repo-name}.git'
                 credentials-id: '{repo-credentials}'
                 branches:
-                    - $branch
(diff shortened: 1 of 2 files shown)


ci/jenkins/templates/conformance-template.yaml
```diff
@@ -25,8 +24,13 @@
                 url: 'https://github.com/{repo-owner}/{repo-name}.git'
                 credentials-id: '{repo-credentials}'
                 branches:
-                    - '{branch}'
+                    - $branch
```
pablochacin: Shouldn't this be `${BRANCH}'?

ci/jenkins/templates/conformance-template.yaml
```diff
@@ -13,7 +12,7 @@
     parameters:
         - string:
             name: BRANCH
-            default: '{branch}'
+            default: master
```
pablochacin: not sure if this must have quotes: `"master`"

ci/jenkins/templates/conformance-template.yaml
Itxaka: if we use `${BRANCH}` then jjb tries to interpolate it and fails due to not finding the branch value anywhere. Using simply `$branch` make jjb ignore it and jenkins use the proper `BRANCH` parameter that we define above :D

ci/jenkins/templates/conformance-template.yaml
Itxaka: mmmh, good point, lets see what jjb prints

Itxaka: no need! jjb output looks good.

ci/jenkins/templates/conformance-template.yaml
aojea: IIRC for jjb you have to use double brackets on job templates if you want to embed bash variables 
https://docs.openstack.org/infra/jenkins-job-builder/definition.html#macro-notes
Also, I think that is recommended to default the variable always
`${{BRANCH:-}}`

Itxaka: nice, thanks for linking to the proper docs, I had a hard time finding that.

jordimassaguerpla:
Already addressed by https://github.com/SUSE/skuba/pull/1094

PR #1011, opened 2020-03-19
Add CI feedback questionaire to the pull request template

jordimassaguerpla:
so that users can give feedback

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -58,3 +58,10 @@ We are in *v4-maintenance phase*, so we will restrict what can be merged to prev
 
 <!-- Remember, if this is a work in progress please pre-append [WIP] to the title until you are ready! 
     If you can, please apply all applicable labels to help reviews out! -->
+
+
+## CI Feedback
+
+Is the CI being useful to you? Does it help preventing buggy code to be merged? Or do you think we should focus more on faster tests?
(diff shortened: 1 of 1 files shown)


pablochacin:
Not convinced of this. I mean, the PR template is already complex and asking for feedback while people in the middle of setting this up it's maybe not the best moment. My 2 cents

jordimassaguerpla:
> Not convinced of this. I mean, the PR template is already complex and asking for feedback while people in the middle of setting this up it's maybe not the best moment. My 2 cents

Where would you put it instead?

thanks

drpaneas:
> Where would you put it instead?

Not here, the PR template is not the place to gather such feedback. Schedule a meeting with the development team and ask them to cast their vote.

My thoughts:
The CI is testing far too many things per pull-request basis. You have integrations tests which talk to infrastructure, spinning VMs and installing software using all sorts of dependencies (network, filesystems, you name it). Unless you run it in a space devoid of matter, this is going to be unreliable. For the PR merge, I would recommend you to run only unit-tests and inject indirect dependencies using mocking.

Integration tests can be handled _outside_ of Github, since you are doing internal infra stuff, which is private anyways.

tldlr; github ci is for merging code. internal ci is for shipping the product.

jenting:
IMO, the PR template should not include a feedback URL.

PR #1020, opened 2020-03-23
Terraform libvirt lb fixes

flavio:
## Why is this PR needed?

This PR fixes the terraform-libvirt provisioning when the nodes of the cluster are provisioned in the same way a real SUSE customer would do. That means the nodes are registered against RMT/SCC to get SLE-HA, CaaSP and the containers module packages.

## What does this PR do?

Two variables related with load balancer was already available but were not used at all by the terraform states.

With this PR it will be possible to:

  * Prevent the creation of load balancer nodes (great for small scale testing)
  * Successfully provision a cluster that is using one or more load balancer

Nothing really special to handle.

## Info for QA

Without this PR it's simply impossible to provision a CaaSP cluster using terraform-libvirt by having access only to the same set of resources as our customers do. Right now the CI system is using the terraform-libvirt provider to run our tests, but it's consuming packages from internal private repositories.

This PR allows to provision a cluster by using an official SLE 15 SP1 image and by enabling the CaaSP and SLE-HA products (the latter one is required to have HAProxy).

It's impossible to:
  * Create a cluster with 0 load balancer, even when the dedicated terraform variable is set to 0 there's always 1 load balancer used
  * Successfully run `terraform apply` when one or more load balancers are created. That happens because the LB node fails provisioning because there's no source to get the `haproxy` package. That happens because the node is not registered against *any* SLE repository.

You can deploy a cluster with 0, 1 or more LB nodes.

## Docs

Nothing to do, this is not an officially supported deployment mechanism.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


ci/infra/libvirt/lb-instances.tf
```diff
@@ -94,12 +127,15 @@ resource "libvirt_domain" "lb" {
...
+    volume_id = element(
+      libvirt_volume.lb.*.id,
+      count.index,
+    )
   }
 
   network_interface {
     network_id     = libvirt_network.network.id
-    hostname       = "${var.stack_name}-lb"
+    hostname       = "${var.stack_name}-lb-${count.index}"
```
pablochacin: This will break `testrunner`'s logic for retrieving the lb's ip address:

https://github.com/SUSE/skuba/blob/f6fd30f300904f06262e3bcfc8d41f6b3b43c95e/ci/infra/testrunner/platforms/terraform.py#L77-L82

flavio: How many LB do we need? I think 1 would be good enough... if that's the case I can just change the PR to allow the creation of 1 or 0 LB.

pablochacin: I'm not quite it would work with '0' LBs. Need to double-check.

pablochacin: No, it won't work. The existence of at least one LB was an assumption. 

https://github.com/SUSE/skuba/blob/f6fd30f300904f06262e3bcfc8d41f6b3b43c95e/ci/infra/testrunner/skuba/skuba.py#L51-L65

flavio: The default value is still 1. I can revert back the change partially to ensure the name of the LB doesn't change.

It's still useful to allow the creation of 0 LB when doing small cluster deployments.

evrardjp: Agreed on the 0 LB.

flavio:
@pablochacin, @evrardjp I've updated the PR to allow the creation of 1 or 0 LB. By default 1 LB will be created, the name will be very same one we had before.

I also preserved the output produced by terraform, for example (with `flavio-caasp` as stack name):

```ip_load_balancer = { "flavio-caasp-lb" = "172.30.1.0" }
ip_masters = {
  "flavio-caasp-master-0" = "172.30.2.0"
}
ip_workers = {
  "flavio-caasp-worker-0" = "172.30.3.0"
}
```

ci/infra/libvirt/lb-instances.tf
flavio: I've updated the code to reflect that.

pablochacin:
@flavio I fail to understand why you are removing the load balancer. If not about the changes, but the rational.  

I'm concern about introducing some complexity for uses cases that we don't actually have. IIRC we are not shipping this libvirt to customers, are we? 

Considering we only support configurations with multiple masters, and LB is required. I don't see the point of having this "feature" which potentially breaks the tests.

evrardjp:
> I'm concern about introducing some complexity for uses cases that we don't actually have. IIRC we are not shipping this libvirt to customers, are we?

I thought we did, or that we'll do soon.

> Considering we only support configurations with multiple masters, and LB is required. I don't see the point of having this "feature" which potentially breaks the tests.

I don't see a reason to force the presence of a load balancer for "test cases" customer environments, which only run one master node. It forces them to have an extra machine, with a license on it, for no valid reason.

pablochacin:
> > I'm concern about introducing some complexity for uses cases that we don't actually have. IIRC we are not shipping this libvirt to customers, are we?
> 
> I thought we did, or that we'll do soon.
> 

No, for libvirt, we dont't.

> > Considering we only support configurations with multiple masters, and LB is required. I don't see the point of having this "feature" which potentially breaks the tests.
> 
> I don't see a reason to force the presence of a load balancer for "test cases" customer environments, which only run one master node. It forces them to have an extra machine, with a license on it, for no valid reason.

As per my previous comment, this seems not a valid argument, as we don't ship this to customers.

flavio:
We don't ship the terraform-libvirt deployment to our customers. That was intended to be used by developers and CI.

As a developer in some cases I want to spare the LB resources so that I can have more worker nodes on my "tiny" workstation.

The code changes introduced to handle the case of having 0 LB are really limited. The majority of the changes are about DRYing up the code (reducing the duplication between the LB cloud-init and the other ones) and making it possible to deploy a system by using official packages (the fact nobody ever noticed this was not possible kinda concerns me).

flavio:
Rebased the PR against master before doing the merge

flavio:
@pablochacin I think the failures are unrelated with this PR, they are happening on a VMware cluster.

Can I go ahead and merge this PR?

flavio:
The code didn't change since the last approval, the failures are unrelated with this PR.

Forcefully merging as discussed with @pablochacin

PR #1022, opened 2020-03-25
Fix: setup fixture cleanup fails

pablochacin:
## Why is this PR needed?

The setup fixture's cleanup tries to retrieve the logs from nodes.
If the platform could not be allocated, it fails preventing the
actual cluster cleanup to be executed.

Fixes https://github.com/SUSE/avant-garde/issues/1432

appear in the changelog.

## What does this PR do?

Ignores exceptions trying to retrieve logs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/tests/conftest.py
+++ b/ci/infra/testrunner/tests/conftest.py
@@ -94,8 +94,11 @@ def platform(conf, target):
 @pytest.fixture
 def setup(request, platform, skuba):
     def cleanup():
-        platform.gather_logs()
-        platform.cleanup()
+        # if platform was not allocated, gather_logs may fail. Ignore.
+        try:
+            platform.gather_logs()
+        finally:
+            platform.cleanup()
 
     request.addfinalizer(cleanup)
 


ci/infra/testrunner/tests/conftest.py
```diff
@@ -94,9 +94,13 @@ def platform(conf, target):
 @pytest.fixture
 def setup(request, platform, skuba):
     def cleanup():
-        platform.gather_logs()
+        # if platform was not allocated, gather_logs may fail. Ignore.
+        try:
+            platform.gather_logs()
+        except:
+            pass
```
davidcassany: Nit:

In general I'd be more in favor of structures like
```
try:
  platform.gather_logs()
finally:
  platform.cleanup()
```

This way we still have the chance of adding some exception handling being certain that the  `platform.cleanup` will be called. 

There is not much difference on this specific case, thus I am also fine with the current code if you prefer it.

davidcassany:
All tests green, this minimalist change solving issues on OpenStack clearing resources (causing issues to QA) and strictly tied to CI. Also needed to ensure that openstack e2e tests do not over eat resources that QA needs. Merging

PR #1024, opened 2020-03-27
Migrate the usage of hyperkube

evrardjp:
For 1.18 onwards, we shouldn't have UseHyperKube set to true.

Closes: SUSE/avant-garde#1431

PR #1026, opened 2020-03-30
Update for Kubernetes v1.18.0

mjura:
It is adding Kubernetes v1.18 and all necessary containers to skuba.

It depends on: https://github.com/SUSE/skuba/pull/1024

### DIFF
--- a/internal/pkg/skuba/kubernetes/versions.go
+++ b/internal/pkg/skuba/kubernetes/versions.go
@@ -81,6 +81,27 @@ type ClusterAddonsKnownVersions = func(clusterVersion *version.Version) AddonsVe
 
 var (
 	supportedVersions = KubernetesVersions{
+		"1.18.0": KubernetesVersion{
+			ComponentHostVersion: ComponentHostVersion{
+				KubeletVersion:          "1.18.0",
+				ContainerRuntimeVersion: "1.17.0",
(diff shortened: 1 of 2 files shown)


internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -81,6 +81,27 @@ type ClusterAddonsKnownVersions = func(clusterVersion *version.Version) AddonsVe
 
 var (
 	supportedVersions = KubernetesVersions{
+    "1.18.0": KubernetesVersion{
```
evrardjp: indentation seems off.

ci/infra/openstack/cpi.auto.tfvars
```diff
@@ -1,5 +1,5 @@
 # Enable CPI integration with OpenStack
-#cpi_enable = true
+# cpi_enable = true
```
evrardjp: Not sure this change needs to be in ;)

internal/pkg/skuba/kubernetes/versions.go
mjura: `go fmt`

mjura:
It will be continued in https://github.com/SUSE/skuba/pull/1028

PR #1031, opened 2020-04-01
Delete pods before trying to upgrade the crio package

ereslibre:
Make sure we delete all existing pods in the target node before we try
to upgrade the crio and kubelet packages.

As a result of this, we also need to make sure that whatever comes
after that requires the API Server to be up appropriately retries the
request until it succeeds or timeouts.

## Why is this PR needed?

Fixes: https://github.com/SUSE/avant-garde/issues/1451

**This will need a backport to the 4.2 branch**

/cc @saschagrunert

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/cri.go
+++ b/internal/pkg/skuba/deployments/ssh/cri.go
@@ -29,6 +29,8 @@ import (
 func init() {
 	stateMap["cri.configure"] = criConfigure
 	stateMap["cri.start"] = criStart
+	stateMap["cri.stop"] = criStop
+	stateMap["cri.wipe-pods"] = criWipePods
 }
 
 func criConfigure(t *Target, data interface{}) error {
@@ -60,5 +62,15 @@ func criConfigure(t *Target, data interface{}) error {
 
 func criStart(t *Target, data interface{}) error {
 	_, _, err := t.ssh("systemctl", "enable", "--now", "crio")
+	return err
+}
+
+func criStop(t *Target, data interface{}) error {
+	_, _, err := t.ssh("systemctl", "stop", "crio")
+	return err
+}
+
+func criWipePods(t *Target, data interface{}) error {
+	_, _, err := t.ssh("crictl", "rmp", "-fa")
 	return err
 }
--- a/internal/pkg/skuba/kubernetes/retry/retry.go
+++ b/internal/pkg/skuba/kubernetes/retry/retry.go
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2020 SUSE LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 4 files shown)


internal/pkg/skuba/kubernetes/retry/retry.go
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package retry
```
chentex, line 18: nice!

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -138,6 +141,16 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 	} else if err := target.Apply(nil, "kubeadm.upgrade.node"); err != nil {
 		return err
 	}
+	if nodeVersionInfoUpdate.HasMajorOrMinorUpdate() {
```
evrardjp, line 144: Curious: wouldn't an upgrade of crio happen already when running the kubernetes.install-node-pattern above?

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -62,3 +64,13 @@ func criStart(t *Target, data interface{}) error {
...
 	return err
 }
+
+func criStop(t *Target, data interface{}) error {
+	_, _, err := t.ssh("systemctl", "stop", "crio")
+	return err
+}
+
+func criWipePods(t *Target, data interface{}) error {
+	_, _, err := t.ssh("crictl", "rmp", "-fa")
```
evrardjp, line 74: This is very impactful, and is basically the core of the new behaviour we introduce here. Did we fully investigate its impact?

pkg/skuba/actions/node/upgrade/apply.go
ereslibre: That will install the transient pattern that won't include kubelet or crio.

internal/pkg/skuba/deployments/ssh/cri.go
ereslibre: It's the only solution we have at the moment. Any workload that is formed only by one pod will have downtime.

ereslibre:
Another possibility would be to release Kubernetes 1.17 with cri-o 1.16 at this time, hence the `do-not-merge` label.

We have to double check the implications this would have. But we have two ways: either we move to cri-o 1.17 with this patch, or we keep cri-o 1.16 without this patch.

ereslibre:
I think it would be good to have @saschagrunert opinion on this before we take a final decision.

pkg/skuba/actions/node/upgrade/apply.go
evrardjp: Awesome. I thought it was the definitive already, my bad. Good then!

saschagrunert:
> I think it would be good to have @saschagrunert's opinion on this before we take a final decision.

I think it could work between those two versions, but I'd like to point out that nobody has tested this scenario. So if we run the usual e2e / conformance tests and everything looks green then this would be an option, too.

evrardjp:
I didn't run conformance testing when I first did my change into versions. So it indeed needs deep digging to ensure everything would work fine. But it's probably less impactful than the deletion of pods in a non-parallel fashion.

evrardjp:
I think we can abandon this for now. The upgrade from 1.16 to 1.17 seems to work just fine while staying at cri-o 1.16.

ereslibre:
Let's close this for now. Thanks to everyone involved.

PR #1033, opened 2020-04-01
Delete pods before trying to upgrade the crio package

ereslibre:
Backport of https://github.com/SUSE/skuba/pull/1031

Make sure we delete all existing pods in the target node before we try
to upgrade the crio and kubelet packages.

As a result of this, we also need to make sure that whatever comes
after that requires the API Server to be up appropriately retries the
request until it succeeds or timeouts.

## Why is this PR needed?

Fixes: https://github.com/SUSE/avant-garde/issues/1451

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/kubernetes.go
+++ b/internal/pkg/skuba/deployments/ssh/kubernetes.go
@@ -40,7 +40,8 @@ func init() {
 	stateMap["kubernetes.bootstrap.upload-secrets"] = kubernetesUploadSecrets(KubernetesUploadSecretsContinueOnError)
 	stateMap["kubernetes.join.upload-secrets"] = kubernetesUploadSecrets(KubernetesUploadSecretsFailOnError)
 	stateMap["kubernetes.install-node-pattern"] = kubernetesInstallNodePattern
-	stateMap["kubernetes.restart-services"] = kubernetesRestartServices
+	stateMap["kubernetes.start-kubelet"] = kubernetesStartKubelet
+	stateMap["kubernetes.stop-kubelet"] = kubernetesStopKubelet
 }
 
 func kubernetesUploadSecrets(errorHandling KubernetesUploadSecretsErrorBehavior) Runner {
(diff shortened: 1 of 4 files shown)


ereslibre:
Copying from https://github.com/SUSE/skuba/pull/1031:

Another possibility would be to release Kubernetes 1.17 with cri-o 1.16 at this time, hence the do-not-merge label.

We have to double check the implications this would have. But we have two ways: either we move to cri-o 1.17 with this patch, or we keep cri-o 1.16 without this patch.

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -77,7 +78,12 @@ func kubernetesInstallNodePattern(t *Target, data interface{}) error {
...
 
-func kubernetesRestartServices(t *Target, data interface{}) error {
-	_, _, err := t.ssh("systemctl", "restart", "crio", "kubelet")
+func kubernetesStartKubelet(t *Target, data interface{}) error {
+	_, _, err := t.ssh("systemctl", "enable", "--now", "kubelet")
+	return err
+}
+
+func kubernetesStopKubelet(t *Target, data interface{}) error {
+	_, _, err := t.ssh("systemctl", "stop", "kubelet")
```
jenting, line 87: Maybe it's better put this to `internal/pkg/skuba/deployments/ssh/kubelet.go`.

ereslibre: You are absolutely right! Let's first see if we need this PR at all (if we go with cri-o 1.16 we don't need it) -- if we do, will update, forgot about `ssh/kubelet.go` completely :)

dannysauer:
It's been a couple of weeks.  Did we get the "do not merge" figured out yet?  Probably oughtta decide if this is going or not so we can finally get new k8s pointed in the direction of the door. :)

If so, the commit message is 53 characters.  Here are a couple of sub-50 char suggestions to make the PR author check happy. :)
`Delete pods before trying to upgrade crio package`
`Delete pods before upgrading the crio package`

The code lint is failing due to insertion of `k8s.io/client-go/kubernetes/fake` in modules.txt.  I thought that was fixed, but I can't find the reference where.  Hopefully someone remembers.

I'm not sure about the other two CI errors.  Looks like terraform is sad.

ereslibre:
Let's close this for now. Thanks to everyone involved. Check https://github.com/cri-o/cri-o/pull/3545 and related issues.

PR #1039, opened 2020-04-06
Allow support for trustworthy JWTs in the API Server

saschagrunert:
## Why is this PR needed?

Applications like Istio with enabled Secure Gateways / Secret Discovery
Service (SDS) rely on trustworthy JWT's. Those can be enabled by the API
Server flags `--service-account-issue` and
`--service-account-signing-key-file`.

To achieve that we now change the `kubeadm-init.conf` to enable those
flags per default.

## What does this PR do?

Enable the mentioned kube-apiserver flags.

Kubeflows internal Istio deployment relies on that for example.

## Info for QA

None

### Related info

None

Trustworthy JWTs are not enabled.

Trustworthy JWTs work as expected.

## Docs

None, or do we have to document that somewhere?

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubeadm/configmap.go
+++ b/internal/pkg/skuba/kubeadm/configmap.go
@@ -124,6 +124,7 @@ func RemoveAPIEndpointFromConfigMap(client clientset.Interface, node *corev1.Nod
 func UpdateClusterConfigurationWithClusterVersion(initCfg *kubeadmapi.InitConfiguration, clusterVersion *version.Version) {
 	setApiserverAdmissionPlugins(initCfg, clusterVersion)
 	setContainerImagesWithClusterVersion(initCfg, clusterVersion)
+	setApiserverArgs(initCfg)
 }
 
 func setApiserverAdmissionPlugins(initCfg *kubeadmapi.InitConfiguration, clusterVersion *version.Version) {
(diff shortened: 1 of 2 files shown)


jenting:
This works for greenfield cluster, but how to deals existing brownfield cluster?

saschagrunert:
> This works for greenfield cluster, but how to deals existing brownfield cluster?

Do we migrate API server configs on upgrade?

saschagrunert:
> Thank you @saschagrunert. As @jenting comments this will only touch the init.
> 
> We have https://github.com/suse/skuba/blob/fba022ccf1cecfa110dff5b8e3565dbba41830db/internal/pkg/skuba/kubeadm/configmap.go#L122-L127, that will allow us to mutate the configuration both on init and upgrade.

Thank you for the hint, I updated the PR to change those parameters too.

pablochacin:
@saschagrunert It seams you should rebase your PR because you are using an outdated version of the PR pipeline.

saschagrunert:
Thanks for the hint @pablochacin, I rebased on top of the latest master branch

saschagrunert:
Looks like that I’m not a valid PR author. 😢

jenting:
@ereslibre PTAL :bow:

PR #1046, opened 2020-04-15
FIX: check node timeout set to None

pablochacin:
## Why is this PR needed?

If the caller doesn't have a sensible default for timeout and passes None,
the default value is overridden. 

Fixes https://github.com/SUSE/avant-garde/issues/1489

## What does this PR do?

As the function is called from multiple points, a safety check is added to ensure sensible default are enforced. Same applies for backoff

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/checks/checks.py
+++ b/ci/infra/testrunner/checks/checks.py
@@ -73,6 +73,13 @@ class Checker:
 
 
     def check_node(self, role, node, checks=None, stage=None, timeout=180, backoff=20):
+
+        #Prevent defaults to be accidentally overridden by callers with None
+        if timeout is None:
+            timeout = 180
+        if backoff is None:
+            backoff = 20
+
         if checks:
             check_names = checks
             checks = []


ci/infra/testrunner/checks/checks.py
```diff
@@ -73,6 +73,13 @@ def __init__(self, conf, platform):
 
 
     def check_node(self, role, node, checks=None, stage=None, timeout=180, backoff=20):
+
+        #Prevent defaults to be accidentally overridden by callers with None
```
davidcassany, line 77: I am curious to track where it happens. But the change seems sane to me.

pablochacin: In case of the bootstrap fixture it calls skuba node_bootstrap without a timeout which then defaults to None. In turn, this function calls check_node passing its timeout (None).

The issue is hard to solve because either we provide a sensible default for every function that calls check_node or those functions should not pass the timeout unless they are provided a valid value. I plaayed with both options but neither one is safe. There's a better option I would probably implement which is a wrapper that removed parameters with value None if they have a default value. Stay tuned.

PR #1054, opened 2020-04-21
Use an already existing virtual network

davidcassany:
## Why is this PR needed?

This PR is needed to run multiple clusters over the same libvirt infraesctructure.  Part of SUSE/avant-garde#1297

## What does this PR do?

This PR enables to use an already existing virtual network configuration. So terraform does not require use a new selfcreated virtual network for each deployment.

Before merging this PR the `caasp-network` needs to be defined within the kvm host.

### DIFF
--- a/ci/infra/libvirt/lb-instances.tf
+++ b/ci/infra/libvirt/lb-instances.tf
@@ -134,7 +134,8 @@ resource "libvirt_domain" "lb" {
   }
 
   network_interface {
-    network_id     = libvirt_network.network.id
+    network_name   = var.network_name
+    network_id     = var.network_name == "" ? libvirt_network.network.0.id : null
     hostname       = "${var.stack_name}-lb"
     wait_for_lease = true
   }
--- a/ci/infra/libvirt/master-instance.tf
+++ b/ci/infra/libvirt/master-instance.tf
@@ -86,7 +86,8 @@ resource "libvirt_domain" "master" {
   }
 
   network_interface {
-    network_id     = libvirt_network.network.id
+    network_name   = var.network_name
+    network_id     = var.network_name == "" ? libvirt_network.network.0.id : null
     hostname       = "${var.stack_name}-master-${count.index}"
     wait_for_lease = true
   }
--- a/ci/infra/libvirt/network.tf
+++ b/ci/infra/libvirt/network.tf
@@ -1,4 +1,5 @@
 resource "libvirt_network" "network" {
+  count  = var.network_name == "" ? 1 : 0
   name   = "${var.stack_name}-network"
   mode   = var.network_mode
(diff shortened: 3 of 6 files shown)


ci/infra/libvirt/main.tf
```diff
@@ -1,5 +1,5 @@
 provider "libvirt" {
-  uri = var.libvirt_keyfile == "" ? var.libvirt_uri : "${var.libvirt_uri}?keyfile=${var.libvirt_keyfile}"
+  uri = var.libvirt_keyfile == null || var.libvirt_keyfile == "" ? var.libvirt_uri : "${var.libvirt_uri}?keyfile=${var.libvirt_keyfile}"
```
pablochacin, line 2: This is totally unrelated, right? I would prefer to move to another PR.

ci/infra/libvirt/cloud-init/common.tpl
```diff
@@ -42,7 +42,7 @@ bootcmd:
   - ip link set dev eth0 mtu 1400
   # Hostnames from DHCP - otherwise localhost will be used
   - /usr/bin/sed -ie "s#DHCLIENT_SET_HOSTNAME=\"no\"#DHCLIENT_SET_HOSTNAME=\"yes\"#" /etc/sysconfig/network/dhcp
-  - netconfig update -f
+  # - netconfig update -f
```
pablochacin, line 45: This is a test to address the zypper update failure, right? I think it shouldn't be part of this PR.

davidcassany:
Note PR only adds the conditional in to use a pre defined virtual network when the tf variable `network_name` is defined, otherwise it keeps creating its own network. This is the case for the current PR, not `network_name` is defined.

In case `"networ_name" : "caasp-network"` the libvirt server should have a network configured and active equivalent to:

```xml
<network>
  <name>caasp-network</name>
  <uuid>96f42967-53a0-431b-87a8-85aca7ab858c</uuid>
  <forward mode='nat'/>
  <bridge name='virbr1' stp='on' delay='0'/>
  <mac address='52:54:00:d8:52:2f'/>
  <domain name='caasp.local'/>
  <dns enable='yes'/>
  <ip family='ipv4' address='10.17.0.1' prefix='22'>
    <dhcp>
      <range start='10.17.0.2' end='10.17.3.254'/>
    </dhcp>
  </ip>
</network>
```

Currently there is a testing network in `kvm-ci.nue.caasp.suse.net` called `caasp-network` with the above configuration, the network is not active. I'd suggest do a coordinated one shot with the `caasp-network` to validate it and then revert back the  `network_name` to be undefined. This was we test both cases but merge a transparent change. This way in a followup oneline PR we can start using a predefined network when we have the server configured as we want.

PR #1055, opened 2020-04-21
cleanup kubelet.service file for disarm-kubulete

suseclee:
## Why is this PR needed?

During Skuba disarms kubelete, kubelet.service file was left over. 

## What does this PR do?

This PR cleans /etc/systemd/system/kubelet.service for disarming kubelet.

In the removed nodes from cluster, we should not see kubelet.service file.

## Info for QA

### Related info

* upstream issues

After `skuba node remove worker-2`, we see kubelet.service left from the worker-2.

## Docs

Do not need to change any Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


evrardjp:
Also, (not a systemd expert), but wouldn't removing files around need a `daemon-reload` ?

suseclee:
disarm-kubelet seems to mask the kubelet.service purposely.

PR #1060, opened 2020-04-24
Add provision and deploy stages to e2e tests

pablochacin:
## Why is this PR needed?

E2E tests provision and deploy clusters using fixtures. It makes difficult to detect situations when the provisioning fails due to infrastructure issues or the deployment fails due to issues with packages or images required by skuba.

## What does this PR do?

Separate the provision and deploying from e2e tests and adds cluster sanity tests.
Adds the skip setup option to e2e test target to prevent tests to deploy a cluster using fixtures.

TODO:
- [ ] Handle the special case of upgrade tests, which must deploy the cluster specifying the kubernetes version

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/Makefile
+++ b/ci/Makefile
@@ -7,13 +7,16 @@ JENKINS_JOBS_DIR = ${MKFILE_DIR}/jenkins
 JENKINS_JOB_CONFIG ?= ${JENKINS_JOBS_DIR}/jenkins_jobs.ini
 PLATFORM ?= openstack
 ifdef SUITE
-SUITE_CMD=--suite ${SUITE}.py
+SUITE_OPT=--suite ${SUITE}.py
 endif
 ifdef TEST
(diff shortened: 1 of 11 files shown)


pablochacin:
@davidcassany I've been working on this PR but as mentioned in the description, there's an issue with the upgrade tests. They are special in the sense that the test itself must deploy skuba specifying a version. Therefore, the cluster cannot be deployed in the pipeline before the tests.

Some options are:
- [ ] Add logic to the pipeline to skip the deployment phase for some tests. Makes the pipeline logic complex.
- [ ] Create a separate pipeline for update tests. It is not clear how to make this tests still part of the e2e test project. Maybe requires creating another project (similar to the conformance project).

PR #1061, opened 2020-04-24
Update crio CaaSP configuration for v1.18.0

mjura:
## Why is this PR needed?

For cri-o `v1.18.0` we would like stop using sysconfig crio configuration and use more appropriate way `/etc/crio/crio.conf.d/`

## What does this PR do?

It is installing cri-o CaaSP configuration in `/etc/crio/crio.conf.d/01-caasp.conf`
For previous cri-o versions we have to keep backwards compatibility.

To upgrade cluster from previous versions 

```
skuba cluster init --control-plane <PUBLIC_IP> /tmp/test-cluster
mv /tmp/test-cluster/addons/cri/conf.d <CURRENT-CLUSTER-NAME>/addons/cri/
rm <CURRENT-CLUSTER-NAME>/addons/cri/default_flags
```

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/pkg/skuba/actions/cluster/init/init.go
+++ b/pkg/skuba/actions/cluster/init/init.go
@@ -115,7 +115,12 @@ func Init(initConfiguration InitConfiguration) error {
 		return errors.Errorf("cluster configuration directory %q already exists", initConfiguration.ClusterName)
 	}
 
-	scaffoldFilesToWrite := scaffoldFiles
+	scaffoldFilesToWrite := criScaffoldFiles["criconfig"]
+	kubernetesVersion := initConfiguration.KubernetesVersion
+	if kubernetesVersion.Minor() < 18 {
+		scaffoldFilesToWrite = criScaffoldFiles["sysconfig"]
+	}
+
 	if len(initConfiguration.CloudProvider) > 0 {
 		if cloudScaffoldFiles, found := cloudScaffoldFiles[initConfiguration.CloudProvider]; found {
 			scaffoldFilesToWrite = append(scaffoldFilesToWrite, cloudScaffoldFiles...)
--- a/pkg/skuba/actions/node/bootstrap/bootstrap.go
+++ b/pkg/skuba/actions/node/bootstrap/bootstrap.go
@@ -118,9 +118,11 @@ func coreBootstrap(initConfiguration *kubeadmapi.InitConfiguration, bootstrapCon
 		return errors.Wrap(err, "error writing init configuration")
 	}
 
-	var criConfigure string
-	if _, err := os.Stat(skuba.CriDockerDefaultsConfFile()); err == nil {
-		criConfigure = "cri.configure"
+	var criSetup string
+	if _, err := os.Stat(skuba.CriDefaultsConfFile()); err == nil {
+		criSetup = "cri.configure"
+	} else if _, err := os.Stat(skuba.CriDockerDefaultsConfFile()); err == nil {
+		criSetup = "cri.sysconfig"
 	}
 
 	// bsc#1155810: generate cluster-wide kubelet root certificate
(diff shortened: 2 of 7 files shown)


pkg/skuba/actions/cluster/init/init.go
```diff
@@ -115,7 +115,12 @@ func Init(initConfiguration InitConfiguration) error {
 		return errors.Errorf("cluster configuration directory %q already exists", initConfiguration.ClusterName)
 	}
 
-	scaffoldFilesToWrite := scaffoldFiles
+	scaffoldFilesToWrite := criScaffoldFiles["criconfig"]
```
evrardjp, line 118: that supposes the criconfig is the same for < 18 and >= 18.
Yet criDefaultsConf file is different, it contains more things.

Could this be an "else"?

pkg/skuba/actions/node/bootstrap/bootstrap.go
```diff
@@ -118,9 +118,11 @@ func coreBootstrap(initConfiguration *kubeadmapi.InitConfiguration, bootstrapCon
 		return errors.Wrap(err, "error writing init configuration")
 	}
 
-	var criConfigure string
-	if _, err := os.Stat(skuba.CriDockerDefaultsConfFile()); err == nil {
-		criConfigure = "cri.configure"
+	var criSetup string
+	if _, err := os.Stat(skuba.CriDefaultsConfFile()); err == nil {
```
evrardjp, line 122: it makes me sad to have this kind of wiring tbh, but I don't see a different approach possible, as we don't have a cluster-init cleaner, and we don't send the files smartly. LGTM.

mjura:
> Is there anything in the migration of files existing on the cluster during an upgrade?
Upgrade works fine, once user will provide cri-o `01-caasp.conf` or it will generate it with skuba. Instruction as follows 
```
skuba cluster init --control-plane <PUBLIC_IP> /tmp/test-cluster
mv /tmp/test-cluster/addons/cri/conf.d <CURRENT-CLUSTER-NAME>/addons/cri/
rm <CURRENT-CLUSTER-NAME>/addons/cri/default_flags
```
it will be added to release notes and documented here https://github.com/SUSE/avant-garde/issues/1420

> How do we remove the sysconfig files on the destination nodes?
It is overwritten by cri-o v1.18.0 sysconfig from the package. 

> How do we deal with the updated files as users won't go through a bootstrap/init/join ?
Users don't have to go through `bootstrap/init/join`, they need to only 
```
mkdir <CURRENT-CLUSTER-NAME>/addons/cri/conf.d
```
write `01-caasp.conf` file or generate it by 
```
skuba cluster init --control-plane <PUBLIC_IP> /tmp/test-cluster
mv /tmp/test-cluster/addons/cri/conf.d <CURRENT-CLUSTER-NAME>/addons/cri/
```
and the do `skuba node upgrade ...`

pkg/skuba/actions/cluster/init/init.go
mjura: We would like to keep this as default and drop `if` condition in CaaSP 5.0 `else` will cause branch prediction jump which is expensive in coding.

evrardjp: But this crio configuration needs to be different per version, doesn't it?

evrardjp:
I have renamed the PR, so it i's clearer in terms of scope.

pkg/skuba/actions/cluster/init/init.go
mjura: Nope, this is change regarding format configuration. We don't change it

mjura:
@mjura Thank you

PR #1066, opened 2020-04-27
Use Leap 15.1 for builder image


evrardjp:
The integration tests are using a builder image
running tumbleweed, which could be breaking at any
time. I suspect that the real tests are merely
using the builder data artifacts, and apply those
in the proper images.

In other words, it means that the builder does
not need to be a moving target (as it is just
a tooling). For stability purposes, I am moving
the image to a more stable image, based on Leap.

### DIFF
--- a/skuba-update/test/os/suse/Dockerfile.builder
+++ b/skuba-update/test/os/suse/Dockerfile.builder
@@ -1,4 +1,4 @@
-FROM registry.opensuse.org/opensuse/tumbleweed
+FROM registry.opensuse.org/opensuse/leap:15.1
 
 RUN zypper ref && zypper -n in rpm-build rpmdevtools createrepo libcreaterepo_c-devel
 RUN rm /var/run/reboot-needed


evrardjp:
Unrelated test failure.

evrardjp:
skuba-update acceptance is now passing though. Hurray.

skuba-update/test/os/suse/Dockerfile.builder
```diff
@@ -1,4 +1,4 @@
-FROM registry.opensuse.org/opensuse/tumbleweed
+FROM registry.opensuse.org/opensuse/leap:15.1
```
mjura, line 1: Maybe we should skip to `Leap 15.2` ?

evrardjp: Proposed here: https://github.com/SUSE/skuba/pull/1075

evrardjp:
Follow up https://github.com/SUSE/skuba/pull/1075

PR #1067, opened 2020-04-27
Refactor node reboot test

pablochacin:
## Why is this PR needed?

Improve robustness of node reboot test and make failures to generate output easier to debug.

Fixes https://github.com/SUSE/avant-garde/issues/1529

## What does this PR do?

Use node_is_ready wrapped in a wait function instead of blocking in a kubectl wait --condition.

Give a prudential time for the node to become non-ready (after the readiness probe's grace period) before waiting for it to get ready.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/tests/test_node_reboot.py
+++ b/ci/infra/testrunner/tests/test_node_reboot.py
@@ -17,12 +17,12 @@ def check_node_is_ready(platform, kubectl, role, nr):
 
 @pytest.mark.pr
 @pytest.mark.parametrize('role,node', [('master', 1), ('worker', 0)])
-def test_hard_reboot(deployment, platform, skuba, kubectl, role, node):
+def test_hard_reboot(deployment, platform, kubectl, role, node):
     """ Reboots master and worker nodes and checks they are back ready.
     For masters, reboot master 1, as master 0 is expected to be the
     cluster leader and rebooting it may introduce transient etcd errors.
     """
-    assert skuba.num_of_nodes(role) > node
+    assert kubectl.get_num_nodes_by_role(role) > node
 
     platform.ssh_run(role, node, 'sudo reboot &')
 
--- a/ci/infra/testrunner/tests/test_node_reboot.py
+++ b/ci/infra/testrunner/tests/test_node_reboot.py
@@ -4,58 +4,45 @@ import time
 
 import pytest
 
-from tests.utils import wait
+from tests.utils import (check_pods_ready, node_is_ready, wait)
 
 from timeout_decorator import timeout
 
 logger = logging.getLogger("testrunner")
 
 
-def check_pods_ready(kubectl):
-    pods = json.loads(kubectl.run_kubectl('get pods --all-namespaces -o json'))['items']
-    for pod in pods:
-        pod_status = pod['status']['phase']
-        pod_name   = pod['metadata']['name']
-        assert pod_status in ['Running', 'Completed'], f'Pod {pod_name} status {pod_status} != Running or Completed'
+def check_node_is_ready(platform, kubectl, role, nr):
+    assert node_is_ready(platform, kubectl, role, nr)
 
 
 @pytest.mark.pr
-@pytest.mark.parametrize('node_type,node_number', [('master', 1), ('worker', 0)])
-def test_hard_reboot(deployment, platform, skuba, kubectl, node_type, node_number):
-    assert skuba.num_of_nodes(node_type) > node_number
+@pytest.mark.parametrize('role,node', [('master', 1), ('worker', 0)])
+def test_hard_reboot(deployment, platform, skuba, kubectl, role, node):
+    """ Reboots master and worker nodes and checks they are back ready.
+    For masters, reboot master 1, as master 0 is expected to be the
+    cluster leader and rebooting it may introduce transient etcd errors.
+    """
+    assert skuba.num_of_nodes(role) > node
 
-    logger.info('Wait for all the nodes to be ready')
-    wait(kubectl.run_kubectl, 
-         'wait --for=condition=Ready node --all --timeout=0',
-         wait_timeout=10,
-         wait_backoff=30,
-         wait_elapsed=300,
-         wait_allow=(RuntimeError))
+    platform.ssh_run(role, node, 'sudo reboot &')
 
-    logger.info(f'Rebooting {node_type} {node_number}')
-
-    platform.ssh_run(node_type, node_number, 'sudo reboot &')
-
-    # wait the node to reboot
-    wait(platform.ssh_run,
-         node_type,
-         node_number,
-         'echo hello',
-         wait_backoff=30,
-         wait_timeout=30,
-         wait_elapsed=300,
-         wait_allow=(RuntimeError))
+    # Allow time for kubernetes to check node readiness and detect it is not
+    # ready
+    time.sleep(60)
 
     # wait the node to become ready
-    wait(kubectl.run_kubectl,
-         'wait --for=condition=Ready node --all --timeout=0',
+    wait(check_node_is_ready,
+         platform,
+         kubectl,
+         role,
+         node,
          wait_delay=30,
          wait_timeout=10,
          wait_backoff=30,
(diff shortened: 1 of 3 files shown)


ci/infra/testrunner/tests/test_node_reboot.py
```diff
@@ -4,58 +4,44 @@
...
+    assert node_is_ready(platform, kubectl, role, nr)
 
 
 @pytest.mark.pr
-@pytest.mark.parametrize('node_type,node_number', [('master', 1), ('worker', 0)])
-def test_hard_reboot(deployment, platform, skuba, kubectl, node_type, node_number):
-    assert skuba.num_of_nodes(node_type) > node_number
+@pytest.mark.parametrize('role,node', [('master', 1), ('worker', 0)])
+def test_hard_reboot(deployment, platform, skuba, kubectl, role, node):
+    """ Reboots master and worker nodes and checks the are back ready.
```
davidcassany, line 21: typo: the -> they

ci/infra/testrunner/tests/test_node_reboot.py
```diff
@@ -4,58 +4,44 @@
...
-    wait(platform.ssh_run,
-         node_type,
-         node_number,
-         'echo hello',
-         wait_backoff=30,
-         wait_timeout=30,
-         wait_elapsed=300,
-         wait_allow=(RuntimeError))
+    # Allow time for the node to reboot and kubernetes to check readiness
+    time.sleep(60)
```
davidcassany, line 30: It is not clear to me where do we wait for the node to boot. This sleep should be counted after the k8s daemon is running again. IMHO I'd wait for the node to be reachable again and then start counting the grace period.

Having an actively pulling call to check if the node is reachable can prevent failures if for some reason the node takes longer. We could have a longer timeout just for safety and it would not lead to longer run in most cases.

ci/infra/testrunner/tests/test_node_reboot.py
```diff
@@ -4,58 +4,44 @@
...
 
     # wait the node to become ready
-    wait(kubectl.run_kubectl,
-         'wait --for=condition=Ready node --all --timeout=0',
+    wait(check_node_is_ready,
```
davidcassany, line 33: couldn't this call fail if the network is not ready, assume this is only eating AssertionError exceptions. Do we get this sort of exception if the node is not reachable?

ci/infra/testrunner/tests/test_node_reboot.py
pablochacin: We are waiting here for k8s to notice the node has gone not ready before we check it is ready. If we just check after rebooting the node, k8s may not have noticed it went not ready and the test will pass. However, the node can later go not ready and never come back ready.

ci/infra/testrunner/tests/test_node_reboot.py
pablochacin: yeap, thanks.

ci/infra/testrunner/tests/test_node_reboot.py
pablochacin: No, it won't fail as we are calling the API server which is in HA setup.

pablochacin:
> It is not clear to me why to we do not have an active wait on the node to reboot. Assume that shutdown and reboot is going to happen always in less than 60s seems too brave to me, isn't the node check after that going to fail if the node is not even reachable?

Maybe there's a misunderstanding here. We are not expecting the reboot to happen in 60 seconds. We are waiting 60 seconds to be "sure" that k8s had time to notice the node went non ready. The node can take longer to reboot. We are giving it 5 minutes to reboot. I would say it is a fair amount of time.

ci/infra/testrunner/tests/test_node_reboot.py
davidcassany: Correct, forget my previous comment. I did a wrong assumption that the verification call for readiness was done against the specific node, however the call is against the cluster.

davidcassany:
On my local deployment if I make shutdown, just right after the shutdown I see this:

```
$ kubectl --kubeconfig=admin.conf get nodes caasp-master-1 -o jsonpath='{range @.status.conditions[*]}{@.type}={@.status};{end}'
Error from server (InternalError): an error on the server ("") has prevented the request from succeeding (get nodes caasp-master-1)
```

The node needs to be reachable for this call to succeed, I guess k8s internally does some check on the node host. The exception caused by this call (`Exception` type) is not catched by the wait and retry approach, hence it fails.

I also noticed that a calls like
```
kubectl --kubeconfig=admin.conf get nodes -o jsonpath='{.items[*].metadata.name}'
```
to get the number of nodes also fail during the reboot, until the node not again up and running it doesn't start to succeed. I did not manage to see `kubectl get nodes` to remove one node on reboot, it just fails.

pablochacin:
> The exception caused by this call (`Exception` type) is not catched by the wait and retry approach, hence it fails.

You mean `InternalError`, right? This can be easily added to the wait condition.

pablochacin:
> I also noticed that a calls like
> 
> ```
> kubectl --kubeconfig=admin.conf get nodes -o jsonpath='{.items[*].metadata.name}'
> ```
> 
> to get the number of nodes also fail during the reboot, until the node not again up and running it doesn't start to succeed. I did not manage to see `kubectl get nodes` to remove one node on reboot, it just fails.

This is strange and makes no sense. Unless you were removing the master that happened to be the cluster leader. It makes no sense that a failing node makes a call like this fail. The call for a specific node makes more sense as maybe the api server makes some validation (however, I doubt it)

davidcassany:
I think I just figure it out. I think the problem is that on my own local environment I was running the test with two masters only, instead of three. Having three nodes the above failures do no happen, I guess it requires some redundancy to actually work properly, even I am surprised this is that sensible for this basic queries.

I did again my local tests with three masters, I shutdown master 1 and kubectl reports as expected. However it is interesting to note that it takes some time to detect the node when down (thus the wait for the grace period is a must). Then once one master is down I also shutdown a second one, leaving just one master alive. At that point `kubectl` started to run the failures I reported above. After booting again one of the two stopped masters `kubectl` works as expected again.

ci/infra/testrunner/tests/test_node_reboot.py
```diff
@@ -4,58 +4,45 @@
...
 
 
 @pytest.mark.pr
-@pytest.mark.parametrize('node_type,node_number', [('master', 1), ('worker', 0)])
-def test_hard_reboot(deployment, platform, skuba, kubectl, node_type, node_number):
-    assert skuba.num_of_nodes(node_type) > node_number
+@pytest.mark.parametrize('role,node', [('master', 1), ('worker', 0)])
+def test_hard_reboot(deployment, platform, kubectl, role, node):
+    """ Reboots master and worker nodes and checks they are back ready.
+    For masters, don't master 1, as master 0 is expected to be the
```
davidcassany, line 22: Looks like this sentence has some copy & paste error. Probably it was intented to be `For masters, reboot master 1,...` :wink:

pablochacin: copy&paste considered harmful

pablochacin:
> Then once one master is down I also shutdown a second one, leaving just one master alive. At that point `kubectl` started to run the failures I reported above. After booting again one of the two stopped masters `kubectl` works as expected again

We should limit the test to rebooting one master as we are not testing etcd, but caasp. That is, we should only test viable k8s deployments. etcd requires at least `n/1+1` live masters (for example, 2 out of 3).

pablochacin:
> I believe that this is mergeable if get some consecutive successful runs.

I noticed the execution fails in a check that is made in a fixture before the test is started. I removed that check because it should not be executed if the cluster is already deployed. 

However, I cannot explain why the check is failing.

pablochacin:
@davidcassany I got three successful runs of the tests for the PR. I think the bootstrap test is more reliable than before. I suggest merging and monitoring in the following days.

PR #1068, opened 2020-04-28
Cleanup tox.ini


evrardjp:
We don't need all of this configuration.
This makes it more succint.

### DIFF
--- a/skuba-update/tox.ini
+++ b/skuba-update/tox.ini
@@ -1,58 +1,20 @@
 [tox]
+minversion = 2.5.0
 skip_missing_interpreters = True
+envlist = check, py38, py36, py34
 skipsdist = True
-envlist =
-    check,
-    py36,
-    py34,
 
 [testenv]
-whitelist_externals =
-    /bin/bash
-    /sbin/bash
-basepython =
-    check: python
-    py36: python3.6
-    py34: python3.4
-envdir =
-    check: {toxworkdir}/check
-    py36: {toxworkdir}/py36
-    py34: {toxworkdir}/py34
-passenv =
-    *
 usedevelop = True
-deps =
-    -rdev-requirements.txt
-
-[testenv:py36]
-skip_install = True
-usedevelop = True
+deps = -r{toxinidir}/dev-requirements.txt
 setenv =
-    PYTHONPATH={toxworkdir}/py36
-    PYTHONUNBUFFERED=yes
+    VIRTUAL_ENV={envdir}
+    PYTHONUNBUFFERED=1
     WITH_COVERAGE=yes
 passenv =
-    *
-deps = {[testenv]deps}
-changedir=test/unit
+    HOME
+    SKUBA_UPDATE_OPTIONS
 commands =
-    bash -c 'cd ../../ && ./setup.py develop'
-    py.test --cov=skuba_update --cov-report=term-missing \
-        --cov-fail-under=100 --cov-config .coveragerc
-
-[testenv:py34]
-skip_install = True
-usedevelop = True
-setenv =
-    PYTHONPATH={toxworkdir}/py34
-    PYTHONUNBUFFERED=yes
-    WITH_COVERAGE=yes
-passenv =
-    *
-deps = {[testenv]deps}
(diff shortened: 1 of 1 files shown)


evrardjp:
Acceptance test failure fix: https://github.com/SUSE/skuba/pull/1066

evrardjp:
Rebased.

skuba-update/tox.ini
```diff
@@ -1,71 +1,31 @@
 [tox]
+minversion = 2.5.0
 skip_missing_interpreters = True
```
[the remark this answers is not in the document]
evrardjp: Good catch, it's so obvious I _thought_ we had it, and we indeed don't have it.

evrardjp:
https://github.com/SUSE/skuba/pull/1169

skuba-update/tox.ini
```diff
@@ -1,71 +1,32 @@
...
-skipsdist = True
-envlist =
-    check,
-    py36,
-    py34,
+envlist = check, py38, py36, py34
+skipdist = True
```
cmurphy, line 5: This seems wrong, the old version `skipsdist` looks right https://tox.readthedocs.io/en/latest/config.html#conf-skipsdist

skuba-update/tox.ini
```diff
@@ -1,71 +1,32 @@
...
 # https://github.com/koalaman/shellcheck/wiki/SC1117 (too pedantic)
 # https://github.com/koalaman/shellcheck/wiki/SC1090 (shellcheck specific limitation)
 [testenv:check]
-deps = {[testenv]deps}
-skip_install = True
-usedevelop = True
-setenv =
-    PYTHONPATH={toxworkdir}/check
-    PYTHONUNBUFFERED=yes
+basepython = python3
```
cmurphy, line 26: Could probably move this to `[testenv]`?

Itxaka: if we do that, tox will raise a warning (for now![0]) as other envs will try to use whatever python3 is by default and resulting in a conflict between the basepython and the environment name.

So its better to use basepython to specify an environment default python rather than making it global.

[0] https://tox.readthedocs.io/en/latest/config.html#conf-ignore_basepython_conflict

evrardjp: Oh great, I now notice that my comment sent during github outage wasn't sent. But yeah, the tl;dr of what I said is in that message of itxaka.
What I wrote was something akin to:

This environment is using an external python tool, and therefore we should set the basepython to use the default host's python version. Which is why basepython is set to "python3" not a major.minor version.
For the other environments, version is inferred from the environment name.


I am now heading to my other comments I might have left during that downtime.

Itxaka (dismissed):

skuba-update/tox.ini
cmurphy: Good points, thanks!

cmurphy (dismissed):

tdaines42:
> i'm not tox expert, but from what I understand this looks good. @tdaines42 could you look over this?
> 
> Thanks for cleaning this up!

Sorry forgot to respond I took a look but tox really isn't my thing. Seems like people who know better have gotten to it

jordimassaguerpla:
@evrardjp  Do you have the issue number or bsc for this? If there is none, could you create one? We need it for planning the releases. Thanks.

evrardjp:
There is none, this is not a bug, just an improvement

jordimassaguerpla:
> There is none, this is not a bug, just an improvement

Could you create one in avant-garde? We use Issues for planning the release work and this will help us. Otherwise they can escape our radar. Thanks in advance.

evrardjp:
So much paperwork :)
Ok let's do it.

jordimassaguerpla:
> Done: [SUSE/avant-garde#1931](https://github.com/SUSE/avant-garde/issues/1931)

Thanks. This helps.

Itxaka:
upstream card was closed so I think this can be closed @chentex

evrardjp:
Not sure many folks cared anyway, let's close.

PR #1069, opened 2020-04-28
Add static version file for skuba-update


evrardjp:
skuba-update version is only correct when rpm
building. This is because there is hidden black
magic in the spec file, generating a VERSION
file based on the value of the Version of the
rpm. This Version is edited in place in the
spec file by a script, triggered from the Makefile,
which itself generates the version based on
git describe information.

This PR ensures that the VERSION file is always
present in our repo, by generating it directly
during any "make" step, and adding a
pre-commit hook to ensure that the VERSION
file is always correct.


skuba-update/VERSION
```diff
@@ -0,0 +1 @@
+1.2.6
```
jenting, line 1: Do we need to version control this file?

evrardjp: It's probably better, but it's not required, you're right! Fair point!

If it's version controlled, it just makes it simpler in the package building I guess: It won't require a `make` to build, as all the artifacts will already be present.

evrardjp: You're probably right, it's maybe even less error prone to not have it versioned (what if someone doesn't pull the last tags?).

In that case, does the pre-commit hook still make sense?

I am puzzled, on the fence between the two implementations. I like the readability of the file being present, and therefore no generation needs to happen in rpm building. But I realise the errors the developers might see.

evrardjp:
Rebased

skuba-update/setup.py
```diff
@@ -14,47 +14,15 @@
...
-    license="Apache License 2.0",
-    keywords="CaaSP",
-    url="https://github.com/SUSE/skuba-update",
-    packages=['skuba_update'],
-    classifiers=[
-        'Intended Audience :: Developers',
-        'License :: OSI Approved :: Apache License 2.0',
-        'Operating System :: POSIX :: Linux',
-    ], data_files=[
+    data_files=[
```
cmurphy, line 21: Why not put this in setup.cfg too? https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

skuba-update/VERSION
cmurphy: Keeping it in version control means it will be easy to unintentionally commit it since it will always be changing when developers run `make`. The version number should only be updated when the release manager actually wants to make a release.

jenting: The version file should be updated by release engineer, so the pre-commit hook is no longer needed :smile:

evrardjp: > Keeping it in version control means it will be easy to unintentionally commit it since it will always be changing when developers run `make`. The version number should only be updated when the release manager actually wants to make a release.

Wouldn't it be better in this case to have it version controlled, and _remove_ all the make wiring?

Wondering what's the right thing here, because having it _not version controlled_ means reintroducing the change in ``ci/packaging/suse/skuba_spec_template``, which makes it less obvious to track.

IMO it's not only when release needs to happen, but when at development time, we are ready to say "This deserves a bump in the release version".

I am definitely on the fence.

skuba-update/setup.py
evrardjp: It can, but by default the - are replaced with underscore IIRC, when setup.cfg is used. I didn't want to deal with the migration of those files.

The best would indeed move to setup.cfg for everything, or better, pyproject.toml.

However I am not sure our packaging is ready for this.

evrardjp:
I am abandonning this, I don't see much value in it. If it really had some value, we would probably have merged it by now. Anyone can resurrect this if necessary.

PR #1070, opened 2020-04-28
Clarify some code


evrardjp:
Without this patch, one might be confused by the meaning of
restart, and why we are doing actions twice.

This patch makes it clear that the method used to evaluate
if zypper should run twice is in fact dependent of zypper itself
and it's return code (not on some kind of obscure code output).
It should be easier to follow/review with this kind of wording.

It also removes useless methods, to be more succint.

### DIFF
--- a/skuba-update/skuba_update/skuba_update.py
+++ b/skuba-update/skuba_update/skuba_update.py
@@ -43,7 +43,6 @@ ZYPPER_REBOOT_NEEDED_PATH = '/var/run/reboot-needed'
 REBOOT_REQUIRED_PATH = '/var/run/reboot-required'
 
 # Exit codes as defined by zypper.
-
 ZYPPER_EXIT_INF_UPDATE_NEEDED = 100
 ZYPPER_EXIT_INF_SEC_UPDATE_NEEDED = 101
 ZYPPER_EXIT_INF_REBOOT_NEEDED = 102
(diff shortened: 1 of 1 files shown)


evrardjp:
Acceptance test failure fix: https://github.com/SUSE/skuba/pull/1066

davidcassany:
LGTM

evrardjp:
Rebased.

skuba-update/skuba_update/skuba_update.py
```diff
@@ -256,10 +249,10 @@ def is_zypper_error(code):
     return code != 0 and code < ZYPPER_EXIT_INF_UPDATE_NEEDED
 
 
-def is_restart_needed(code):
+def zypper_needs_transaction_restart(code):
```
dirkmueller, line 252: what is the _transaction_ good for? this is just an update stack update (not really a transaction restart afaik). I would find something like 

```
zypper_packagemanager_needs_restart()
```

would be more intuitive to me (thats at least matching the nomenclature in the man page)

evrardjp: Yup, even better name! Thanks!

evrardjp:
The problem in update acceptance should be fixed with https://github.com/SUSE/skuba/pull/1169 when the CI is back green.

flavio:
Overall LGTM, but please take a look at the test that is failing. I suspect it might not be related with this change, but I'm not 100% sure about that

evrardjp:
Rebased.

PR #1071, opened 2020-04-28
Clarify dev environment

evrardjp:
The Makefile is only used for testing purposes, and
is not used in packaging. It most likely won't change
anytime soon.

The `make install` instruction can go away:
For test purposes, we have tox (and make test).
We should probably not install things on the host
without a venv anyway with `make install` (at best
we should use `pip install -e .`).

The `make test` uses tox to install the software, and
should stay.

The other two make instructions are currently broken,
and exposed through skuba itself, so I don't expect
this to be required.

Next to this, the README file had to be updated.

### DIFF
--- a/skuba-update/Makefile
+++ b/skuba-update/Makefile
@@ -1,17 +1,5 @@
 .PHONY: all
-all: install
-
-.PHONY: install
-install:
-	./setup.py install
-
-.PHONY: suse-package
-suse-package:
-	ci/packaging/suse/rpmfiles_maker.sh
-
-.PHONY: suse-changelog
-suse-changelog:
-	ci/packaging/suse/changelog_maker.sh "$(CHANGES)"
+all: test
 
 .PHONY: test
 test:
--- a/skuba-update/README.md
+++ b/skuba-update/README.md
@@ -18,20 +18,14 @@ To set the development environment consider the following commands:
 # Get into the repository folder
 $ cd skuba-update
 
-# Initiate the python3 virtualenv
-$ python3 -m venv .env3
+# Ensure you have Shellcheck (haskell) and tox (python)
+$ sudo zypper install ShellCheck python-tox
 
-# Activate the virutalenv
-$ source .env3/bin/activate
+# Alternatively: You can also install tox in user space (shellcheck still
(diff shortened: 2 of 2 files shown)


evrardjp:
Acceptance test is still blocked by https://github.com/SUSE/skuba/pull/1066

evrardjp:
Rebased.

skuba-update/Makefile
```diff
@@ -1,17 +1,5 @@
 .PHONY: all
-all: install
-
-.PHONY: install
-install:
-	./setup.py install
```
cmurphy, line 6: If developers like using `make` as the interface for installing why change that?

evrardjp: because that's the wrong command?

suse-package/suse-changelog are not working, proving it's not used by devs, and has rotten (it's a duplicate from the time skuba and skuba-update were separate).

For install, I suppose ``pip install -e .`` should be used instead ?

evrardjp:
> The commit message/PR description could be clarified slightly, it currently says `The Makefile...` and I first assumed it was referring to the top-level Makefile.

Done.

skuba-update/Makefile
Klaven, line 6: my question here, is this file still needed? I am not super famillier with the project but it did not seem like this was used elsewhere? can we also delete setup.py while we are at it?

skuba-update/README.md
```diff
@@ -18,20 +18,14 @@ To set the development environment consider the following commands:
...
-# Install development dependencies
-$ pip install -r dev-requirements.txt
-
-# We need the shellcheck utility. You can install it from the package manager
-# you might be using.
-$ sudo zypper install ShellCheck
-
-# Run tests and code style checks
+# Run tests and code style checks without containers
 $ make test
```
flavio, line 29: Maybe something to classify under "tech debt": why don't we provide a make target that builds a container with all the test dependencies and another make target that runs the tests inside of this container?

This is a common approach used by lots of projects

Klaven: :+1:

evrardjp:
Well, if nobody cares, I will just close this.

PR #1073, opened 2020-04-29
Add gather logs of crio


evrardjp:
It would be useful to also have crio service logs.

This should probably do it.


cmurphy:
lgtm, there's now a crio.log under the platform logs directory

PR #1075, opened 2020-04-29
Update builder to 15.2


evrardjp:


mjura:
> Purpose?

We would like to update to newer version :)

evrardjp:
Yes, it's not _necessary_, but there is no need to lag behind ...

jenting:
Let's me the ball rolls :)

PR #1076, opened 2020-04-29
Send crio files on upgrade

evrardjp:
Without this patch, deployers will have to deploy our default
configuration files at each upgrade on every node, should they
change.

This ensures the files are in sync.

### DIFF
--- a/pkg/skuba/actions/node/upgrade/apply.go
+++ b/pkg/skuba/actions/node/upgrade/apply.go
@@ -19,6 +19,7 @@ package upgrade
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -33,6 +34,7 @@ import (
 	"github.com/SUSE/skuba/internal/pkg/skuba/kured"
 	"github.com/SUSE/skuba/internal/pkg/skuba/node"
 	upgradenode "github.com/SUSE/skuba/internal/pkg/skuba/upgrade/node"
+	"github.com/SUSE/skuba/pkg/skuba"
 )
 
 func Apply(client clientset.Interface, target *deployments.Target) error {
@@ -128,6 +130,29 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 			return err
 		}
 	}
+
+	/*
+		Always upload crio files, regardless of the version (allows to
+		enforce user behavior during patch updates).
+		During the upgrade from 1.16 to 1.18 crio, the cri-o package will
+		handle overriding the old crio sysconfig to an "empty" sysconfig,
+		and the cri.configure action will be enough.
+		We can remove the conditionals and only
+		keep the cri.configure action when caasp 4.2.0 is not supported
+		anymore (everyone has updated crio to 1.18)
+	*/
+	if _, err := os.Stat(skuba.CriDefaultsConfFile()); err == nil {
+		err = target.Apply(nil, "cri.configure")
+		if err != nil {
+			return err
+		}
+	} else if _, err := os.Stat(skuba.CriDockerDefaultsConfFile()); err == nil {
+		err = target.Apply(nil, "cri.sysconfig")
+		if err != nil {
+			return err
+		}
+	}
+
 	if nodeVersionInfoUpdate.HasMajorOrMinorUpdate() {
 		err = target.Apply(deployments.KubernetesBaseOSConfiguration{
 			UpdatedVersion: nodeVersionInfoUpdate.Update.KubeletVersion.String(),


pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -128,6 +128,12 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 			return err
 		}
 	}
+	if nodeVersionInfoUpdate.HasMajorOrMinorUpdate() {
```
cmurphy, line 131: This is the same condition as the one below, why not combine them?

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -128,6 +128,12 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
+		err = target.Apply(nil, "cri.configure")
```
cmurphy, line 132: In the bootstrap and join cases, this target is guarded by a check for the addon config file: https://github.com/SUSE/skuba/blob/ecf11c5b5f5510fb2cd660dd196cbc9afc0f33a4/pkg/skuba/actions/node/bootstrap/bootstrap.go#L122-L126 is that not required here as well?

mjura: Instead sending crio configuration file directly on the upgraded nodes, we should first take care about updating <CLUSTER-INIT>/addons/crio/conf.d because this is the place where user will look for any changes. If this would happen we don't have to this change at all, because PR https://github.com/SUSE/skuba/pull/1061 is relaying on configuration from <CLUSTER-INIT>

pkg/skuba/actions/node/upgrade/apply.go
evrardjp: Yes, my bad. This was the basis for a convo. The PR is not yet ready, I didn't even test it.

mjura: What if somebody will start to upgrade from CaaSP 4.0 GM, where there is `crio v1.15` ?

pkg/skuba/actions/node/upgrade/apply.go
evrardjp: > In the bootstrap and join cases, this target is guarded by a check for the addon config file:
> 
> https://github.com/SUSE/skuba/blob/ecf11c5b5f5510fb2cd660dd196cbc9afc0f33a4/pkg/skuba/actions/node/bootstrap/bootstrap.go#L122-L126
> is that not required here as well?

The file in sysconfig wasn't changed since its inception (until now, where we can remove it). So I guess the sysconfg doesn't need to run on upgrades. However you're right, we should probably have extra actions to behave in a consistent manner. For example, something to cleanup the file, if version of the upgrade is 17.

evrardjp:
I added the WIP to clarify that this is _indeed_ wip and a basis for convos

pkg/skuba/actions/node/upgrade/apply.go
mjura: > The file in sysconfig wasn't changed since its inception (until now, where we can remove it). So I guess the sysconfg doesn't need to run on upgrades. However you're right, we should probably have extra actions to behave in a consistent manner. For example, something to cleanup the file, if version of the upgrade is 17.

sysconfig file is written by the cri-o package everytime when it is upgraded

evrardjp: But we won't need it starting from crio 1.18, isn't it? I thought this was removed from the package.

evrardjp:
This is not ready for review, this was just a base for convo :)

pkg/skuba/actions/node/upgrade/apply.go
mjura: cri-o `sysconfig` is still OS method for providing service configuration, for v1.18 we want to only stop using it

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -128,6 +128,12 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
+		if err != nil {
+			return err
+		}
+	}
 	if nodeVersionInfoUpdate.HasMajorOrMinorUpdate() {
```
mjura, line 137: Please move code here and then we will merge it

pkg/skuba/actions/node/upgrade/apply.go
evrardjp: That's the difference I didn't know. I expected it was removed from the package. Thanks for the clarification.

evrardjp: I have added a comment in the code, I hope it clarifies.

pkg/skuba/actions/node/upgrade/apply.go
evrardjp: Yes the code is wrong, needs the guard @cmurphy was mentioning.
Should be fixed now.

evrardjp:
Please do not review yet :)

evrardjp:
I have removed the WIP tag, so that I can get reviews while I am doing my last bit of testing.

evrardjp:
worked on my machine, ready for review.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -128,6 +130,29 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
+	/*
+		Always upload crio files, regardless of the version (allows to
+		enforce user behavior during patch updates).
+		During the upgrade from 1.16 to 1.18 crio, the cri-o package will
+		handle overriding the old crio sysconfig to an "empty" sysconfig,
+		and the cri.configure action will be enough.
+		We can remove the conditionals and only
+		keep the cri.configure action when caasp 4.2.0 is not supported
+		anymore (everyone has updated crio to 1.18)
+	*/
```
jenting, line 143: nit: could you please make the comments shift a tab space?

evrardjp: Not sure what you meant here? It's my IDE that did the shifting :p 
Do you mean indent with one more tab?

jenting: > Do you mean indent with one more tab?

yes

evrardjp: I just tried with one more and one less tab for the indented text, and my IDE is always overriding. Probably due to the autoformatting config I have.

It doesn't complain on
```
	/*
	 *	Always upload crio files, regardless of the version (allows to
	 *	enforce user behavior during patch updates).
	 *	During the upgrade from 1.16 to 1.18 crio, the cri-o package will
	 *	handle overriding the old crio sysconfig to an "empty" sysconfig,
	 *	and the cri.configure action will be enough.
	 *	We can remove the conditionals and only
	 *	keep the cri.configure action when caasp 4.2.0 is not supported
	 *	anymore (everyone has updated crio to 1.18)
	 */
```

evrardjp: I can also do  ``// `` on everyline. It doesn't matter to me :)

jenting: I personally prefer `//` on every line.

mjura:
This change require https://github.com/SUSE/skuba/pull/1086

evrardjp:
I think this change can merge now that things are green, but those two patches certainly needs to be in.

pkg/skuba/actions/node/upgrade/apply.go
evrardjp: Will do if we need another patch and another person wants this.
No point for restarting the whole CI machine for a nit I guess...

PR #1077, opened 2020-04-29
Add disable repo feature


evrardjp:
This will disable a repo if the kubelet minor version is outdated.
Patch version is not taken into consideration.

Fixes: bsc#1165577
Closes: https://github.com/SUSE/avant-garde/issues/1371

### DIFF
--- a/skuba-update/skuba_update/skuba_update.py
+++ b/skuba-update/skuba_update/skuba_update.py
@@ -25,6 +25,7 @@ from datetime import datetime
 from pathlib import Path
 from xml.etree import ElementTree
 
+
 import pkg_resources
 
 # Since zypper 1.14.0, it will automatically create a `/var/run/reboot-needed`
@@ -43,7 +44,6 @@ ZYPPER_REBOOT_NEEDED_PATH = '/var/run/reboot-needed'
 REBOOT_REQUIRED_PATH = '/var/run/reboot-required'
 
 # Exit codes as defined by zypper.
-
 ZYPPER_EXIT_INF_UPDATE_NEEDED = 100
 ZYPPER_EXIT_INF_SEC_UPDATE_NEEDED = 101
 ZYPPER_EXIT_INF_REBOOT_NEEDED = 102
@@ -59,6 +59,26 @@ KUBE_DISRUPTIVE_UPDATES_KEY = 'caasp.suse.com/has-disruptive-updates'
 KUBE_CAASP_RELEASE_VERSION_KEY = 'caasp.suse.com/caasp-release-version'
 
 
+class ZypperSearchException(Exception):
+    pass
+
+
+class ZypperVersionTooLow(Exception):
+    pass
+
+
+class UnparseableZypperVersion(Exception):
+    pass
+
+
+class PrivilegesTooLow(Exception):
+    pass
+
+
+class ZypperRunError(Exception):
+    pass
+
+
 def main():
     """
     main-entry point for program.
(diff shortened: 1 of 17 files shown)


skuba-update/skuba_update/skuba_update.py
```diff
@@ -315,7 +447,7 @@ def run_zypper_command(command, needsOutput=False):
 
     process = run_command(zypperCommand, needsOutput)
     if is_zypper_error(process.returncode):
-        raise Exception('"{0}" failed'.format(' '.join(zypperCommand)))
+        raise ZypperRunError('"{0}" failed'.format(' '.join(zypperCommand)))
```
evrardjp, line 450: Side note for reviewers: This is necessary to avoid catching _all_ exceptions above, but also it's a better practice to send custom exceptions.

evrardjp:
autopep8 was not breaking up lines appropriately, so I did it manually.

skuba-update/skuba_update/skuba_update.py
```diff
@@ -68,19 +88,39 @@ def main():
...
 
     if os.geteuid() != 0:
-        raise Exception('root privileges are required to run this tool')
+        raise PrivilegesTooLow('root privileges are required to run this tool')
 
     run_zypper_command(['ref', '-s'])
+
     if not args.annotate_only:
-        code = update()
+        pkgs = parse_kubelet_pkglist(get_kubelet_packages_details())
```
evrardjp, line 103: should I just catch the case where no kubelet is installed, directly log() an error, and  fail?

davidcassany: It is already failing if no kubelet if found right? I don't think there is need to be more specific or having an special way of handling this particular case. I just checked `skuba-update` package requires `kubernetes-client`, probably it would make sense to also add `kubernetes-kubelet` ass a requirement. However if this requirement is not matched I'd say this the node is terribly broken and I'd consider `skuba-update` failures mostly meaningless in that case.

evrardjp:
> LGTM, I just see a corner case that I am not sure we should try to cover. What if the user manually disables the updates channel? Then skuba-update would atomatically re-enable it. I consider this is unexpected from users POV but not necessarily wrong. After all if skuba-update is running we should expect that the user wants to get latest updates, also if the user does not want to get updates but the notifications, there is always the chance to configure skuba-update to not apply any change.

Agreed. 

Thanks for the review. I will figure out why the update-acceptance tests fails now.

skuba-update/skuba_update/skuba_update.py
evrardjp: I am not sure if it's required to have kubelet as a requirement upon installing. Sensu stricto, skuba-update doesn't require kubelet. It just won't work until kubelet is installed. I am okay if someone wants to change the packaging though.

evrardjp:
I've modified the tests to add the required kubernetes-kubelet package.

evrardjp:
We have enough to merge now, but I am hesitating to merge it now, as this would effectively make that part of 4.2.x, not 4.3.x.

evrardjp:
I will wait for @jordimassaguerpla 's greenlight before merging it. Except if we want this for 4.2.x (and avoid the release notes we discussed in the related issue?)

evrardjp:
Can this merge now?  :D

jordimassaguerpla:
Let me replay with a question ;)

Is this about k8s 1.18? k8s 1.18 is the only feature that I see can't be merged because it is planned for later.

Otherwise, is this ready to be released? Meaning there is a PR for docs (if needed) and you have QA gave you a "thumbsup" (basically so they know about it so this can be added to their test plan).

jordimassaguerpla:
> Let me replay with a question ;)
> 
> Is this about k8s 1.18? k8s 1.18 is the only feature that I see can't be merged because it is planned for later.
> 
> Otherwise, is this ready to be released? Meaning there is a PR for docs (if needed) and you have QA gave you a "thumbsup" (basically so they know about it so this can be added to their test plan).

I didn't say, but to be explicit, if  your answers are "No, it is not k8s 1.18 only" and "Yes this is ready", go for it ;)

evrardjp:
It's not for k8s 1.18 per se.
It's a relatively major change in behaviour, but it's not really user facing. You're right this should probably have a release note, and have QA approval.

I will ping QA for this.

atighineanu:
LGTM. Bootstrap & update of cluster with skuba from this PR worked fine.

Martin-Weiss:
I have not read through the whole thread nor did I check the changes in the code - so please forgive me if my concerns are already covered!

I see two problems with disabling repos:

1. How do you know which repos to disable? Repo names are NOT standardized or static as they change i.e. via SUSE manager

2. In case we disable a repo the dependency resolution can get broken and especially during upgrades / updates or post-installation of other packages with dependencies (i.e. via a SUSE Manager salt state) can cause problems

evrardjp:
1. Repo names are inferred from where "kubernetes-kubelet" package can be found.
2. Disabling the repo indeed can break dependency resolution.

However, the only package that I foresee possible problems would be "cni-plugins". The rest of the packages is pretty much isolated. I think it's an "okay" stop-gap until we have a better solution, more cluster centric.

Martin-Weiss:
Just tried to test this repo disablement and had to realize that disabling the caasp-update channel alone does NOT solve the problem! It seems that also the containers channel also would need to be deactivated..

grafik

Martin-Weiss:
So for a quick workaround I disabled the containers-updates and the caasp-updates channel to be able to run skuba-update without complaining.
After running skuba-update I re-enabled these two channels.

evrardjp:
It's probably worth reverting.

davidcassany:
I think we should revert this... this will turn into a mess since it looks like that our CaaSP MR end up in multiple channels from customers PoV...

evrardjp:
https://github.com/SUSE/skuba/pull/1103

jordimassaguerpla:
Is this because we have k8s client package into the containers module?

evrardjp:
yes, it's not fully isolated.

PR #1082, opened 2020-05-04
Fix expected formatting for golang 1.14

evrardjp:
With golang 1.14, the strings expectations have a little bit changed.
Instead of refactoring the test, this takes the lazy approach of
just matching what's expected. Sorry.

### DIFF
--- a/pkg/skuba/actions/auth/login_test.go
+++ b/pkg/skuba/actions/auth/login_test.go
@@ -239,7 +239,7 @@ func Test_Login(t *testing.T) {
 				Password:           mockDefaultPassword,
 				InsecureSkipVerify: true,
 			},
-			expectedErrorMsg: "parse url: parse http://%41:8080/: invalid URL escape \"%41\"",
+			expectedErrorMsg: "invalid URL escape \"%41\"",
 		},
 		{
 			name:  "oidc server with incorrect port number",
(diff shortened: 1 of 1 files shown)


jenting:
Sorry, I did not follow Golang 1.14 recently, could you please elaborate more?

evrardjp:
I don't really know the root cause, and I have been a little bit lazy out there. But it seems those exceptions messages are wrapping the offending text between double quotes.

jenting:
> I don't really know the root cause, and I have been a little bit lazy out there. But it seems those exceptions messages are wrapping the offending text between double quotes.

okay, I'll look into it if I have time, thanks :bow:

evrardjp:
Oh that would be awesome. I had the intention to have a look when we'll move to 1.14 officially, but if you could have a look right now, that sounds even better.

jenting:
Release not on [golang 1.14](https://golang.org/doc/go1.14)
> When parsing of a URL fails (for example by Parse or ParseRequestURI), the resulting Error message will now quote the unparsable URL. This provides clearer structure and consistency with other parsing errors.

I think this is the reason why an unparsable URL got quoted at Golang 1.14.

jenting:
Let's defer merge this until we officially use Go1.14 _or_ maybe we could change the check error message logic to compatible all Golang version, this might be better for all users.

jenting:
@evrardjp I use your branch to fix this PR to both compatible to go >= 1.14.X and go < 1.14.X.
I can't think of a better way unless using a substring match. PTAL :bow:

PR #1072, opened 2020-04-28
vsphere: support set hostname from dhcp server

jenting:
## Why is this PR needed?

With vSphere `cpi_enable=true` and bootstrapping a cluster with `cloud-provider=vsphere`, the node name has to be the node hostname.

Current CaaSP vSphere environment, the deployed node's hostname is the DNS name, this makes the user have to find the current hostname via the vSphere web client, not the terraform output.

Therefore, for user experience _or_ automation tests, it better supports to set node's hostname from DHCP server.

## What does this PR do?

Setup hostname by cloud-init and reload wicked daemon to apply a new DHCP configuration.

## Anything else a reviewer needs to know?

N/A

## Info for QA

This is info for QA so that they can validate this. This is **mandatory** if this PR fixes a bug.
If this is a new feature, a good description in "What does this PR do" may be enough.

### Related info

Info that can be relevant for QA:
* link to other PRs that should be merged together: #1050
* link to packages that should be released together
* upstream issues

### Status **BEFORE** applying the patch

After terraform apply, the node's hostname is a 6 octets number (comes from DHCP server).

```
$ terraform output
ip_load_balancer = {
  "jenting-vsphere-cloud-provider-lb-0" = "10.84.72.14"
}
ip_masters = {
  "jenting-vsphere-cloud-provider-master-0" = "10.84.72.212"
}
ip_workers = {
  "jenting-vsphere-cloud-provider-worker-0" = "10.84.72.57"
  "jenting-vsphere-cloud-provider-worker-1" = "10.84.73.118"
}
```

```
$ skuba cluster init --control-plane 10.84.72.14 --cloud-provider vsphere my-cluster
$ cd my-cluster
$ cp cloud/vsphere/vsphere.conf.template cloud/vsphere/vsphere.conf
$ vim cloud/vsphere/vsphere.conf
$ skuba node bootstrap -s -u sles -t 10.84.72.212 <master-0-node-6-octets-number> -v 5
$ skuba node join --role worker -s -u sles -t 10.84.72.57 <worker-0-node-6-octets-number> -v 5
$ skuba node join --role worker -s -u sles -t 10.84.73.118 <worker-1-node-6-octets-number> -v 5
```

### Status **AFTER** applying the patch

Set `hostname_from_dhcp = false` and after terraform apply, the node's hostname is the same as `terraform output` hostname.

```
$ vim terraform.tfvars
hostname_from_dhcp = false

$ terraform output
ip_load_balancer = {
[...]

PR #1084, opened 2020-05-05
libvirt/openstack: support set hostname from dhcp server

jenting:
## Why is this PR needed?

Sync hostname setting for libvirt/openstack platform.
xref #1072

## What does this PR do?

Enable/disable hostname setting for libvirt/openstack platform.

N/A

## Info for QA

### Related info

* link to other PRs that should be merged together #1072 
* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/libvirt/cloud-init/common.tpl
+++ b/ci/infra/libvirt/cloud-init/common.tpl
@@ -34,17 +34,17 @@ ${repositories}
     download.use_deltarpm: "true"
 
 # WARNING!!! Do not use cloud-init packages module when SUSE CaaSP Registraion
-# Code is provided. In this case repositories will be added in runcmd module 
+# Code is provided. In this case repositories will be added in runcmd module
 # with SUSEConnect command after packages module is ran
 #packages:
 
 bootcmd:
   - ip link set dev eth0 mtu 1400
-  # Hostnames from DHCP - otherwise localhost will be used
-  - /usr/bin/sed -ie "s#DHCLIENT_SET_HOSTNAME=\"no\"#DHCLIENT_SET_HOSTNAME=\"yes\"#" /etc/sysconfig/network/dhcp
-  - netconfig update -f
 
 runcmd:
+  # Enable/Disable hostname from DHCP
+  - /usr/bin/sed -ie "s#DHCLIENT_SET_HOSTNAME=\"yes\"#DHCLIENT_SET_HOSTNAME=\"${hostname_from_dhcp}\"#" /etc/sysconfig/network/dhcp
+  - systemctl restart wicked
 ${register_scc}
 ${register_rmt}
 ${commands}
--- a/ci/infra/libvirt/cloud-init/common.tpl
+++ b/ci/infra/libvirt/cloud-init/common.tpl
@@ -43,7 +43,7 @@ bootcmd:
 
 runcmd:
   # Enable/Disable hostname from DHCP
-  - /usr/bin/sed -ie "s#DHCLIENT_SET_HOSTNAME=\"yes\"#DHCLIENT_SET_HOSTNAME=\"${hostname_from_dhcp}\"#" /etc/sysconfig/network/dhcp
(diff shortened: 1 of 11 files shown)


jenting:
rebase and force push

ci/infra/libvirt/cloud-init/common.tpl
```diff
@@ -34,17 +34,20 @@ ${repositories}
...
 bootcmd:
   - ip link set dev eth0 mtu 1400
-  # Hostnames from DHCP - otherwise localhost will be used
-  - /usr/bin/sed -ie "s#DHCLIENT_SET_HOSTNAME=\"no\"#DHCLIENT_SET_HOSTNAME=\"yes\"#" /etc/sysconfig/network/dhcp
-  - netconfig update -f
 
 runcmd:
+  # Set node's hostname from DHCP server
+  - sed -i -e '/^DHCLIENT_SET_HOSTNAME/s/^.*$/DHCLIENT_SET_HOSTNAME=\"${hostname_from_dhcp}\"/' /etc/sysconfig/network/dhcp
+  - systemctl restart wicked
```
davidcassany, line 50: Looks great to me restarting wicked rather than calling netconfig directly, this way we do not interfere with the network-online.target (I am quite convinced this `netconfig update -f` command is a source of issues, however I could not verify it in my tests). In any case, it is not clear to me why this has been moved from bootcmd to runcmd. I would expect the network configuration to happen completely at boot time, before the network-online.target is reached. For that the configuration should happen at bootcmd, is restarting wicked at bootcmd causing issues?

jenting: Thanks for giving me that `runcmd` is at post network-online.target.

The main reason I did is due to the cloud-init doc:
> bootcmd will run on every boot
> runcmd only runs during the first boot

But I agree that the setting network configuration at boot time. Let me change back to bootcmd and verify it.

jenting: Change from runcmd to bootcmd doesn't work after testing :thinking:

davidcassany: Interesting, probably we are facing the same race condition we faced in libvirt that forced us to include an ugly sleep on runcmd commands for libvirt. We are missing something here I could not manage to figure out a couple of weeks ago. Anyway, note that I did not dislike to move the wicked restart at runcmd, just that I did not understand it neither expect it, probably this was a correct approach from the very beginning, it just pisses me of not understanding it what's actually happening :P

By the way, how are you testing? Are you testing on libvirt?

jenting: No, I tested on vSphere, I think the baseOS and cloud-init behavior are the same. It should be platform independent.

But if you like, I could make a env with DHCP server in libvirt.

davidcassany: Right if having those commands on runcmd works fine I am also fine with it. Do you think it would make sense to include those commands as part of `cloud-init/commands.tpl` rather than include them here directly?

jenting: Well, put these parts into `commands.tpl` is also fine for me.
But include here is more intuitively at least to me. So personally I'd prefer to keep it in `common.tpl`.

If there is any good purpose change to `commands.tpl`, I agree to move it.

davidcassany: I have no strong arguments in any direction, thus I am fine if you prefer to include them here. I also have no clear clue on why the `commands.tpl` is treated separately.

I am sorry if my review took you some extra cycles, at the end, looks like we might end up with your initial proposal. Anyway I think it is good we discussed it.

jenting: Thanks for giving me the hint that `commands.tpl` exist :bow:, I did not notice this file before.

PR #1086, opened 2020-05-05
Improve upgrading crio.conf.d file

mjura:
## Why is this PR needed?

It is improving crio configuration 

## What does this PR do?

Instead `mv` it is `cp` whole directory with with caasp and users configuration files

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/cri.go
+++ b/internal/pkg/skuba/deployments/ssh/cri.go
@@ -52,7 +52,10 @@ func criConfigure(t *Target, data interface{}) error {
 		}
 	}
 
-	_, _, err = t.ssh("mv -f /tmp/crio.conf.d/01-caasp.conf /etc/crio/crio.conf.d/01-caasp.conf")
+	if _, _, err = t.ssh("mkdir -p /etc/crio/crio.conf.d"); err != nil {
+		return err
+	}
+	_, _, err = t.ssh("cp -r /tmp/crio.conf.d/*.conf /etc/crio/crio.conf.d")
 	return err
 }
 
--- a/pkg/skuba/actions/cluster/init/manifests.go
+++ b/pkg/skuba/actions/cluster/init/manifests.go
@@ -18,6 +18,8 @@
 package cluster
 
 const (
+	CriConfFolderReadme = `This directory contains CRI-O configuration files that are uploaded to CaaSP nodes. All the files (except this README) will be uploaded to the /etc/crio/conf.d/ folder. If it is required user custom files name as 99-custom.conf
+	`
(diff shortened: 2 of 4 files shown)


mjura:
> Wouldn't it be better to only push *.conf files from the /tmp/crio.conf.d/ into /etc/crio/crio.conf.d instead?

Done

> Avoid failures when we'll start introducing README files around.

As your wish :+1:

evrardjp:
I still have to test it, sorry for my dismiss comment which might be confusing.

mjura:
This depends on https://github.com/SUSE/skuba/pull/1095

evrardjp:
This doesn't depend on #1095 anymore, but requires the a patch which includes 1.18 version support.

pkg/skuba/actions/cluster/init/manifests.go
```diff
@@ -18,14 +18,18 @@
...
+	`
 	criDockerDefaultsConf = `## Path           : System/Management
 ## Description    : Extra cli switches for crio daemon
 ## Type           : string
 ## Default        : ""
 ## ServiceRestart : crio
 #
 CRIO_OPTIONS=--pause-image={{.PauseImage}}{{if not .StrictCapDefaults}} --default-capabilities="CHOWN,DAC_OVERRIDE,FSETID,FOWNER,NET_RAW,SETGID,SETUID,SETPCAP,NET_BIND_SERVICE,SYS_CHROOT,KILL,MKNOD,AUDIT_WRITE,SETFCAP"{{end}}`
-	criDefaultsConf = `{{if not .StrictCapDefaults}}[crio.runtime]
+	criDefaultsConf = `# PLEASE DON'T EDIT THIS FILE!
```
evrardjp, line 30: nit: there is a trailing whitespace after the exclamation point.

mjura: Fixed

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -52,7 +52,10 @@ func criConfigure(t *Target, data interface{}) error {
 		}
 	}
 
-	_, _, err = t.ssh("mv -f /tmp/crio.conf.d/01-caasp.conf /etc/crio/crio.conf.d/01-caasp.conf")
+	if _, _, err = t.ssh("mkdir -p /etc/crio/crio.conf.d"); err != nil {
```
evrardjp, line 55: nit (optimization): This will technically upload all files, and then only copy the ones relevant. Can we not upload the README file over?

PR #1091, opened 2020-05-07
Add provision deploy stages daily jobs

pablochacin:
## Why is this PR needed?

E2E tests provision and deploy clusters using fixtures. It makes difficult to detect situations when the provisioning fails due to infrastructure issues or the deployment fails due to issues with packages or images required by skuba.

## What does this PR do?

- Separates the provision and deploying from e2e tests and adds cluster sanity tests.
- Adds the skip setup option to e2e test target to prevent tests to deploy a cluster using fixtures.
- Handle the special case of upgrade tests, which must deploy the cluster specifying the kubernetes version

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


jordimassaguerpla:
If we set k8s to be 1.16.2 in the upgrade tests, shouldn't we reflect that in the naming of the test?

PR #1095, opened 2020-05-08
Maintain /etc/sysconfig/crio for CRI-O v1.18 upgrade

mjura:
## Why is this PR needed?

Cri-o v1.18.0 has a nasty bug which can't parse 
```
--default-capabilities="CHOWN,DAC_OVERRIDE,FSETID,FOWNER,NET_RAW,SETGID,SETUID,SETPCAP,NET_BIND_SERVICE,SYS_CHROOT,KILL,MKNOD,AUDIT_WRITE,SETFCAP"
```
from /etc/sysconfig/crio. Probably it will be fixed in cri-o v1.18.1 but we can't block upgrade. Besides this cri-o v1.18.0 introducing users custom configuration files which can be installed in /etc/crio/crio.conf.d and this is the way which we are going to use for CaaSP.

## What does this PR do?

It is cleaning up `/etc/sysconfig/crio` and removing `--default-capabilities=" ... ` during CaaSP upgrade to Kubernetes and cri-o v1.18

Before CaaSP upgrade we have to generate new crio.conf.d configuration by 

```
skuba cluster init --control-plane <IP> /tmp/<TEST_CLUSTER_NAME>
cp -r /tmp/<TEST_CLUSTER_NAME>/addons/crio/conf.d <CURRENT_CLUSTER_NAME>/addons/crio
```

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/pkg/skuba/actions/node/upgrade/apply.go
+++ b/pkg/skuba/actions/node/upgrade/apply.go
@@ -132,24 +132,23 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 	}
 
 	/*
-		Always upload crio files, regardless of the version (allows to
-		enforce user behavior during patch updates).
-		During the upgrade from 1.16 to 1.18 crio, the cri-o package will
-		handle overriding the old crio sysconfig to an "empty" sysconfig,
-		and the cri.configure action will be enough.
+		Upload crio.conf.d files only when Kubernetes target version 
+		is v1.18. For cri-o package v1.18.0 we are migrating CaaSP 
+		configuration from /etc/sysconfig/crio to /etc/crio/crio.conf.d/01-caasp.conf
+		During the upgrade to 1.18 crio, skuba has to clean /etc/sysconfig/crio.
 		We can remove the conditionals and only
 		keep the cri.configure action when caasp 4.2.0 is not supported
 		anymore (everyone has updated crio to 1.18)
 	*/
-	if _, err := os.Stat(skuba.CriDefaultsConfFile()); err == nil {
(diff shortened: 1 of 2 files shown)


pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -132,24 +132,23 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
-	if _, err := os.Stat(skuba.CriDefaultsConfFile()); err == nil {
-		err = target.Apply(nil, "cri.configure")
-		if err != nil {
-			return err
-		}
-	} else if _, err := os.Stat(skuba.CriDockerDefaultsConfFile()); err == nil {
-		err = target.Apply(nil, "cri.sysconfig")
-		if err != nil {
-			return err
+	if kubernetes.LatestVersion().Minor() == 18 {
```
evrardjp, line 143: two problems with this:
1) We are not actively always uploading the files anymore, which could be a problem in terms of expectations for our customers.
2) We are now applying this only when _latestVersion_ is 18, not when the upgrade is going through 18. My problem is that someone from 1.17 could be waiting for 1.19 before triggering a 1.17->1.18->1.19 upgrade, on which this won't apply, AFAIK.

evrardjp:
Two questions:
1) Why isn't the packaging overwriting the file?
2) Shouldn't we simply apply the "copy the files" over, and ask the user to update the files in its cluster config folder? (Sounds not great though)

mjura:
Indeed this will be fixed in the package

PR #1108, opened 2020-05-14
Update base kubernetes version to upgrade from


evrardjp:
With now k8s 1.18.2 available in Devel, we should be dealing
with the upgrade of 1.17.4 to it in CI.

This should take care of it.

Needed-By: ``<Patch to implement script running at arbitrary times>``
Closes: https://github.com/SUSE/avant-garde/issues/1585, https://github.com/SUSE/avant-garde/issues/1586

### DIFF
--- a/ci/infra/testrunner/tests/utils.py
+++ b/ci/infra/testrunner/tests/utils.py
@@ -2,8 +2,8 @@ import signal
 import time
 import yaml
 
-PREVIOUS_VERSION = "1.16.2"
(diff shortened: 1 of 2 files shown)


evrardjp:
I suppose we'll need something to simulate the release notes content, for which we'll generate the crio files, remove the default_ file for crio, ... before merging.

evrardjp:
Though this can be done in two different phases, and it will probably bring the CI a little bit greener.

PR #1110, opened 2020-05-15
[WIP] Update cluster init before upgrade

mjura:
## Why is this PR needed?

For migration between `CaaSP 4.2.1` and `CaaSP 4.3` cluster addons directory should be updated. It was already covered by release notes
https://github.com/SUSE/caasp-release-notes/pull/51/files
This has to be also covered by our CI framework

## What does this PR do?

```
skuba cluster init --control-plane <PUBLIC_IP> /tmp/test-cluster
mv /tmp/test-cluster/addons/cri/conf.d <CURRENT-CLUSTER-NAME>/addons/cri/
rm <CURRENT-CLUSTER-NAME>/addons/cri/default_flags
```

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/skuba/skuba.py
+++ b/ci/infra/testrunner/skuba/skuba.py
@@ -50,7 +50,7 @@ class Skuba:
 
 
     @step
-    def cluster_init(self, kubernetes_version=None, cloud_provider=None):
+    def cluster_init(self, kubernetes_version=None, cloud_provider=None, path=None):
         logger.debug("Cleaning up any previous cluster dir")
         self.utils.cleanup_file(self.cluster_dir)
 
@@ -59,9 +59,12 @@ class Skuba:
             k8s_version_option = "--kubernetes-version {}".format(kubernetes_version)
         if cloud_provider:
             cloud_provider_option = "--cloud-provider {}".format(type(self.platform).__name__.lower())
+        path_option = "test-cluster"
+        if path:
+            path_option = "{}".format(path)
 
-        cmd = "cluster init --control-plane {} {} {} test-cluster".format(self.platform.get_lb_ipaddr(),
-                                                                          k8s_version_option, cloud_provider_option)
+        cmd = "cluster init --control-plane {} {} {} {}".format(self.platform.get_lb_ipaddr(),
+                                                                          k8s_version_option, cloud_provider_option, path_option)
         # Override work directory, because init must run in the parent directory of the
         # cluster directory
         self._run_skuba(cmd, cwd=self.workdir)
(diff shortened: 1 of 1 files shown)


ci/infra/testrunner/skuba/skuba.py
```diff
@@ -50,18 +50,20 @@ def cluster_deploy(self, kubernetes_version=None, cloud_provider=None,
 
 
     @step
-    def cluster_init(self, kubernetes_version=None, cloud_provider=None):
+    def cluster_init(self, kubernetes_version=None, cloud_provider=None, path=None):
```
pablochacin, line 53: We should add the `path` parameter also to the testrunner command in both the `bootstrap` and `deploy` commands, which call `cluster_init`  https://github.com/SUSE/skuba/blob/master/ci/infra/testrunner/testrunner.py#L167-L181  and https://github.com/SUSE/skuba/blob/master/ci/infra/testrunner/testrunner.py#L43-L56

And pass it when invoking the cluster init here https://github.com/SUSE/skuba/blob/master/ci/infra/testrunner/skuba/skuba.py#L43-L49 and here https://github.com/SUSE/skuba/blob/master/ci/infra/testrunner/skuba/skuba.py#L71-L74.

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -154,6 +156,10 @@ def node_remove(self, role="worker", nr=0):
     def node_upgrade(self, action, role, nr, ignore_errors=False):
         self._verify_bootstrap_dependency()
 
+        self.cluster_init("", "", "/tmp")
```
pablochacin, line 159: I'm worried if this is a one-time fix. That is, IIRC this applies only when upgrading from cri-0 `1.17` to `1.18` but not in other cases.

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -223,3 +229,18 @@ def _run_skuba(self, cmd, cwd=None, verbosity=None, ignore_errors=False):
             cwd=cwd,
             ignore_errors=ignore_errors,
         )
+
+    def _run_os_cmd(self, cmd, cwd=None, ignore_errors=False):
```
pablochacin, line 233: I only see you are using this `_run_os_cmd` once. Any reason for not calling `runshlelcommand` directly?

ci/infra/testrunner/skuba/skuba.py
```diff
@@ -154,6 +156,10 @@ def node_remove(self, role="worker", nr=0):
     def node_upgrade(self, action, role, nr, ignore_errors=False):
         self._verify_bootstrap_dependency()
 
+        self.cluster_init("", "", "/tmp")
+        cmd = "rsync -a /tmp/test-cluster/addons/ addons/"
```
pablochacin, line 160: I don't like this hardcoded path for several reasons. First, In the CI, for example it is unsafe to use `/tmp`. I would be safer to create a `tmp` directory in the workspace directory.  

Also, instead of hardcoding `test-cluster` you could better use the configuration parameter `self.config.skuba.cluster` because the `cluster_init` function will use it.

jenting:
related to #1251

mjura:
it is not  needed any  more

PR #1111, opened 2020-05-15
(bsc#1171533) Validate skuba cluster init input

evrardjp:
This will prevent issues like `\;`, empty strings.
I took the conservative approach of not allowing
any utf-8 char for now, but we might to relax that
later, if it's properly tested elsewhere to have
a name like 🚀.

Closes: https://github.com/SUSE/avant-garde/issues/1586


cmd/skuba/cluster/init.go
```diff
@@ -32,6 +35,22 @@ type initOptions struct {
...
+func isValidClusterName(clustername string) bool {
+	// NOT a letter, a number, underscore/dash
+	reg, err := regexp.Compile("[^a-zA-Z0-9_-]+")
+	if err != nil {
+		klog.Fatalf("Error compiling regex: %s", err)
+	}
+	if reg.MatchString(clustername) {
+		return false
+	}
+	if len(clustername) == 0 {
```
chentex, line 48: I would move this as a first step, even before compiling the regexp, is the least expensive to validate.

cmd/skuba/cluster/init.go
```diff
@@ -32,6 +35,22 @@ type initOptions struct {
 	StrictCapDefaults bool
 }
 
+// Checks if clustername provided is correct
+func isValidClusterName(clustername string) bool {
+	// NOT a letter, a number, underscore/dash
+	reg, err := regexp.Compile("[^a-zA-Z0-9_-]+")
```
chentex, line 41: is a `.` a character we also want to allow?

evrardjp: you tell me! :)

chentex: Let's block it for the moment and let's see what happens later.

evrardjp:
Retriggered testing.

cmd/skuba/cluster/init.go
```diff
@@ -32,6 +35,22 @@ type initOptions struct {
 	StrictCapDefaults bool
 }
 
+// Checks if clustername provided is correct
+func isValidClusterName(clustername string) bool {
+	if len(clustername) == 0 {
+		return false
+	}
+	// NOT a letter, a number, underscore/dash
+	reg, err := regexp.Compile("[^a-zA-Z0-9_-]+")
```
cmurphy, line 44: Not sure I agree with this change. This was working before:

```
$ skuba cluster init --control-plane 192.168.122.100 ⛄
[init] configuration files written to /home/colleen/clusters/⛄
$ cd ./⛄
$ skuba node bootstrap --user devuser --sudo --target 192.168.122.100 master
W0521 15:39:47.633782    7929 ssh.go:311] 
The authenticity of host '192.168.122.100:22' can't be established.
ECDSA key fingerprint is 4d:0e:01:f0:bd:af:a0:5b:ad:4a:98:61:0e:8a:89:e0.
I0521 15:39:47.633860    7929 ssh.go:312] accepting SSH key for "192.168.122.100:22"
I0521 15:39:47.633871    7929 ssh.go:313] adding fingerprint for "192.168.122.100:22" to "known_hosts"
[bootstrap] updating init configuration with target information
[bootstrap] writing init configuration for node
[bootstrap] applying init configuration to node
E0521 15:42:53.636361    7929 ssh.go:195] W0521 22:42:53.382455    2328 removeetcdmember.go:79] [reset] No kubeadm config, using etcd pod spec to get data directory
E0521 15:42:55.662885    7929 ssh.go:195] W0521 22:42:55.409223    2328 cleanupnode.go:81] [reset] Failed to remove containers: output: time="2020-05-21T22:42:55Z" level=fatal msg="failed to connect: failed to connect, make sure you are running as root and the runtime has been started: context deadline exceeded"
E0521 15:42:55.662914    7929 ssh.go:195] , error: exit status 1
E0521 15:42:56.845857    7929 ssh.go:195] No files found for firewalld.service.
E0521 15:43:00.836760    7929 ssh.go:195] Created symlink /etc/systemd/system/multi-user.target.wants/crio.service → /usr/lib/systemd/system/crio.service.
E0521 15:43:02.751170    7929 ssh.go:195] Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /usr/lib/systemd/system/kubelet.service.
E0521 15:43:03.165058    7929 ssh.go:195] W0521 22:43:02.911409    3877 configset.go:202] WARNING: kubeadm cannot validate component configs for API groups [kubelet.config.k8s.io kubeproxy.config.k8s.io]
E0521 15:43:03.201876    7929 ssh.go:195]       [WARNING Hostname]: hostname "master" could not be reached
E0521 15:43:03.201906    7929 ssh.go:195]       [WARNING Hostname]: hostname "master": lookup master on 192.168.122.1:53: no such host
E0521 15:46:16.262865    7929 ssh.go:195] W0521 22:46:16.007950    3877 manifests.go:225] the default kube-apiserver authorization-mode is "Node,RBAC"; using "Node,RBAC"
E0521 15:46:16.263964    7929 ssh.go:195] W0521 22:46:16.009205    3877 manifests.go:225] the default kube-apiserver authorization-mode is "Node,RBAC"; using "Node,RBAC"
E0521 15:46:46.065721    7929 ssh.go:195] Created symlink /etc/systemd/system/timers.target.wants/skuba-update.timer → /usr/lib/systemd/system/skuba-update.timer.
[bootstrap] successfully bootstrapped core components on node "192.168.122.100" with Kubernetes: "1.18.2"
[bootstrap] downloading secrets from bootstrapped node "192.168.122.100"
[bootstrap] deploying core add-ons on node "192.168.122.100"
[bootstrap] successfully bootstrapped core add-ons on node "192.168.122.100"
$ kubectl get node
NAME     STATUS   ROLES    AGE     VERSION
master   Ready    master   5m35s   v1.18.2
```

Seems like an unnecessary restriction that's unrelated to the bug that was reported.

evrardjp: That's interesting. Neither Kind nor Kops allow those names. Kubeadm is not testing those use cases, so I suppose everything is allowed from kubeadm perspective.
I thought we had to restrict the cluster names (for gangway), similarly to those tools above. I took the more lenient approach than the strict dns name validation checking.
You're right on the fact this isn't ready. Either I should be stricter, or looser.

evrardjp: Stricter here could mean follow the dns rfc, or do like docker: 
https://github.com/moby/moby/blob/master/daemon/names/names.go#L6

It seems crio doesn't have that restriction, runC doesn't validate those names, and kubeadm doesn't test those, so that's why it works. The safeguards in kind are implemented because it uses docker/moby behind the scenes, which _is_ restricted. 

I will relax this.

cmd/skuba/cluster/init.go
```diff
@@ -54,7 +56,15 @@ func NewInitCmd() *cobra.Command {
 				klog.Fatalf("init failed due to error: %s", err)
 			}
 		},
-		Args: cobra.ExactArgs(1),
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return fmt.Errorf("accepts 1 arg, received %d", len(args))
+			}
+			if len(args[0]) == 0 {
+				return fmt.Errorf("Invalid cluster name: expected non empty cluster name, received: %s", args[0])
```
jenting, line 64: The args is empty, so what's the purpose on prints `received: %s", args[0]` :thinking:

evrardjp: Done.

cmd/skuba/cluster/init.go
```diff
@@ -54,7 +56,15 @@ func NewInitCmd() *cobra.Command {
...
 			}
 		},
-		Args: cobra.ExactArgs(1),
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return fmt.Errorf("accepts 1 arg, received %d", len(args))
+			}
+			if len(args[0]) == 0 {
+				return fmt.Errorf("Invalid cluster name: expected non empty cluster name")
+			}
```
jenting, line 65:
```suggestion
			if len(args[0]) == 0 {
				return fmt.Errorf("Invalid cluster name: expected non empty cluster name")
			}
			if len(args) != 1 {
				return fmt.Errorf("accepts 1 arg, received %d", len(args))
			}
```

chentex: If you validate in this order you will get a panic if args in empty

jenting: Thanks, make some change~

cmd/skuba/cluster/init.go
```diff
@@ -54,7 +56,15 @@ func NewInitCmd() *cobra.Command {
 				klog.Fatalf("init failed due to error: %s", err)
 			}
 		},
-		Args: cobra.ExactArgs(1),
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return fmt.Errorf("accepts 1 arg, received %d", len(args))
```
cmurphy, line 61: This case is already covered without this change:

```
$ skuba cluster init --control-plane 1.1.1.1
Error: accepts 1 arg(s), received 0
Usage:
...
```

```
$ skuba cluster init --control-plane 1.1.1.1 x y z
Error: accepts 1 arg(s), received 3
Usage:
...
```

evrardjp: yes, exact args tests the number of arguments. If we want a custom validation function, we are removing this cobra.ExactArgs. For consistency, I used the exact same text as cobra. However, without this patch, --control-plane 1.1.1.1 "" would incorrectly pass.

Can we just please move on?

cmurphy: You're right, thanks for explaining :+1:

evrardjp: Sorry for the last bit of my comment. That's why I shouldn't work during my time-off.

evrardjp:
Rebased due to CI changes.

evrardjp:
Oh great, tests are now breaking again. Let me rebase first, as it was asked in a mail.

cmd/skuba/cluster/init.go
evrardjp: @jenting @chentex meant that it will panic if in the wrong order. Anyway it won't cover `cluster init "" --control-plane 127.0.0.1`. It's the length of the argument that can't be empty, not the amount of arguments. The amount of arguments is already taken care of by len(args).

jenting, line 65: This code would never be executed, please considering move before `if len(args) != 1`, thanks.

evrardjp: It looks okay to me:
```
jean-philippe@laptop-dell /t/skuba-test> skuba cluster init "" --control-plane 10.86.0.5
Error: Invalid cluster name: expected non empty cluster name
```

jenting: Ugh, okay, I did not think use `""`, smart :100:

PR #1112, opened 2020-05-18
[WIP] Deregister instances when destroying

lcavajani:
This idea comes from what is done with `validator`, thanks to @kravciak and @thehejik !

## Why is this PR needed?

This commit allows to de-register instances from RMT or from SCC automatically when destroying.

One specificity is the use of trigger within the null_resources. This avoids using external references (variables, resource attributes...) with the destroy provisioners which is deprecated.

For AWS, this commit also introduces a stanza to wait for cloud-init to finish.

`terraform fmt` also rewrites some files.

## Info for QA

I have just tested with `VMware` so far.

Non-registered instances are not impacted.
Registered instances are not deregistered.

Non-registered instances should not run the de-register step.
Instances are deregistered, this can be seen in the terraform outputs:

```hcl
null_resource.master[0] (remote-exec): Deregistering system from SUSE Customer Center
null_resource.master[0]: Still destroying... [id=5790196521205169156, 10s elapsed]

null_resource.master[0] (remote-exec): Deactivating caasp 4.0 x86_64 ...
null_resource.master[0] (remote-exec): -> Removing service from system ...
null_resource.master[0] (remote-exec): -> Removing release package ...

null_resource.master[0] (remote-exec): Deactivating sle-module-containers 15.1 x86_64 ...
null_resource.master[0] (remote-exec): -> Removing service from system ...
null_resource.master[0] (remote-exec): -> Removing release package ...
null_resource.master[0]: Still destroying... [id=5790196521205169156, 20s elapsed]

null_resource.master[0] (remote-exec): Deactivating sle-module-server-applications 15.1 x86_64 ...
null_resource.master[0] (remote-exec): -> Removing service from system ...
null_resource.master[0] (remote-exec): -> Removing release package ...

null_resource.master[0] (remote-exec): Deactivating sle-module-basesystem 15.1 x86_64 ...
null_resource.master[0] (remote-exec): -> Removing service from system ...
null_resource.master[0] (remote-exec): -> Removing release package ...
null_resource.master[0]: Still destroying... [id=5790196521205169156, 30s elapsed]

null_resource.master[0] (remote-exec): Cleaning up ...
null_resource.master[0] (remote-exec): Successfully deregistered system
```

## Docs

in progress....

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/libvirt/master-instance.tf
+++ b/ci/infra/libvirt/master-instance.tf
@@ -101,18 +101,23 @@ resource "libvirt_domain" "master" {
   }
 }
 
-resource "null_resource" "master_wait_cloudinit" {
+resource "null_resource" "master" {
   depends_on = [libvirt_domain.master]
   count      = var.masters
 
+  triggers = {
+    master_ips = "${join(",", libvirt_domain.master.*.network_interface.0.addresses.0)}"
+    username   = var.username
+  }
+
   connection {
     host = element(
-      libvirt_domain.master.*.network_interface.0.addresses.0,
-      count.index
+      split(",", self.triggers.master_ips),
+      count.index,
     )
-    user     = var.username
-    password = var.password
-    type     = "ssh"
+    user  = self.triggers.username
+    type  = "ssh"
+    agent = true
   }
 
   provisioner "remote-exec" {
@@ -120,10 +125,17 @@ resource "null_resource" "master_wait_cloudinit" {
       "cloud-init status --wait > /dev/null",
     ]
   }
+
+  provisioner "remote-exec" {
+    when = destroy
+    inline = [
+      "if sudo SUSEConnect -s | grep -qv 'Not Registered'; then sudo SUSEConnect -d; fi"
+    ]
+  }
 }
 
 resource "null_resource" "master_reboot" {
-  depends_on = [null_resource.master_wait_cloudinit]
+  depends_on = [null_resource.master]
   count      = var.masters
 
   provisioner "local-exec" {
--- a/ci/infra/libvirt/master-instance.tf
+++ b/ci/infra/libvirt/master-instance.tf
@@ -127,7 +127,8 @@ resource "null_resource" "master" {
   }
 
   provisioner "remote-exec" {
-    when = destroy
+    when       = destroy
+    on_failure = "continue"
     inline = [
       "if sudo SUSEConnect -s | grep -qv 'Not Registered'; then sudo SUSEConnect -d; fi"
     ]
--- a/ci/infra/libvirt/master-instance.tf
+++ b/ci/infra/libvirt/master-instance.tf
@@ -128,7 +128,7 @@ resource "null_resource" "master" {
 
   provisioner "remote-exec" {
     when       = destroy
-    on_failure = "continue"
+    on_failure = continue
     inline = [
       "if sudo SUSEConnect -s | grep -qv 'Not Registered'; then sudo SUSEConnect -d; fi"
     ]
--- a/ci/infra/libvirt/worker-instance.tf
+++ b/ci/infra/libvirt/worker-instance.tf
@@ -101,18 +101,23 @@ resource "libvirt_domain" "worker" {
   }
 }
 
-resource "null_resource" "worker_wait_cloudinit" {
+resource "null_resource" "worker" {
   depends_on = [libvirt_domain.worker]
   count      = var.workers
 
+  triggers = {
+    worker_ips = "${join(",", libvirt_domain.worker.*.network_interface.0.addresses.0)}"
+    username   = var.username
+  }
+
   connection {
     host = element(
-      libvirt_domain.worker.*.network_interface.0.addresses.0,
-      count.index
+      split(",", self.triggers.worker_ips),
+      count.index,
     )
-    user     = var.username
-    password = var.password
-    type     = "ssh"
+    user  = self.triggers.username
+    type  = "ssh"
+    agent = true
   }
 
   provisioner "remote-exec" {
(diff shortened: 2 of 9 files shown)


ci/infra/libvirt/worker-instance.tf
```diff
@@ -101,25 +101,37 @@ resource "libvirt_domain" "worker" {
...
+
+  provisioner "remote-exec" {
+    when = destroy
+    inline = [
+      "if sudo SUSEConnect -s | grep -v 'Not Registered' > /dev/null 2>&1; then sudo SUSEConnect -d; fi"
+    ]
+  }
 }
 
 resource "null_resource" "worker_reboot" {
```
thehejik, line 137: this resource uses old null_resource name, it should use:
`depends_on = [null_resource.worker]`

+ the same for master instance

thehejik: also `provisioner "local-exec"` bellow uses the old way for getting host address for connection

lcavajani: > this resource uses old null_resource name, it should use:
> `depends_on = [null_resource.worker]`
> 
>     * the same for master instance

yes good catch, I've pushed the changes for `libvirt` and `openstack`

thehejik:
I tried 1:1 node deployment on libvirt and it works correctly with SCC key and without it as well. The second commit fixed the tf resources dependency issue.

I'll try openstack tmrw.

thehejik:
We should also try how the `when = destroy` will behave when some nodes are off. This is actually broken in libvirt tf provider even with current code in master but it worked with vmware/openstack tf before.

lcavajani:
I would like to try with `SMT` as well. We don't ship terraform libvirt, we should not focus too much time I think if there is a bug but let's try out with other providers.

Martin-Weiss:
What will happen in case the customer does not use SMT/RMT/SCC and instead uses i.e. SUSE Manager and the servers are NOT registered via SUSEConnect?

lcavajani:
> What will happen in case the customer does not use SMT/RMT/SCC and instead uses i.e. SUSE Manager and the servers are NOT registered via SUSEConnect?

Nothing but when you use SUSE Manager, you have your own process since it's not handled by `terraform` so in this case it's up to you to "clean" that.

Martin-Weiss:
> 
> 
> > What will happen in case the customer does not use SMT/RMT/SCC and instead uses i.e. SUSE Manager and the servers are NOT registered via SUSEConnect?
> 
> Nothing but when you use SUSE Manager, you have your own process since it's not handled by `terraform` so in this case it's up to you to "clean" that.

Ok - so as long as the command does not give an error in case the server was not registered via SUSEConnect and as long as the command does not require internet connection I believe we are fine ;-).

kravciak:
This might be a problem:
- run terraform apply with 3 masters 3 workers (all machines are registered)
- change terraform.tfvars to 3 masters 2 workers
- terraform apply again (it removes 1 worker as expected)

Side effect is that all machines are unregistered now.

thehejik:
> I would like to try with `SMT` as well. We don't ship terraform libvirt, we should not focus too much time I think if there is a bug but let's try out with other providers.

RMT works:
```
null_resource.master[0] (remote-exec): Deregistering system from registration proxy https://my-rmt-server.cz
null_resource.master[0] (remote-exec): Deactivating caasp 4.0 x86_64 ...
...
```

lcavajani:
> This might be a problem:
> 
>     * run terraform apply with 3 masters 3 workers (all machines are registered)
> 
>     * change terraform.tfvars to 3 masters 2 workers
> 
>     * terraform apply again (it removes 1 worker as expected)
> 
> 
> Side effect is that all workers are unregistered now.

Aaaarg, that's pretty bad indeed and makes this not viable... 

My first try was to use the `remote-exec` directly in the virtual machine provider which I think would avoid that... but we can't use triggers here to use `var.username` which means I had to use `var.username` directly which leads to the deprecated warning.

Let's see if we find something

thehejik:
It fails on openstack when a node is off.

Unfortunately it is needed by scenario we use in validation process that we turn off a node and then remove offline node by skuba.

lcavajani:
yes, let's close this

lcavajani:
Actually, I think using `    on_failure = "continue"` should fix the problem when instance is off. We still need to figure something out in @kravciak case

ci/infra/libvirt/master-instance.tf
```diff
@@ -101,29 +101,41 @@ resource "libvirt_domain" "master" {
...
     host = element(
-      libvirt_domain.master.*.network_interface.0.addresses.0,
-      count.index
+      split(",", self.triggers.master_ips),
+      count.index,
     )
-    user     = var.username
-    password = var.password
-    type     = "ssh"
+    user  = self.triggers.username
```
pablochacin, line 118: Why use here triggers instead of the var?

lcavajani: This avoids using external references (variables, resource attributes...) with the destroy provisioners which is deprecated.

ci/infra/vmware/master-instance.tf
```diff
@@ -118,4 +124,12 @@ resource "null_resource" "master_wait_cloudinit" {
       "cloud-init status --wait > /dev/null",
     ]
   }
+
+  provisioner "remote-exec" {
+    when       = destroy
+    on_failure = "continue"
```
thehejik, line 130: It works with `on_failure = continue`, now tf is able to destroy cluster (tried on openstack) even when one node is offline.

It looks like the remote-exec is called several times and after 5 minutes it gives up and destroys the cluster anyway.

 But please remove quotes around `continue` as these are deprecated for keywords now:
```
Warning: Quoted keywords are deprecated

  on master-instance.tf line 125, in resource "null_resource" "master":
 125:     on_failure = "continue"

In this context, keywords are expected literally rather than in quotes.
Terraform 0.11 and earlier required quotes, but quoted keywords are now
deprecated and will be removed in a future version of Terraform. Remove the
quotes surrounding this keyword to silence this warning.
```

lcavajani:
I'm closing this, I don't see any way right now to store the user somewhere to avoid dependencies.

PR #1118, opened 2020-05-22
change ssh_sock to use TERRAFORM_STACK_NAME

JanZerebecki:
## Why is this PR needed?

When ssh-agend -a NAME is called with a NAME that is too long it fails
with:
```
unix_listener: path "/home/jenkins/workspace/caasp-jobs/v4/e2e/vmware/test_upgrade_plan_from_previous_with_upgraded_control_plane-daily/agent.sock" too long for Unix domain socket
```

Fixes: SUSE/avant-garde#1596

## What does this PR do?

change ssh_sock to use TERRAFORM_STACK_NAME

This ensure the resulting PATH is a maximum of 1+3+1+70+1+10=86 bytes long.

---

make one different TERRAFORM_STACK_NAME the same

Otherwise it might cut the build number off, which is what ensures it
being uniqe.

Also make the README match.

Choosing a long JOBNAME that results in a long $WORKSPACE does make it fail.

Choosing a long JOBNAME that results in a long $WORKSPACE does not make it fail.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/README.md
+++ b/ci/infra/testrunner/README.md
@@ -356,10 +356,12 @@ skuba:
  
 #### Terraform
 
-It is advisable to use a unique id related to the job execution as the terraform stack name:
+It is advisable to use a unique id related to the job execution as the
+terraform stack name that doesn't contain slash or dash and is at a maximum 70
+bytes long:
 
 ```
-TERRAFORM_STACK_NAME  = "${JOB_NAME}-${BUILD_NUMBER}" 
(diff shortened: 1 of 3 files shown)


pablochacin:
I tested the failing test and it passed the previously failing stage:

Screenshot from 2020-05-25 10-22-15

PR #1121, opened 2020-05-26
Fix log message when service restart fail

jvanz:
## Why is this PR needed?

The call to the "log" method to notify the user that the service restart
failed does not format the message adding the service name. This commit
fixes that adding a call to the "format" method of the string.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/skuba-update/skuba_update/skuba_update.py
+++ b/skuba-update/skuba_update/skuba_update.py
@@ -68,9 +68,8 @@ def main():
 
     # Check that we have the proper zypper version.
     if not check_version('zypper', REQUIRED_ZYPPER_VERSION):
-        raise Exception('zypper version {0} or higher is required'.format(
-            '.'.join([str(x) for x in REQUIRED_ZYPPER_VERSION])
-        ))
+        version = '.'.join([str(x) for x in REQUIRED_ZYPPER_VERSION])
+        raise Exception(f"zypper version {version} or higher is required")
 
     if os.geteuid() != 0:
         raise Exception('root privileges are required to run this tool')
@@ -92,7 +91,7 @@ def parse_args():
 
     annotate_only_msg = \
         'Do not install any update, just annotate there are available updates'
-    version_msg = '%(prog)s {0}'.format(version())
+    version_msg = f"%(prog)s {version()}"
 
     parser = argparse.ArgumentParser(description='Updates a CaaSP node')
     parser.add_argument(
@@ -244,7 +243,8 @@ def restart_services():
     for service in result.output.splitlines():
         cmd = run_command(['systemctl', 'restart', service], needsOutput=False)
         if cmd.returncode != 0:
-            log('Warning! Service \'{0}\' restart returned non zero exit code')
+            log((f"Warning! Service '{service}' restart returned non zero "
+                 "exit code"))
 
 
 def is_zypper_error(code):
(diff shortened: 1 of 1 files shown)


skuba-update/skuba_update/skuba_update.py
```diff
@@ -244,7 +244,7 @@ def restart_services():
     for service in result.output.splitlines():
         cmd = run_command(['systemctl', 'restart', service], needsOutput=False)
         if cmd.returncode != 0:
-            log('Warning! Service \'{0}\' restart returned non zero exit code')
+            log('Warning! Service \'{0}\' restart returned non zero exit code'.format(service))
```
dannysauer, line 247: Starting with Python 3.6, "f-strings" are available which are faster than `string.format` and IMO easier to read.  I'd also prefer to switch to double quotes to avoid the escapes.  With that said, I wouldn't go back and change the code a second time just for this - I'm just sharing for the sake of future knowledge. :)
```suggestion
            log(f"Warning! Service '{service}' restart returned non zero exit code")
```

jodavis-suse:
LGTM too.  I like Danny's comment but this is ready to merge.

skuba-update/skuba_update/skuba_update.py
jvanz: I also prefer the f-string format. But I was just following the pattern already in the file. I`ll update both call then. ;)

jvanz: Done

skuba-update/skuba_update/skuba_update.py
```diff
@@ -244,7 +243,7 @@ def restart_services():
     for service in result.output.splitlines():
         cmd = run_command(['systemctl', 'restart', service], needsOutput=False)
         if cmd.returncode != 0:
-            log('Warning! Service \'{0}\' restart returned non zero exit code'.format(service))
+            log(f"Warning! Service \'{service}\' restart returned non zero exit code")
```
dannysauer, line 246: The inner single quotes don't need escaped here, thanks to switching to the outer double-quotes.  Python, unlike a lot of scripting languages, doesn't handle single or double quotes differently. :)

```python
>>> service='pie'
>>> f"I like \'{service}\' with ice cream"
"I like 'pie' with ice cream"
>>> f"I like '{service}' with ice cream"
"I like 'pie' with ice cream"
```

skuba-update/skuba_update/skuba_update.py
```diff
@@ -68,9 +68,8 @@ def main():
 
     # Check that we have the proper zypper version.
     if not check_version('zypper', REQUIRED_ZYPPER_VERSION):
-        raise Exception('zypper version {0} or higher is required'.format(
-            '.'.join([str(x) for x in REQUIRED_ZYPPER_VERSION])
-        ))
+        version = '.'.join([str(x) for x in REQUIRED_ZYPPER_VERSION])
+        raise Exception(f"zypper version {version} or higher is required")
```
dannysauer, line 72: I like this cleaner (and shorter) structure. :+1:

skuba-update/skuba_update/skuba_update.py
jvanz: Ops... I forgot that. Sorry.

jvanz:
I just pushed one commit fixing a typo.

jvanz:
ping

skuba-update/skuba_update/skuba_update.py
```diff
@@ -244,7 +243,8 @@ def restart_services():
     for service in result.output.splitlines():
         cmd = run_command(['systemctl', 'restart', service], needsOutput=False)
         if cmd.returncode != 0:
-            log('Warning! Service \'{0}\' restart returned non zero exit code')
+            log((f"Warning! Service '{service}' restart returned non zero "
+                 "exit code"))
```
dannysauer, line 247: This could probably just be "exited non-zero" to keep it on one line. 😂

PR #1148, opened 2020-06-10
Support In-tree Azure Cloud Provider

c3y1huang:
## Why is this PR needed?

Support skuba `cloud-provider=azure`.

Ref: https://github.com/SUSE/avant-garde/issues/1582

## What does this PR do?

When provide `cpi_enable = true` in terraform.tfvars sets 
* system managed identity to all cluster virtual machines.
* assign route-table to vnet subnet

When skuba init with `cloud-provider=azure` this populates Azure cloud folder and enables cluster bootstrap with Azure in-tree cloud provider. 

## Info for QA

[New test-cases for Azure as cloud provider](https://github.com/SUSE/caasp-test-cases/pull/44)

### Related info

* [Support Azure infra deployment](https://github.com/SUSE/avant-garde/issues/1496)
* [Azure Storage Integration - for Internal Preview Only](https://confluence.suse.com/x/poInHg)

```
Flags:
      --cloud-provider string        Enable cloud provider integration with the chosen cloud. Valid values: aws, openstack, vsphere
```

```
Flags:
      --cloud-provider string        Enable cloud provider integration with the chosen cloud. Valid values: aws, azure, openstack, vsphere
```

## Docs

[Add Azure cloud provider support document](https://github.com/SUSE/doc-caasp/pull/878)

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

Signed-off-by: Chin-Ya Huang <chin-ya.huang@suse.com>


pkg/skuba/actions/cluster/init/assets.go
```diff
@@ -27,8 +27,9 @@ var (
 This directory contains all supported cloud provider integration configurations
 and instructions.
  
-- [~Openstack~](openstack/README.md)
 - [~AWS~](aws/README.md)
+- [~Azure~](aws/README.md)
```
innobead, line 31: It shoud be `azure/README.md`

c3y1huang:
rebased.

ci/infra/azure/terraform.tfvars.example
```diff
@@ -90,3 +90,6 @@ authorized_keys = []
 # List of Virtual Network IDs to join via virual network peering
 # EXAMPLE:`
 # peer_virutal_network_id = ["vnet-id1", "vnet-id2"]
+
+# Enable CPI integration with vSphere
```
jenting, line 94: vSphere?

ci/infra/azure/variables.tf
```diff
@@ -173,3 +173,8 @@ variable "peer_virutal_network_id" {
   default     = []
   description = "IDs of a Virtual Network to connect to via a peering connection"
 }
+
+variable "cpi_enable" {
+  default     = false
+  description = "Enable CPI integration with Azure"
+}
```
jenting, line 180: I did not saw `cpi_enable` in AWS. So I don't know what's the purpose for let users determine `cpi_enable=false` because I thought for public cloud, the CPI always on.
Could you please elaborate more.

pkg/skuba/actions/cluster/init/assets.go
```diff
@@ -166,5 +167,52 @@ folder = "<VC_VM_FOLDER>"
...
+
+	azureReadme = strings.ReplaceAll(`# ~Azure~ integration
+
+Create a file inside this directory named ~azure.conf~, with the [supported contents](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/docs/cloud-provider-config.md):
+
+You can find a template named azure.conf.template inside this directory.
+
+~~~
+{
+    "cloud":"AzurePublicCloud",
```
jenting, line 180: nit 
```suggestion
    "cloud": "AzurePublicCloud",
```

pkg/skuba/actions/cluster/init/assets.go
```diff
@@ -166,5 +167,52 @@ folder = "<VC_VM_FOLDER>"
...
+
+## Important
+
+When the cloud provider integration is enabled, it's very important to bootstrap
+and join nodes with the node names same as ~Azure~ virtual machine's hostnames,
+as these names will be used by services to reconcile node metadata.
+`, "~", "`")
+
+	azureCloudConfTemplate = `{
+    "cloud":"AzurePublicCloud",
```
jenting, line 205: nit
```suggestion
    "cloud": "AzurePublicCloud",
```

ci/infra/azure/variables.tf
c3y1huang: Currently we support deployment with and without cloud provider. If all agree I can remove `cpi_enable` and consolidate the resource creation. WDYT @innobead , @cclhsu .

ci/infra/azure/terraform.tfvars.example
c3y1huang: Ugh! thanks.

pkg/skuba/actions/cluster/init/assets.go
c3y1huang: thanks

pkg/skuba/actions/cluster/init/assets.go
c3y1huang: thanks.

ci/infra/azure/variables.tf
innobead: > I did not saw `cpi_enable` in AWS. So I don't know what's the purpose for let users determine `cpi_enable=false` because I thought for public cloud, the CPI always on.
> Could you please elaborate more.

A good question to rethink a bit.

There is no AWS cloud provider integration yet in AWS TFs, so this is why `cpi_enable` not there. I think we are implicitly adopting the way coming from the initial examples of `openstack` and `vsphere`, so this is why `cpi_enable` introduced here.

Let's look around the market, `rancher` provides customers to decide if setting the cloud config of a specific infra for integrating the corresponding cloud provider.  ref: https://rancher.com/docs/rke/latest/en/config-options/cloud-providers/. However, openshift 4 seems not to disclose any details about it, but 3 does the way as Rancher RKE. 

I would suggest let's keep this design to have two options for users like rancher does, and look for any feedback in the future to see if we should stick on always enabling cloud provider integration. Also, this can keep the same design as what we already provide like `openstack` and `vsphere`.

- disable cloud provider, and use on-prem like deployment on cloud
- enable cloud provider, and use cloud provider integration

jenting: Cool, because I was thinking if _without_ cloud provider enabled in public cloud, how the user enables ingress & load balancer related function :thinking: .
But I'm lack of knowledge for this because my past experience was always enabled for the public cloud.

But as you mentioned we did not have AWS cloud provider integration yet in AWS TFs. But if I remember correctly, after AWS deployment, I could see the external IPs from `kubectl get nodes -o wide`. So I'm not sure if AWS TFs by default is _cpi_enable=true_ or _cpi_enable=false_.

I agree to keep current design but just curious what's the default value `cpi_enable` should be.

innobead: Yes, you are right. native ingress integration is provided by Azure cloud provider integration, ref: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/docs/services/README.md.

After discussing with @jenting again, some conclusions to follow up later on.
- deployment: skuba provides optionally enabling cloud provider (as what we have no by providing `--cloud-provider`)
- infra: should we have `cpi_enable` in TFs, or we should have agnostic terraforms scripts regardless if having cloud provider integration or not. To have better user experience. 

To sum up the current cloud providers we have right now. 
- `openstack`: `cpi_enable` is used to automatically generate and copy openstack.conf to each node.
- `vsphere`: cpi_enable` is used to manage `vsphere_folder` resource
- `azure`: @c3y1huang is still working on that. 

@c3y1huang let's see if it's possible to remove `cpi_enable`, and it can be working regardless `--cloud-provider` enabled or not. If yes, this always achieves what we discussed.

And, later we definitely need to revisits `vsphere` and `openstack` to make them consistent behavior.

c3y1huang:
Verified managed identity:
* file storage, disk storage; 
* in-line volume, persistent volume, storage class.

c3y1huang:
Verified sp secret:

* single node of each type, multi nodes
* file storage, disk storage;
* in-line volume, persistent volume, storage class.

ci/infra/azure/README.md
```diff
@@ -9,11 +9,25 @@ to deploy CaaS Platform on top of it.
...
+
+You can also use Azure CLI to automate the task. For installation instructions, refer to link:https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest[Install the Azure CLI].
+```bash
+az ad sp create-for-rbac --role="Contributor" --scopes="/subscriptions/<SUBSCRIPTION_ID>"
+```
+
+Update and source `container-openrc.sh` before deploying terraform script.
+```bash
+source container-openrc.sh
+```
```
innobead, line 28: Is this for Azure? it seems like for `openstack`?

ci/infra/azure/README.md
```diff
@@ -40,11 +54,7 @@ the [Azure Linux Agent](https://docs.microsoft.com/en-us/azure/virtual-machines/
 
 ### Using spot instances
 
-It's possible to create a [spot VMs](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/spot-vms)
-both for the master and worker nodes.
-
-This can be done by setting these variables to `true` (they are set to
-`false` by default):
+User can enable nodes to use [spot VMs](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/spot-vms) by setting below to `true` (`false` by default).
```
innobead, line 57:
```suggestion
User can enable nodes to use [spot VMs](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/spot-vms) by setting below variables to `true` (`false` by default).
```

ci/infra/azure/README.md
```diff
@@ -157,3 +167,17 @@ It is possible to join existing network to the cluster.  It can be setup by addi
 It is possible to enable multiple zone.  It can be set `enable_zone` to `true` and master/worker node will distribute sequentially based on zones defined in `azure_availability_zones`.
 
   * As of June 2020, the [Azure Availability Zones](https://docs.microsoft.com/en-us/azure/availability-zones/az-region) is not available in all Azure regions.
+
+#enable-cloud-provider-interface
+## Enable Cloud Provider Interface
+
+Configure `cpi_enable` to `true` enables terraform to provision the cluster with setups for cloud provider interface.
```
innobead, line 174:
```suggestion
Configuring `cpi_enable` to `true` enables terraform to provision the cluster with setups for the cloud provider interface.
```

ci/infra/azure/network.tf
```diff
@@ -66,3 +66,22 @@ resource "azurerm_private_dns_zone_virtual_network_link" "internal_dns" {
   virtual_network_id    = azurerm_virtual_network.virtual_network.id
   registration_enabled  = true
 }
+
+# Route Table
+resource "azurerm_route_table" "nodes" {
```
innobead, line 71: Are route resources here for managing routing rules by azure cloud provider?

ci/infra/azure/README.md
c3y1huang: Yes this is for Azure. I believe the idea did came from openstack, right @cclhsu ?
ref: https://github.com/SUSE/skuba/blob/master/ci/infra/azure/container-openrc.sh

ci/infra/azure/network.tf
c3y1huang: Yes need to exist for the CCM.

ci/infra/azure/README.md
innobead: Do we really need this? @cclhsu I think we already use the service principal to do the authentication instead of environment variables.

c3y1huang: There are two possible ways for terraform takes credential configs:
* environment variable (recommended) <- where we are
* in-line
ref: https://www.terraform.io/docs/providers/azurerm/guides/service_principal_client_secret.html

innobead: Yes, I misread here :(, the thing is correct as what we decided to do. Thanks for clarifying. All good now thanks @c3y1huang and @cclhsu .

cclhsu: Yes.  Same method is used for openstack and vsphere, and the question was also raised during demo (I think for vshpere) if this way is possible.

innobead: Thanks for clarifying @cclhsu :+1:

c3y1huang:
rebased.

pkg/skuba/actions/cluster/init/assets.go
```diff
@@ -27,8 +27,9 @@ var (
 This directory contains all supported cloud provider integration configurations
 and instructions.
  
-- [~Openstack~](openstack/README.md)
 - [~AWS~](aws/README.md)
+- [~Azure~](azure/README.md)
+- [~Openstack~](openstack/README.md)
 - [~vSphere~](vsphere/README.md)
```
jenting, line 34: Thanks for the sorting :heart:

ci/infra/azure/master-instance.tf
```diff
@@ -70,3 +77,16 @@ resource "azurerm_virtual_machine_extension" "master" {
     }
 SETTINGS
 }
+
+locals {
+  master_principal_ids = var.cpi_enable ? azurerm_linux_virtual_machine.master.*.identity.0.principal_id : []
+}
+
+resource "azurerm_role_assignment" "master" {
```
pgonin, line 85: What role are you assigning to the masters here ?

It should be an option to manage this with terraform, as it requires the Operator to have full control over the subscription, which is not an ideal security model (at least we wanted not QA engineers to have 'Owner' role over SUSE Engineering subscription as a concrete example)
Instead documentation should explain how the owner could provide the required configuration.

maximenoel8:
### QA information

#### Configuration
###### skuba
 - build from skuba / devel

###### Deployment 
- deploy from airgapped server with devel container images and packages
- zone is disable during deployment ( `"enable_zone": false,`) 
- lb / 2 master / 3 workers

###### Storage class 

```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: sample-disk-sc
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
provisioner: kubernetes.io/azure-disk
parameters:
  skuName: StandardSSD_LRS
  location: "West Europe"
```

#### Results

I was able to create pvc and deploy services using pvc.

maximenoel8:
### QA information

#### Configuration
###### skuba
 - build from skuba / devel

###### Deployment 
- deploy from airgapped server with devel container images and packages
- zone is enable during deployment ( `"enable_zone": true,` and `"azure_availability_zones": ["1","2","3"],`) 
- lb / 2 master / 3 workers

###### Storage class 

```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: sample-disk-sc
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
provisioner: kubernetes.io/azure-disk
parameters:
  skuName: StandardSSD_LRS
  location: "West Europe"
```

#### Results

Using the storage class is failing.


#### Error 
```
RetryAfter: 0s, HTTPStatusCode: 400, RawError: {
  "error": {
    "code": "BadRequest",
    "message": "Disk /subscriptions/c011786b-59d7-4817-880c-7cd8a6ca4b19/resourceGroups/mnoel-cluster-x8rc-1-resource-group/providers/Microsoft.Compute/disks/cluster-dynamic-pvc-a63a9428-26fb-47a7-b96c-c7409b86b938 cannot be attached to the VM because it is not in zone '2'."
  }
}
```

innobead:
> ### QA information
> #### Configuration
> ###### skuba
> * build from skuba / devel
> 
> ###### Deployment
> * deploy from airgapped server with devel container images and packages
> * zone is enable during deployment ( `"enable_zone": true,` and `"azure_availability_zones": ["1","2","3"],`)
> * lb / 2 master / 3 workers
> 
> ###### Storage class
> ```
> kind: StorageClass
> apiVersion: storage.k8s.io/v1
> metadata:
>   name: sample-disk-sc
>   annotations:
>     storageclass.kubernetes.io/is-default-class: "true"
> provisioner: kubernetes.io/azure-disk
> parameters:
>   skuName: StandardSSD_LRS
>   location: "West Europe"
> ```
> 
> #### Results
> Using the storage class is failing.
> 
> #### Error
> ```
> RetryAfter: 0s, HTTPStatusCode: 400, RawError: {
>   "error": {
>     "code": "BadRequest",
>     "message": "Disk /subscriptions/c011786b-59d7-4817-880c-7cd8a6ca4b19/resourceGroups/mnoel-cluster-x8rc-1-resource-group/providers/Microsoft.Compute/disks/cluster-dynamic-pvc-a63a9428-26fb-47a7-b96c-c7409b86b938 cannot be attached to the VM because it is not in zone '2'."
>   }
> }
> ```

@maximenoel8 please help create an issue for this failed case. Thanks for the regression :+1: 

cc @c3y1huang @cclhsu @jenting

PR #1152, opened 2020-06-10
Add kucero addon to auto rotates control plane certificates

jenting:
## Why is this PR needed?

Add kucero as an addon to auto rotates control plane certificates.

kucero includes two components:
- A Daemonset to pulling and rotates the kubeadm-managed certificates.
- A Controller to sign the kubelet server CSR.

refs: https://github.com/SUSE/avant-garde/issues/1632

## What does this PR do?

Add kucero as an addon.

We add kucero on both `1.17.4` and `1.18.2` because we want kucero deployed before node upgrade. Add kucero into the previous version make the user have to install kucero addon first.
Otherwise, the kubelet server CSR will in pending state since no signer signs the kubelet server certificate.

## Info for QA

Previously, the kubelet server certificate is self-signed by skuba with kubelet CA cert/key pair.
With this PR, the kubelet sends server CSR in K8s cluster, and kucero signed the server certificate to kubelet with kubelet CA cert/key pair, then kubelet save the server certificate in `/var/lib/kubelet/pki/kubelet-server-current.pem`.

Note that, the metrics-server trusted kubelet server certificate by kubelet CA cert.
If the kubelet server certificate signed by other CA, the metrics-server will in CrashLoopBack.

### Related info

Packages to be released:
* https://build.suse.de/package/show/Devel:CaaSP:5/kucero
* https://build.suse.de/package/show/Devel:CaaSP:5:Containers:CR/caasp-kucero-image

1. The kubeadm-managed server/client certificates will expire after 1 year.
2. The kubelet server certificate is self-signed by skuba with kubelet CA cert/key pair.

1. The kucero will periodically checks the kubeadm-managed server/client certificates residual time, and rotates the certificates if necessary automatically, and restart kubelet after rotates certificates finished.
2. The kubelet will set `serverTLSBootstrap: true`, and kubelet will send server CSR. Then kucero will sign the kubelet server certificate by kubelet CA cert/key pair and kubelet saves the server certificate in `/var/lib/kubelet/pki/kubelet-server-current.pem`. The time kubelet send server CSR is handled by kubelet daemon.

## Docs

https://github.com/SUSE/doc-caasp/pull/879

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/kubelet.go
+++ b/internal/pkg/skuba/deployments/ssh/kubelet.go
@@ -18,17 +18,11 @@
 package ssh
 
 import (
-	"crypto/x509"
 	"io/ioutil"
-	"net"
 	"os"
 	"path/filepath"
-	"strings"
 
-	"github.com/pkg/errors"
-	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/kubernetes/cmd/kubeadm/app/constants"
-	"k8s.io/kubernetes/cmd/kubeadm/app/util/pkiutil"
 	"sigs.k8s.io/yaml"
 
 	"github.com/SUSE/skuba/internal/pkg/skuba/deployments"
@@ -39,7 +33,6 @@ import (
 
 func init() {
 	stateMap["kubelet.rootcert.upload"] = kubeletUploadRootCert
-	stateMap["kubelet.servercert.create-and-upload"] = kubeletCreateAndUploadServerCert
 	stateMap["kubelet.configure"] = kubeletConfigure
 	stateMap["kubelet.enable"] = kubeletEnable
 }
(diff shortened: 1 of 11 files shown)


internal/pkg/skuba/deployments/ssh/kubelet.go
```diff
@@ -62,84 +55,6 @@ func kubeletUploadRootCert(t *Target, data interface{}) error {
...
-		return err
-	}
-	if err := os.Remove(keyPath); err != nil {
-		return err
-	}
-
-	return nil
-}
-
 func kubeletConfigure(t *Target, data interface{}) error {
```
innobead, line 58: Removing `kubelet server cert creation and upload` is because we enable `server rotation` so it's expected everything will be setup automatically?

jenting: yes, I intended to add kucero at previous version `1.17.4` is because I want to install kucero first and enable kubelet `serverTLSBootstrap: true` when cluster upgrade.

innobead:
cc @c3y1huang @cclhsu please help review. thanks.

PR #1156, opened 2020-06-11
Schedule daily jobs by version

pablochacin:
## Why is this PR needed?

The introduction of the `maintenance-caas-v4` branch duplicates the number of daily jobs executed. This may cause resource contention if the jobs of both versions are scheduled concurrently.

This change extends the daily job execution window. As the development team is spread across multiple time zones, it is desirable to prevent the daily jobs delay PR jobs. 

Fixes https://github.com/SUSE/avant-garde/issues/1689

## What does this PR do? 

Schedule the jobs of each version in a different time range.

Use labels to ensure jobs are executed in dedicated workers, to avoid daily jobs to delay pr jobs.
 
# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/jenkins/jobs.yaml
+++ b/ci/jenkins/jobs.yaml
@@ -21,6 +21,7 @@
           - vmware
       version:
           - v4:
+              schedule: '1-3'
               branch: maintenance-caasp-v4 
           - v5
       jobs:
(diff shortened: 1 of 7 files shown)


jordimassaguerpla:
This is the error

Error: timeout waiting for an available IP address

jordimassaguerpla:
Actually this is https://github.com/SUSE/avant-garde/issues/1682 and this PR is to fix this.
@pablochacin should I force merge this?

PR #1162, opened 2020-06-12
Enable toggling of the CNI plugin used in the cluster

rktidwell:
## Why is this PR needed?

Introduce special handling of CNI addons
    
This change introduces special handling of addons that introduce CNI plugins. The change introduces the following new functionality:
    
    - Introduce '--cni-plugin' flag for use with 'skuba cluster init'
    - Addons must now declare whether they provide a CNI plugin or not
    - The user-supplied CNI plugin request is checked against addons that provide
      CNI plugins
    - If '--cni-plugin' is not specified, cilium selected as the default CNI
      addon. This maintains backward-compatible behavior for 'skuba cluster init'
    - Rather than write config files for all CNI plugin addons,
      configuration files are only written for the selected CNI addon

This enables future support of CNI plugins beyond Cilium.

## What does this PR do?

This introduces a new optional flag for `skuba cluster init` that allows for specifying the CNI plugin to configure in the cluster. This change does not add suport for any new CNI's, it simply creates the framework by which new CNI plugins can be integrated. Support for new CNI plugins will be handled with future PR's.

`--cni-plugin` is an optional flag, and at this time only allows `cilium` to be passed as an option. These changes are backward-compatible: invoking `skuba cluster init` without `--cni-plugin` will emit manifests for Cilium as has been previously done. As more CNI plugins are added, they can be specified using `skuba cluster init --cni-plugin <CNI plugin name>`.

This PR adds automated test cases. All tests we have had in place are still valid.

## Info for QA

N/A

### Related info

* https://github.com/SUSE/avant-garde/issues/1673

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/cmd/skuba/cluster/init.go
+++ b/cmd/skuba/cluster/init.go
@@ -32,6 +32,7 @@ type initOptions struct {
 	KubernetesVersion string
 	CloudProvider     string
 	StrictCapDefaults bool
+	CniPlugin         string
 }
 
 // NewInitCmd creates a new `skuba cluster init` cobra command
@@ -47,7 +48,8 @@ func NewInitCmd() *cobra.Command {
 				initOptions.CloudProvider,
 				initOptions.ControlPlane,
 				initOptions.KubernetesVersion,
-				initOptions.StrictCapDefaults)
+				initOptions.StrictCapDefaults,
+				initOptions.CniPlugin)
 			if err != nil {
 				klog.Fatalf("init failed due to error: %s", err)
 			}
@@ -74,6 +76,8 @@ func NewInitCmd() *cobra.Command {
 	_ = cmd.MarkFlagRequired("control-plane")
 
 	cmd.Flags().BoolVar(&initOptions.StrictCapDefaults, "strict-capability-defaults", false, "All the containers will start with CRI-O default capabilities")
+
+	cmd.Flags().StringVar(&initOptions.CniPlugin, "cni-plugin", "cilium", "Specify the CNI plugin to be used across the cluster. Valid values: cilium")
 
 	return cmd
 }
--- a/internal/pkg/skuba/addons/README.md
+++ b/internal/pkg/skuba/addons/README.md
@@ -20,9 +20,21 @@ https://golang.org/doc/effective_go.html#init
 ## How to create a new Addon
 
 To create a new addon, check the rest of the addons source file to
-understand what is required. Again, as long as the addon is in this
-directory, it will get called.
-
-You should also express the version mapping with kubernetes in:
+understand what is required. You should express the version mapping
+with kubernetes in:
 
 https://github.com/SUSE/skuba/blob/master/internal/pkg/skuba/kubernetes/versions.go
+
+Each addon must also declare its `AddOnType`. Any number of addons of type
+`CniAddOn`can be created. However, only a single addon of type `CniAddOn`
+can be used in the cluster and as such only one addon providing a CNI
+plugin will be activated and deployed.
+
+The CNI addon that is activated defaults to cilium, but can be toggled by
+passing `--cni-plugin` to `skuba cluster init`.
+
+Example:
+
+```sh
+skuba cluster init --control-plane load-balancer.example.com --cni-plugin cilium company-cluster
+```
--- a/internal/pkg/skuba/addons/addons.go
+++ b/internal/pkg/skuba/addons/addons.go
@@ -74,15 +74,23 @@ patches:
 `
 )
 
+const (
+	CniAddOn     AddOnType = "CNI"
+	GenericAddOn AddOnType = "GENERIC"
+)
+
 var Addons = map[kubernetes.Addon]Addon{}
 
+type AddOnType string
+
 type Addon struct {
-	addon              kubernetes.Addon
+	Addon              kubernetes.Addon
 	templater          addonTemplater
 	preflightTemplater preflightAddonTemplater
 	callbacks          addonCallbacks
 	addonPriority      addonPriority
 	getImageCallbacks  []getImageCallback
+	AddOnType          AddOnType
 }
 
 type addonCallbacks interface {
@@ -110,7 +118,7 @@ func (renderContext renderContext) AnnotatedVersion() string {
 }
 
 func (renderContext renderContext) ManifestVersion() string {
-	addonVersion := kubernetes.AddonVersionForClusterVersion(renderContext.addon.addon, renderContext.config.ClusterVersion)
+	addonVersion := kubernetes.AddonVersionForClusterVersion(renderContext.addon.Addon, renderContext.config.ClusterVersion)
 	if addonVersion == nil {
 		return ""
 	}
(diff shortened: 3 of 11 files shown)


rktidwell:
> I am wondering about upgrades: we run a command that upgrades all the addons and I fear that, if we have CNI_A and CNI_B, then the system will apply both addons in an upgrade and thus deploy both CNIs.

Yes, upgrades will need to be handled with some special-casing as well. This PR isn't meant to address the upgrade case, although it might not be too painful to include those changes here. I should update the commit messages and such to reflect that.

internal/pkg/skuba/addons/addons.go
```diff
@@ -83,6 +84,7 @@ type Addon struct {
 	callbacks          addonCallbacks
 	addonPriority      addonPriority
 	getImageCallbacks  []getImageCallback
+	isCniAddon         bool
```
pablochacin, line 87: Maybe it could be generalized as `addOnType AddOnType` and define 
```
type AddOnType string

const (
     CniAddOn AddOnType "CNI"
)
```
In this way, if in the future other types of addons are supported, it won't be necessary to maintain multiple flags, but add more

internal/pkg/skuba/addons/addons.go
```diff
@@ -119,14 +121,27 @@ func (renderContext renderContext) ManifestVersion() string {
...
-func registerAddon(addon kubernetes.Addon, addonTemplater addonTemplater, preflightAddonTemplater preflightAddonTemplater, callbacks addonCallbacks, addonPriority addonPriority, getImageCallbacks []getImageCallback) {
-	Addons[addon] = Addon{
-		addon:              addon,
-		templater:          addonTemplater,
-		preflightTemplater: preflightAddonTemplater,
-		callbacks:          callbacks,
-		addonPriority:      addonPriority,
-		getImageCallbacks:  getImageCallbacks,
+func registerAddon(addon kubernetes.Addon, isCniAddon bool, addonTemplater addonTemplater, preflightAddonTemplater preflightAddonTemplater, callbacks addonCallbacks, addonPriority addonPriority, getImageCallbacks []getImageCallback) {
+	if isCniAddon {
```
pablochacin, line 125: It scapes to me totally why you need two maps of addons with the same structure. Maybe will be enough and you can filter by addon type.

internal/pkg/skuba/addons/addons.go
rktidwell: Good idea, I'll make that change.

internal/pkg/skuba/addons/addons.go
rktidwell: My thinking was that maintaining 2 maps would make it easier to perform logic specific to CNI plugins. On second thought it really is a bad idea. What is really needed is logic that filters addons, so I'll continue maintaining a single map and develop that addon filtering logic.

internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -33,6 +33,7 @@ type Component string
 
 const (
 	Cilium        Addon = "cilium"
+	CNI           Addon = "cni"
```
jenting, line 36: I assume cilium is one of the CNI, so having two Cilium and CNI both types is Addon seems a bit weird to me.

pkg/skuba/actions/cluster/init/init_test.go
```diff
@@ -142,6 +144,68 @@ func TestInitCloudProvider(t *testing.T) {
 	}
 }
 
+func TestInitCiliumCniPlugin(t *testing.T) {
```
jenting, line 147: Considering consolidates the two func `TestInitCiliumCniPlugin` and `TestInitUnknownCniPlugin` to one and use table-driven test cases.

pkg/skuba/actions/cluster/init/init_test.go
```diff
@@ -142,6 +144,68 @@ func TestInitCloudProvider(t *testing.T) {
...
+		t.Errorf("unexpected error: %v", err)
+	}
+	defer switchBackAndCleanFilesystem(ctx)
+	ciliumManifestPath := "addons/" + ciliumCni + "/base/" + ciliumCni + ".yaml"
+	found, err := doesFileExist(ctx, clusterName, ciliumManifestPath)
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
+	}
+	if !found {
+		t.Error("Cilium manifests not properly rendered")
```
jenting, line 178: If an error happens, it should return. Otherwise, the following code would run.

internal/pkg/skuba/kubernetes/versions.go
rktidwell: Oops, that got left around as an artifact from some experimentation earlier. Good catch.

pkg/skuba/actions/cluster/init/init_test.go
rktidwell: Done

pkg/skuba/actions/cluster/init/init_test.go
rktidwell: Done

cmd/skuba/cluster/init.go
```diff
@@ -65,5 +67,7 @@ func NewInitCmd() *cobra.Command {
 
 	cmd.Flags().BoolVar(&initOptions.StrictCapDefaults, "strict-capability-defaults", false, "All the containers will start with CRI-O default capabilities")
 
+	cmd.Flags().StringVar(&initOptions.CniPlugin, "cni-plugin", "cilium", "Specify the CNI plugin to be used across the cluster. Valid values: cilium")
```
soumynathan, line 70: Should the help sring say, 'default is 'cilium'. ?

internal/pkg/skuba/addons/README.md
```diff
@@ -20,9 +20,21 @@ https://golang.org/doc/effective_go.html#init
...
-understand what is required. Again, as long as the addon is in this
-directory, it will get called.
-
-You should also express the version mapping with kubernetes in:
+understand what is required. You should express the version mapping
+with kubernetes in:
 
 https://github.com/SUSE/skuba/blob/master/internal/pkg/skuba/kubernetes/versions.go
+
+If the addon provides a CNI plugin, this must declared when invoking
```
soumynathan, line 28: 'this must be'

cmd/skuba/cluster/init.go
rktidwell: That would be a nice touch. I'll add that.

internal/pkg/skuba/addons/README.md
rktidwell: Good catch.

cmd/skuba/cluster/init.go
jenting: It already fills the default value in "cilium", then the library "github.com/spf13/cobra" generated the default value in command help for you. So we don't need to say 'default is cilium' in help string. Otherwise, the modified current help string would be
```
Specify the CNI plugin to be used across the cluster. Default is 'cilium'. Valid values: cilium (default "cilium")
```

rktidwell:
> Please fix linter error, run `make lint && git status`

`make lint` has been succeeding locally. CI is complaining about not having go 1.13 in go.mod, which is unrelated to this change. Do I need to rebase? Are we adding this to go.mod now?

rktidwell:
> > Please fix linter error, run `make lint && git status`
> 
> `make lint` has been succeeding locally. CI is complaining about not having go 1.13 in go.mod, which is unrelated to this change. Do I need to rebase? Are we adding this to go.mod now?

I went ahead and rebased. Looks like some changes were pushed between when I created my fork and creating this PR. Things are looking better now.

internal/pkg/skuba/addons/addons.go
```diff
@@ -225,19 +234,26 @@ func (addon Addon) HasToBeApplied(addonConfiguration AddonConfiguration, skubaCo
 		//      now, just return that it doesn't have to be applied.
 		return false, nil
 	}
+	// Check whether this is a CNI addon and whether its base config has been rendered.
+	// If it's a CNI plugin and the base config has not been rendered, we can assume
+	// the user requested a different CNI plugin and this addon does not need to be
+	// applied.
+	if info, _ := os.Stat(addon.addonDir()); addon.AddOnType == CniAddOn && !info.IsDir() {
```
flavio, line 241: Shouldn't you look also at the `error` returned by `os.Stat` and handle that? Especially if `os.IsNotExist(err)`

rktidwell: Yep, fixed this.

cmd/skuba/cluster/init.go
rktidwell: Yep, just noticed this when doing some testing. I reverted this change and the help from the CLI should now look correct.

PR #1169, opened 2020-06-17
Tumbleweed container has no diff cmd

evrardjp:
The new tumbleweed images don't have the diff command by default.
This quick hack makes sures to install diffutils if necessary
when the diff command is used.

Partial-Fix: https://github.com/SUSE/avant-garde/issues/1660

### DIFF
--- a/skuba-update/test/os/suse/suse.sh
+++ b/skuba-update/test/os/suse/suse.sh
@@ -116,6 +116,7 @@ check_kubectl_calls() {
         for i in "$@"; do
             echo "$i" >> /commands.txt
         done
+        type -p diff || zypper -n in diffutils
         diff -w /commands.txt /tmp/kubectl-commands &> /dev/null
     fi
 }


evrardjp:
1. I think the intent was to support tumbleweed to catch if a change in tw would trigger a breakage here. One could argue that this isn't the place to test those, and whether we could extend zypper test coverage to solve our issues.
2. Yes that's definitely confusing,  validate_ could probably be a better wording. Yet, I wanted this PR to be the strictly smallest bit of change to get things done.
3. There is only one way to be expert in it: Just get involved. You're doing it, so you're in the right path :) To what I quickly have seen (I am not expert either) skuba-update has plenty of areas to improve. Those can be done in a follow-up commit, however.

evrardjp:
Rebased to avoid lint issues due to version differences.

skuba-update/test/os/suse/suse.sh
```diff
@@ -116,6 +116,7 @@ check_kubectl_calls() {
         for i in "$@"; do
             echo "$i" >> /commands.txt
         done
+        command -v diff || zypper install -y diffutils
```
dirkmueller, line 119: there is a install_package() wrapper that calls zypper with the right arguments (namely -n). 

so suggest to use

```bash
 command -v diff || install_package diffutils
```

evrardjp: meh that will reinstall other package by default...

evrardjp:
I didn't have tech-debt cards, when I saw a problem, I just go fix it.

dannysauer:
Some alternate solutions to consider:

Instead of installing diff, this could be pretty easily done by simply reading the lines from `/tmp/kubectl_commands` in to an associative array, and then deleting each matching entry in the array when it's read from the list.  Add an entry if no match.  If the array has elements at the end, there's a difference.  [Examples of working with associative arrays in Bash](https://www.artificialworlds.net/blog/2012/10/17/bash-associative-array-examples/) are easy to find. :)

Or, since everyone hates actually using code in bash scripts :rofl:, this could be done with `grep`.  Use file1 as a list of patterns to grep from file2, and then vice-versa.  The return code of `grep -v` is 0 if non-matching lines were found, or 1 if all lines matched patterns (aka, no output).  So, `x` for whole line match, `F` for fixed strings (don't treat commands as regexes), `v` for removing matching lines, and `f` to specify the file with a list of patterns:
```
grep -xFvf /commands.txt /tmp/kubectl-commands && exit 1
grep -xFvf /tmp/kubectl-commands /commands.txt && exit 1
```

Feel free to ignore both of those options.  In that case, I'll still note that I personally prefer `type -p diff` since that gives the same true/false error code, but prints nothing if `diff` is a file or not found (only prints if it's an alias or whatever). ;)  This is fine, too, but avoiding the time associated with installing a tool for just one use might be of value.

evrardjp:
@dannysauer Feel free to propose a patch! I don't mind, quite the opposite... :)

skuba-update/test/os/suse/suse.sh
```diff
@@ -116,6 +116,7 @@ check_kubectl_calls() {
         for i in "$@"; do
             echo "$i" >> /commands.txt
         done
+        command -v diff || install_package diffutils
```
dirkmueller, line 119: actually this fails, as the first argument is the repository name (no idea who came up with this.. )

so it should be


```suggestion
        type -p diff || install_package base diffutils
```

dirkmueller:
unfortunately it still does not work:

```

+ install_package base diffutils

+ zypper -n in -r base diffutils

Loading repository data...

Reading installed packages...

'diffutils' not found in package names. Trying capabilities.

No provider of 'diffutils' found.```

File: skuba-update/test/os/suse/suse.sh [hunk 3]
```diff
@@ -116,6 +116,7 @@ check_kubectl_calls() {
         for i in "$@"; do
             echo "$i" >> /commands.txt
         done
+        type -p diff || install_package base diffutils
```
dirkmueller, line 119:
```suggestion
        type -p diff || zypper -n in diffutils
```

evrardjp: https://media.giphy.com/media/l1J3QMgxq4SS8EHXa/giphy.gif

evrardjp:
FYI: I don't really care enough about that commit to the point I would be babysitting it :p

dirkmueller:
@evrardjo well, this fixe a ci fail in one of my PRs, so I am trying to get it in. 

It looks like things are finally passing.

I am not sure what the policy is but we may have to squash the commits for this to be merged.

evrardjp:
Squashed and rebased. Need new reviews. @dirkmueller @Itxaka could you have a looksie?

PR #1173, opened 2020-06-18
Do not render empty client secret (bsc#1173055)

jenting:
## Why is this PR needed?

`skuba cluster init` would render an empty client secret to dex&gangway.
It might cause an issue when the admin wants to re-install the dex&gangway from generated manifested directly with `kubectl apply -f addons/dex/base/dex.yaml` & `kubectl apply -f addons/gangway/base/gangway.yaml`.

Furthermore, the admin would use kustomize the oidc dex connector by copy `addons/dex/base/dex.yaml` to `addon/dex/patches/dex-patch.yaml`, and modifies connector in `dex-patch.yaml`, this would cause the client secret is empty in the dex server (unmatched with gangway).

Fixes #1173055

## What does this PR do?

* render the client secret when `skuba cluster init` by checking the _client_ is nil or not.
* add retry function when get existed oidc-gangway-config Configmap, this prevents the client secret might regenerate when installing addon due to network issue.

This PR does not address using the single source of the client secret since dex supports it until `v2.23.0`, we'll address it in another PR. Therefore, there might be a case that the admin change either oidc-dex _or_ oidc-gangway client secret in each Configmap, makes the client secret out of sync.

NOTE: this PR is for CaaSP v4 only, for CaaSP v5 we'd have another better approach for this.

## Info for QA

### Related info

xref https://github.com/SUSE/avant-garde/issues/1721

run `skuba cluster init`, the client secret is empty in `addons/dex/base/dex.yaml` & `addons/gangway/base/gangway.yaml`.

run `skuba cluster init`, the client secret is _not_ empty in `addons/dex/base/dex.yaml` & `addons/gangway/base/gangway.yaml`.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/addons/gangway.go
+++ b/internal/pkg/skuba/addons/gangway.go
@@ -40,11 +40,11 @@ func (renderContext renderContext) GangwayImage() string {
 }
 
 func (renderContext renderContext) GangwayClientSecret() string {
+	// ignore error for `skuba cluster init` case
 	client, err := kubernetes.GetAdminClientSet()
 	if err != nil {
-		return ""
+		return gangway.GetClientSecret(nil)
 	}
-
 	return gangway.GetClientSecret(client)
 }
 
--- a/internal/pkg/skuba/addons/gangway_test.go
+++ b/internal/pkg/skuba/addons/gangway_test.go
@@ -21,8 +21,6 @@ import (
 	"regexp"
 	"testing"
 
-	"k8s.io/client-go/kubernetes/fake"
-
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
 	"github.com/SUSE/skuba/internal/pkg/skuba/skuba"
 	img "github.com/SUSE/skuba/pkg/skuba"
@@ -91,11 +89,9 @@ func Test_renderContext_GangwayImage(t *testing.T) {
 func Test_renderContext_GangwayClientSecret(t *testing.T) {
 	for _, ver := range kubernetes.AvailableVersions() {
 		t.Run("get gangway client secret when cluster version is "+ver.String(), func(t *testing.T) {
-			fake.NewSimpleClientset()
-
 			got := renderContext{}.GangwayClientSecret()
-			if got != "" {
-				t.Errorf("expect got client secret empty")
+			if got == "" {
+				t.Errorf("expect got client secret")
 				return
 			}
 		})
--- a/internal/pkg/skuba/gangway/gangway.go
+++ b/internal/pkg/skuba/gangway/gangway.go
@@ -91,6 +91,12 @@ func GetClientSecret(client clientset.Interface) string {
 		return gClientSecret
 	}
 
+	// generate a new client secret when `skuba cluster init`
+	if client == nil {
+		gClientSecret = dex.GenerateClientSecret()
+		return gClientSecret
+	}
+
 	// global client secret not exist, read from ConfigMap
 	clientSecret, err := getClientSecretFromConfigMap(client)
 	if err != nil || clientSecret == "" {
--- a/internal/pkg/skuba/gangway/gangway_test.go
+++ b/internal/pkg/skuba/gangway/gangway_test.go
@@ -151,6 +151,10 @@ customHTMLTemplatesDir: /usr/share/caasp-gangway/web/templates/caasp
 				},
 			}),
 		},
+		{
+			name:   "nil clientset",
+			client: nil,
+		},
 	}
 
 	for _, tt := range tests {


internal/pkg/skuba/addons/gangway.go
```diff
@@ -40,11 +41,8 @@ func (renderContext renderContext) GangwayImage() string {
 }
 
 func (renderContext renderContext) GangwayClientSecret() string {
-	client, err := kubernetes.GetAdminClientSet()
-	if err != nil {
-		return ""
-	}
-
+	// ignore error for `cluster init` case
+	client, _ := kubernetes.GetAdminClientSet()
```
innobead, line 45: I would suggest to handle err, then if err exists, client = nil. 

Because sometimes we can not expect how the things will be handled in `kubernetes.GetAdminClientSet`to cause possible side effect.

internal/pkg/skuba/addons/gangway.go
```diff
@@ -54,12 +55,7 @@ func renderGangwayTemplate(addonConfiguration AddonConfiguration) string {
 
 type gangwayCallbacks struct{}
 
-func (gangwayCallbacks) beforeApply(addonConfiguration AddonConfiguration, skubaConfiguration *skuba.SkubaConfiguration) error {
-	client, err := kubernetes.GetAdminClientSet()
```
Klaven, line 58: :+1: for the dependency injection.

internal/pkg/skuba/addons/gangway_test.go
```diff
@@ -91,11 +91,9 @@ func Test_renderContext_GangwayImage(t *testing.T) {
 func Test_renderContext_GangwayClientSecret(t *testing.T) {
 	for _, ver := range kubernetes.AvailableVersions() {
 		t.Run("get gangway client secret when cluster version is "+ver.String(), func(t *testing.T) {
-			fake.NewSimpleClientset()
-
 			got := renderContext{}.GangwayClientSecret()
-			if got != "" {
-				t.Errorf("expect got client secret empty")
+			if got == "" {
+				t.Errorf("expect got non-empty client secret")
```
Klaven, line 96: NIT: (don't need to fix but if there is something else you already are changing the code for)
`got` in this context is confusing `expected non-empty client secret` the original was a rough read too. :) so no worries! I think it originally was meaning to substitute in the var `got` but... it's also in a test so really don't worry about it if you don't need to update this PR.

internal/pkg/skuba/gangway/gangway.go
```diff
@@ -91,6 +94,12 @@ func GetClientSecret(client clientset.Interface) string {
 		return gClientSecret
 	}
 
+	// generate a new client secret when `skuba cluster init`
+	if client == nil {
```
Klaven, line 98: will `getClientSecretFromConfigMap` not return an error if `client` is nil? if it does, this is not needed and seems to just be extra logic?

internal/pkg/skuba/gangway/gangway.go
```diff
@@ -110,12 +119,27 @@ func getClientSecretFromConfigMap(client clientset.Interface) (string, error) {
 		ClientSecret string `yaml:"clientSecret"`
 	}
 
-	cm, err := client.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(configmapName, metav1.GetOptions{})
```
Klaven, line 113: maybe for my above we should check `client` for being nil :)

internal/pkg/skuba/addons/dex.go
```diff
@@ -45,14 +46,7 @@ func renderDexTemplate(addonConfiguration AddonConfiguration) string {
 
 type dexCallbacks struct{}
 
-func (dexCallbacks) beforeApply(addonConfiguration AddonConfiguration, skubaConfiguration *skuba.SkubaConfiguration) error {
-	var err error
```
dannysauer, line 49: Do you no longer need `var err` here, since `err` is still used  in the cert exists check?

internal/pkg/skuba/gangway/gangway.go
```diff
@@ -110,12 +119,27 @@ func getClientSecretFromConfigMap(client clientset.Interface) (string, error) {
...
+	// adds the retry mechanism to get the gangway config map
+	var lastErr error
+	var cm *v1.ConfigMap
+	err := wait.ExponentialBackoff(retry.DefaultBackoff, func() (bool, error) {
+		var err error
+		cm, err = client.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(configmapName, metav1.GetOptions{})
+		if err != nil {
+			lastErr = err
+			if apierrors.IsNotFound(err) {
+				// return err to immediately leaves
```
dannysauer, line 132: Minor grammar
```suggestion
				// return err to immediately leave
```

internal/pkg/skuba/addons/dex.go
dannysauer: I guess we don't declare variables anywhere else, so...  Nevermind. :)

internal/pkg/skuba/addons/cilium.go
```diff
@@ -79,9 +81,9 @@ func renderCiliumPreflightTemplate(addonConfiguration AddonConfiguration) string
 
 type ciliumCallbacks struct{}
 
-func (ciliumCallbacks) beforeApply(addonConfiguration AddonConfiguration, skubaConfiguration *skuba.SkubaConfiguration) error {
+func (ciliumCallbacks) beforeApply(client clientset.Interface, addonConfiguration AddonConfiguration, skubaConfiguration *skuba.SkubaConfiguration) error {
 	ciliumVersion := kubernetes.AddonVersionForClusterVersion(kubernetes.Cilium, addonConfiguration.ClusterVersion).Version
-	client, config, err := kubernetes.GetAdminClientSetWithConfig()
+	_, config, err := kubernetes.GetAdminClientSetWithConfig()
 	if err != nil {
 		return errors.Wrap(err, "unable to get admin client set")
```
cmurphy, line 88: The error message is slightly misleading now since it's now only getting the config and not the clientset

jenting:
Thanks everyone review. After rethinking, we should focus on the issue `skuba cluster init` renders an empty client secret.
So, this PR only addresses this issue.

internal/pkg/skuba/gangway/gangway.go
jenting: An exception happens when executes
`client.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(configmapName, metav1.GetOptions{})`.

maximenoel8:
I checked the patch. For me the connection is still failing when trying to login. The secret in gangway and dex are different.
selenium-screenshot-3

PR #1174, opened 2020-06-18
Single source of oidc client secret (bsc#1173055)

jenting:
## Why is this PR needed?

After `dex >= v2.23.0`, it supports the client secret from the environment variable, it makes things easier to maintain a single source of oidc client secret between dex&gangway. It prevents client secret out of sync issues.

Fixes #1173055

## What does this PR do?

For dex, it generates the client-secret and stores in Secret `oidc-client-secret`, key `gangway`. And it uses new configuration `secretEnv` then pass the secret from environment variable `OIDC_CLIENT_SECRET`, this environment variable comes from Secret `oidc-client-secret`, key `gangway`.

For gangway, it use the same Secret `oidc-client-secret`, key `gangway`.

This make dex&gangway use single source of client secret.

Furthermore, this PR refactor the existing duplicates code as a unified one “oidc”.

We don't have the clientSecret in rendered addon manifests anymore.

NOTE: this PR is for CaaSP v5 only (dex>=2.23.0), there existed another PR to address on CaaSP v4.

## Info for QA

Greenfield & Brownfield both covered.

## Related info

xref: https://github.com/SUSE/avant-garde/issues/1721

The client secret existed in both dex&gangway Configmap.

The client secret stores in Secret `oidc-client-secret`, key `gangway`, dex&gangway use single source of oidc client secret.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


internal/pkg/skuba/addons/dex.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 SUSE LLC.
+ * Copyright (c) 2019,2020internal/pkg/skuba/addons/dex.go SUSE LLC.
```
c3y1huang, line 2:
```suggestion
 * Copyright (c) 2019,2020 SUSE LLC.
```

internal/pkg/skuba/addons/gangway.go
```diff
@@ -121,10 +113,9 @@ data:
     certFile: /etc/gangway/pki/tls.crt
 
     clientID: "oidc"
-    clientSecret: "{{.GangwayClientSecret}}"
     usernameClaim: "email"
     apiServerURL: "https://{{.ControlPlaneHostAndPort}}"
-    cluster_ca_path: "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+    clusterCAPath: "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
```
innobead, line 118: wow, this was what we missed before. good catch.

internal/pkg/skuba/oidc/resource_test.go
```diff
@@ -0,0 +1,147 @@
...
+					},
+					{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:   "dex-pod-3",
+							Labels: map[string]string{"app": "oidc-dex"},
+						},
+					},
+				},
+			}),
+			labelSelector: "app=oidc-gangway",
```
innobead, line 107: "app=oidc-dex"

internal/pkg/skuba/addons/dex.go
```diff
@@ -45,31 +46,44 @@ func renderDexTemplate(addonConfiguration AddonConfiguration) string {
...
+	// handles dex certificates
+	// TODO: should we keep rotates server certificate when addon upgrade?
+	exist, err = oidc.IsSecretExist(client, oidc.DexCertSecretName)
 	if err != nil {
 		return errors.Wrap(err, "unable to determine if dex certificate exists")
 	}
-	err = dex.CreateCert(client, skubaconstants.PkiDir(), skubaconstants.KubeadmInitConfFile())
-	if err != nil {
-		return errors.Wrap(err, "unable to create dex certificate")
+	if err = oidc.CreateServerCert(client, skubaconstants.PkiDir(), oidc.DexCertCN, addonConfiguration.ControlPlaneHost, oidc.DexCertSecretName); err != nil {
```
innobead, line 74: Maybe there is no need to deal with the cert rotation anymore :), just leave to our new solution `addon cert rotation` to have a single way to do that and this will also avoid any confusing. 

Also, if users choose to change the cert of `dex` or `gangway`, this will possibly overrite the certs with the wrong CA. 

The same suggestion to `gangway.`

internal/pkg/skuba/addons/gangway.go
jenting: The default value is `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`, so I think it's always right because we have default value in it.

But it's good to adjust the naming.
- for snake case variable: read from an environment variable
- for camel case variable: read from config

internal/pkg/skuba/addons/dex.go
jenting: Agree, I forgot custom CA case, I'll remove it, thx.

internal/pkg/skuba/addons/addons.go
```diff
@@ -86,14 +86,16 @@ type Addon struct {
...
 
 type AddonConfiguration struct {
-	ClusterVersion *version.Version
-	ControlPlane   string
-	ClusterName    string
+	ClusterVersion          *version.Version
+	ControlPlane            string
+	ClusterName             string
+	ControlPlaneHost        string
+	ControlPlaneHostAndPort string
```
cmurphy, line 98: Why do `ControlPlane`, `ControlPlaneHost` and `ControlPlaneHostAndPort` need to be split out into separate attributes? Couldn't we just infer one from the other?

internal/pkg/skuba/addons/dex.go
```diff
@@ -187,6 +192,12 @@ spec:
           - /usr/bin/caasp-dex
           - serve
           - /etc/dex/cfg/config.yaml
+        env:
+          - name: OIDC_CLIENT_SECRET
```
cmurphy, line 196: This is just one of two client secrets, the gangway client secret. Could it be renamed to something like `OIDC_GANGWAY_CLIENT_SECRET`? I think the hardcoded CLI secret should also be generated and taken from an environment variable: https://github.com/SUSE/skuba/blob/f2424908c819dda0a7656c02332f8f880dffc293/internal/pkg/skuba/addons/dex.go#L152

internal/pkg/skuba/addons/addons.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 SUSE LLC.
+ * Copyright (c) 2019,2020 SUSE LLC.
```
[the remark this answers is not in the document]
jenting: Shouldn't we keep the year history?

internal/pkg/skuba/addons/dex.go
jenting: The hardcoded CLI secret in for `skuba auth login` and it's spread out to all unauthorized users. Therefore, all the skuba CLI should use the same client secret otherwise it would fail when verifying the client secret.

The upstream had a new solution for the static public client https://github.com/dexidp/dex/pull/1701.
But this PR focus on dex&gangway integration, if we have other concerns for public client secret, I'd suggest having another PR to address it. But in my first thought, it would have a backward-compatible issue because different users use different skuba CLI versions.

internal/pkg/skuba/addons/addons.go
jenting: Good suggestion! I'll update it.

internal/pkg/skuba/addons/dex.go
jenting: Rename to `OIDC_GANGWAY_CLIENT_SECRET`.

internal/pkg/skuba/addons/addons.go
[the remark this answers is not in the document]
jenting: never mind, I give up bump the year.

jenting:
Thanks for the review and comments @Klaven @cmurphy :bow: 
I've rewritten the commits to make it easier to understand each commit purpose.

internal/pkg/skuba/addons/dex.go
cmurphy: [**@jenting**](https://github.com/jenting) commented on [Jun 22, 2020, 5:24 PM PDT](https://github.com/SUSE/skuba/pull/1174#discussion_r443894423 "2020-06-23T00:24:54Z - Replied by Github Reply Comments"):
> The hardcoded CLI secret in for `skuba auth login` and it's spread out to all unauthorized users. Therefore, all the skuba CLI should use the same client secret otherwise it would fail when verifying the client secret.

Thanks for this information, I see now that this is somewhat common practice, e.g. https://community.auth0.com/t/authenticating-an-installed-cli-with-oidc-and-a-third-party-client/7594

> 
> The upstream had a new solution for the static public client https://github.com/dexidp/dex/pull/1701.
> But this PR focus on dex&gangway integration, if we have other concerns for public client secret, I'd suggest having another PR to address it. But in my first thought, it would have a backward-compatible issue because different users use different skuba CLI versions.

:+1:

cmurphy:
[**@jenting**](https://github.com/jenting) commented on [Jun 22, 2020, 6:47 PM PDT](https://github.com/SUSE/skuba/pull/1174#issuecomment-647857331 "2020-06-23T01:47:39Z - Replied by Github Reply Comments"):
> Thanks for the review and comments @Klaven @cmurphy :bow: 
> I've rewritten the commits to make it easier to understand each commit purpose.

Thanks, much easier to follow now! Looks good to me but needs a rebase after https://github.com/SUSE/skuba/pull/1162

internal/pkg/skuba/addons/dex.go
jenting: > Thanks for this information, I see now that this is somewhat common practice, e.g. https://community.auth0.com/t/authenticating-an-installed-cli-with-oidc-and-a-third-party-client/7594
> 

yes, we took google cloud SDK design as the reference when we implemented `skuba auth login`, good finding :100: .

jenting:
rebase onto the master branch to resolve conflicts.

internal/pkg/skuba/oidc/cert.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"github.com/pkg/errors"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/kubernetes/cmd/kubeadm/app/constants"
+	"k8s.io/kubernetes/cmd/kubeadm/app/util/pkiutil"
+
+	"github.com/SUSE/skuba/internal/pkg/skuba/util"
+)
+
+// CreateServerCert creates a erver certificate with subject alternative name (SAN)
```
c3y1huang, line 29: nit
```suggestion
// CreateServerCert creates a server certificate with subject alternative name (SAN)
```

jenting: nice catch!

PR #1176, opened 2020-06-19
activate 5 and not 4.0 product

jordimassaguerpla:

## Why is this PR needed?

activate 5 and not 4.0

Fixes #

appear in the changelog.


kkaempf:
This has lots of `x86_64` in the paths. What about `aarch64` ?

jordimassaguerpla:
About aarch64, we can fix that later. Please create a separate issue for this. I would prefer to go ahead with this fix given is blocking the beta right now.

PR #1177, opened 2020-06-19
Fix the branched repo url

davidcassany:
## Why is this PR needed?

The URL for the branched repo is pointing to a wrong OBS repository,
for CaaSPv5 instead of SLE_15_SP1 we are using SLE_15_SP2.

## What does this PR do?

Update the URL to make use of `SLE_15_SP2`

### DIFF
--- a/ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
+++ b/ci/jenkins/pipelines/prs/skuba-test.Jenkinsfile
@@ -85,7 +85,7 @@ node('caasp-team-private-integration') {
            }
            if (pr_repo_label != null) {
                def branch_name = pr_repo_label.name.split(":")[1]
-               branch_repo = "http://download.suse.de/ibs/Devel:/CaaSP:/${repo_version}:/Branches:/${branch_name}/SLE_15_SP1"
+               branch_repo = "http://download.suse.de/ibs/Devel:/CaaSP:/${repo_version}:/Branches:/${branch_name}/SLE_15_SP2"
 
            }
 


davidcassany:
@pablochacin we could create a new constant variable for the repository too, as we do for the CaaSP version. However I do not have a strong opinion on that, whatever we do will be hardcoded.

@jordimassaguerpla probably we could even consider using `standard` instead of `SLE_15_SP2` so we do not have to update it for every CaaSP version, minor details anyway.

jordimassaguerpla:
> @pablochacin we could create a new constant variable for the repository too, as we do for the CaaSP version. However I do not have a strong opinion on that, whatever we do will be hardcoded.
> 
> @jordimassaguerpla probably we could even consider using `standard` instead of `SLE_15_SP2` so we do not have to update it for every CaaSP version, minor details anyway.

I think is a good idea. standard is also the name used in SUSE....CaaSP:5

davidcassany:
> I think is a good idea. standard is also the name used in SUSE....CaaSP:5

Great, so I'd say lets go now for `SLE_15_SP2` approach (consistent with what we have) and create a card to run the change, I don't know now in how many places we would have to run the change, so better to do it in an organized way.

pablochacin:
The problem I see is that this change would break v4. As this branches is not sync, this will not be inmediate, but if we backport this change to v4, it will break.

davidcassany:
Ok I just added a follow up card to handle the maintainability issue 

SUSE/avant-garde#1738

PR #1178, opened 2020-06-19
Support Patternless kubernetes and crio packages

evrardjp:
Without this patch, skuba is not able to install the new
kind of packages (patternless).

This is a problem, as it's necessary to install caasp5.

This also tackles the upgrade case, where we remove
the previous packages, and install the new kind of pkgs.

Partial-Fixes: https://github.com/SUSE/avant-garde/issues/1663

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/kubernetes.go
+++ b/internal/pkg/skuba/deployments/ssh/kubernetes.go
@@ -39,7 +39,9 @@ const (
 func init() {
 	stateMap["kubernetes.bootstrap.upload-secrets"] = kubernetesUploadSecrets(KubernetesUploadSecretsContinueOnError)
 	stateMap["kubernetes.join.upload-secrets"] = kubernetesUploadSecrets(KubernetesUploadSecretsFailOnError)
-	stateMap["kubernetes.install-node-pattern"] = kubernetesInstallNodePattern
+	stateMap["kubernetes.install-fresh-pkgs"] = kubernetesFreshInstallAllPkgs
+	stateMap["kubernetes.upgrade-stage-one"] = kubernetesUpgradeStageOne
+	stateMap["kubernetes.upgrade-stage-two"] = kubernetesUpgradeStageTwo
 	stateMap["kubernetes.restart-services"] = kubernetesRestartServices
 }
 
@@ -56,24 +58,102 @@ func kubernetesUploadSecrets(errorHandling KubernetesUploadSecretsErrorBehavior)
 	}
 }
 
-func kubernetesInstallNodePattern(t *Target, data interface{}) error {
+func kubernetesParseInterfaceVersions(data interface{}) (string, string, error) {
 	kubernetesBaseOSConfiguration, ok := data.(deployments.KubernetesBaseOSConfiguration)
 	if !ok {
-		return errors.New("couldn't access kubernetes base OS configuration")
+		return "", "", errors.New("couldn't access kubernetes base OS configuration")
 	}
 
 	v, err := version.ParseSemantic(kubernetesBaseOSConfiguration.CurrentVersion)
 	if err != nil {
-		return err
+		return "", "", err
 	}
 	currentVersion := kubernetes.MajorMinorVersion(v)
-	patternName := fmt.Sprintf("patterns-caasp-Node-%s", currentVersion)
+	updatedVersion := ""
 
 	if kubernetesBaseOSConfiguration.UpdatedVersion != "" {
-		updatedVersion := kubernetes.MajorMinorVersion(version.MustParseSemantic(kubernetesBaseOSConfiguration.UpdatedVersion))
-		patternName = fmt.Sprintf("patterns-caasp-Node-%s-%s", currentVersion, updatedVersion)
+		updatedVersion = kubernetes.MajorMinorVersion(version.MustParseSemantic(kubernetesBaseOSConfiguration.UpdatedVersion))
 	}
-	_, _, err = t.ssh("zypper", "--userdata", "skuba", "--non-interactive", "install", patternName)
+	return currentVersion, updatedVersion, nil
+}
+
+func kubernetesFreshInstallAllPkgs(t *Target, data interface{}) error {
+	current, _, err := kubernetesParseInterfaceVersions(data)
+	if err != nil {
+		return err
+	}
+	var pkgs []string
+
+	// Standard packages for a new cluster
+	pkgs = append(pkgs, "+caasp-release", "skuba-update", "supportutils-plugin-suse-caasp", "cri-tools")
+	// Version specific
+	pkgs = append(pkgs, fmt.Sprintf("+kubernetes-%s-kubeadm", current))
+	pkgs = append(pkgs, fmt.Sprintf("+kubernetes-%s-kubelet", current))
+	pkgs = append(pkgs, fmt.Sprintf("+kubernetes-%s-client", current))
+	pkgs = append(pkgs, fmt.Sprintf("+cri-o-%s*", current))
+
+	_, _, err = t.zypperInstall(pkgs...)
+	return err
+}
+
+func kubernetesUpgradeStageOne(t *Target, data interface{}) error {
+	// Stage1 upgrades only kubeadm.
+	// zypper install -- -kubernetes-old-kubeadm +kubernetes-new-kubeadm
+	currentV, nextV, err := kubernetesParseInterfaceVersions(data)
+	if err != nil {
+		return err
+	}
+	if nextV == "" {
+		return errors.New("Incorrect upgrade version")
+	}
+
+	var pkgs []string
+
+	if currentV == "1.17" {
+		// 1.17 is the last version included in CaaSP4. It's the tipping
+		// point where we changed our packaging.
+		// On 1.17 we can't remove kubernetes-1.17-kubeadm, because it doesn't exist.
+		// Removing kubeadm keeps kubelet alive.
+		// The rest needs to be removed on the next stage.
+		pkgs = append(pkgs, "-patterns-caasp-Node-1.17", "-kubernetes-kubeadm", "-cri-o-kubeadm-criconfig")
+	} else {
+		pkgs = append(pkgs, fmt.Sprintf("-kubernetes-%s-kubeadm", currentV))
+	}
+
+	pkgs = append(pkgs, fmt.Sprintf("+kubernetes-%s-kubeadm", nextV))
+	_, _, err = t.zypperInstall(pkgs...)
+	return err
+}
+
+func kubernetesUpgradeStageTwo(t *Target, data interface{}) error {
+	// Stage2 installs the rest of the packages during the upgrade
+	// with zypper install -- -<previous>-<component> +<next>-<component>
+	currentV, nextV, err := kubernetesParseInterfaceVersions(data)
+	if err != nil {
+		return err
+	}
+	if nextV == "" {
+		return errors.New("Incorrect upgrade version")
+	}
+
+	var pkgs []string
+
+	if currentV == "1.17" {
+		// on 1.17 we need to finalize the cleanup for the
+		// caasp4 to 5 migration
+		pkgs = append(pkgs, "-kubernetes-kubelet")
+		pkgs = append(pkgs, "-kubernetes-common")
+		pkgs = append(pkgs, "-kubernetes-client")
+		pkgs = append(pkgs, "-cri-o")
+	} else {
+		pkgs = append(pkgs, fmt.Sprintf("-kubernetes-%s-*", currentV))
+		pkgs = append(pkgs, fmt.Sprintf("-cri-o-%s*", currentV))
+	}
+
+	pkgs = append(pkgs, fmt.Sprintf("+kubernetes-%s-client", nextV))
+	pkgs = append(pkgs, fmt.Sprintf("+kubernetes-%s-kubelet", nextV))
+	pkgs = append(pkgs, fmt.Sprintf("+cri-o-%s*", nextV))
+	_, _, err = t.zypperInstall(pkgs...)
 	return err
 }
 
--- a/internal/pkg/skuba/deployments/ssh/zypper.go
+++ b/internal/pkg/skuba/deployments/ssh/zypper.go
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 SUSE LLC.
+ * Copyright (c) 2020 SUSE LLC.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/SUSE/skuba
 
+go 1.14
```
evrardjp, line 3: note self: to remove.

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -56,27 +58,87 @@ func kubernetesUploadSecrets(errorHandling KubernetesUploadSecretsErrorBehavior)
...
+	return currentVersion, updatedVersion, nil
+}
+
+func kubernetesFreshInstallAllPkgs(t *Target, data interface{}) error {
+	current, _, err := kubernetesParseInterfaceVersions(data)
+	if err != nil {
+		return err
+	}
+
+	pkgName := fmt.Sprintf("+kubernetes-%s-*", current)
```
evrardjp, line 86: note self: needs more surgical approach.

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -56,27 +58,87 @@ func kubernetesUploadSecrets(errorHandling KubernetesUploadSecretsErrorBehavior)
...
 	}
-	_, _, err = t.ssh("zypper", "--userdata", "skuba", "--non-interactive", "install", patternName)
+
+	var pkgs []string
+
+	if currentV == "1.17" {
+		// on 1.17 we can't remove kubernetes-1.17-kubeadm, doesn't exist.
+		// removing kubeadm remove many things, but keeps kubelet alive, to
+		// be removed on the next step
+		pkgs = append(pkgs, "-patterns-caasp-Node-1.17", "-kubernetes-kubeadm", "-cri-o-kubeadm-criconfig")
```
evrardjp, line 107: note self: untested.

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -56,27 +58,87 @@ func kubernetesUploadSecrets(errorHandling KubernetesUploadSecretsErrorBehavior)
...
+		return errors.New("Incorrect upgrade version")
+	}
+
+	var pkgs []string
+
+	if currentV == "1.17" {
+		// on 1.17 we need to finalise the cleanup
+		pkgs = append(pkgs, "-kubernetes-kubelet -kubernetes-common -kubernetes-client")
+	} else {
+		pkgs = append(pkgs, fmt.Sprintf("-kubernetes-%s-*", currentV))
```
evrardjp, line 135: note self: need surgical approach for this line and L138.

internal/pkg/skuba/deployments/ssh/zypper.go
```diff
@@ -15,12 +15,13 @@
...
-	err := t.Apply(kubernetesBaseOSConfiguration, "kubernetes.install-node-pattern")
-	if err != nil {
-		return false, err
-	}
-	return true, nil
+func (t *Target) zypperInstall(packages ...string) (stdout string, stderr string, error error) {
+	// Runs a zypper install command wrapped with the right userdata and parameters
+	var cliArgs []string
+	cliArgs = append(cliArgs, "--userdata", "skuba")
+	cliArgs = append(cliArgs, "--non-interactive", "install", "--")
```
evrardjp, line 24: note self, that looks silly, maybe put it in one line.

internal/pkg/skuba/deployments/ssh/kubernetes.go
evrardjp: note self: 135 doesn't need surgical approach, while L138 does. Fixed.

evrardjp:
Rebased to include latest master changes, including the necessary 1.13 support.

evrardjp:
Retriggered CI now that the build service is configured with the correct repos.

evrardjp:
recheck

evrardjp:
After a discussion at our standup today, we discussed about having extra feedback on patternless.

I discussed with Dirk today, and had a long chat/etherpad about alternatives.

I am now testing a PoC which doesn't imply Conflicts, only Requires, and Obsoletes, with different names. If that PoC is fine, I will update the packages _and this PR_, as it might mean some of the removals in here might not be needed anymore.

evrardjp:
Sadly, we went down the rabbithole with a weird behaviour in libsolv for the Obsoletes.
We're gonna drop that idea, and keep this patch idea the same.
The Patternless and versionappended therefore seems still the way to go.

However, as explained in the etherpad, we found a bug in the cri-o handling. This means that CRI-O also need to be versionappended.

This also means that "patternless" is harder, and might not make sense anymore!

However, the spirit behind patternless to simplify and be "versionappended" seems now more than ever necessary. I just fail to name/brand it properly :)
I am discussing with dirk if we should bring patterns back with versionappended cri-o and kubernetes, or if we keep patternless versionappended crio and k8s.

In the meantime, I have updated the PR to do patternless versionappended.

internal/pkg/skuba/deployments/ssh/zypper.go
```diff
@@ -15,12 +15,13 @@
...
-package deployments
+package ssh
 
-func (t *Target) InstallNodePattern(kubernetesBaseOSConfiguration KubernetesBaseOSConfiguration) (bool, error) {
-	err := t.Apply(kubernetesBaseOSConfiguration, "kubernetes.install-node-pattern")
-	if err != nil {
-		return false, err
-	}
-	return true, nil
+func (t *Target) zypperInstall(packages ...string) (stdout string, stderr string, error error) {
```
evrardjp, line 20: note to self, might want to expose this function.

evrardjp:
skuba-update requires any version of kubernetes client, therefore the packaging of kubernetes-<version>-client should also provide kubernetes-client. I have updated the package. I will retrigger the build when the package is published.

evrardjp:
Seems like the greenfield empty kubernetes deployed just fine, but there was a problem with the addons. I will check what's going on on Monday

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -39,7 +39,9 @@ const (
 func init() {
 	stateMap["kubernetes.bootstrap.upload-secrets"] = kubernetesUploadSecrets(KubernetesUploadSecretsContinueOnError)
 	stateMap["kubernetes.join.upload-secrets"] = kubernetesUploadSecrets(KubernetesUploadSecretsFailOnError)
-	stateMap["kubernetes.install-node-pattern"] = kubernetesInstallNodePattern
+	stateMap["kubernetes.install-fresh-pkgs"] = kubernetesFreshInstallAllPkgs
+	stateMap["kubernetes.upgrade-kubeadm-pkg"] = kubernetesUpgradeKubeadmPkg
+	stateMap["kubernetes.upgrade-other-kubernetes-pkgs"] = kubernetesUpgradeOtherPkgs
```
flavio, line 44: Maybe we can come up with a better name for `upgrade-other-kubernetes-pkgs`. Maybe we can have a `upgrade-k8s-stage-1` (which does `kubeadm`) and a `upgrade-k8s-stage-2`

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -56,24 +58,99 @@ func kubernetesUploadSecrets(errorHandling KubernetesUploadSecretsErrorBehavior)
...
+	}
+	var pkgs []string
+
+	// Standard packages for a new cluster
+	pkgs = append(pkgs, "+caasp-release", "skuba-update", "supportutils-plugin-suse-caasp", "cri-tools")
+	// Version specific
+	pkgs = append(pkgs, fmt.Sprintf("+kubernetes-%s-kubeadm", current))
+	pkgs = append(pkgs, fmt.Sprintf("+kubernetes-%s-kubelet", current))
+	pkgs = append(pkgs, fmt.Sprintf("+kubernetes-%s-client", current))
+	pkgs = append(pkgs, fmt.Sprintf("+cri-o-%s*", current))
```
flavio, line 93: Maybe we could put these patterns inside of a constant list to reuse them later inside of the code (eg: `kubernetesUpgradeOtherPkgs`)?

Maybe two lists: `phase1pkgs` and `phase2kpgs` :thinking:

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -56,24 +58,99 @@ func kubernetesUploadSecrets(errorHandling KubernetesUploadSecretsErrorBehavior)
...
+	}
+	if nextV == "" {
+		return errors.New("Incorrect upgrade version")
 	}
-	_, _, err = t.ssh("zypper", "--userdata", "skuba", "--non-interactive", "install", patternName)
+
+	var pkgs []string
+
+	if currentV == "1.17" {
+		// on 1.17 we can't remove kubernetes-1.17-kubeadm, doesn't exist.
```
flavio, line 112: Maybe expand the comment and state that k8s v1.17 is the last one installed using the pattern model. Just to provide more context to this "magical" number

internal/pkg/skuba/deployments/ssh/kubernetes.go
evrardjp: I hesitated there, because I have the impression they will be fluctuating over time. But indeed, they aren't right now, so that might make sense.

internal/pkg/skuba/deployments/ssh/kubernetes.go
evrardjp: Naming is hard :)

Anything goes. Is stage1/stage2 more explicit than kubeadm/the rest? The rest is definitely _not_ clear, but stages are also hard... I am fine with changing to stage1/stage2 though.

internal/pkg/skuba/deployments/ssh/kubernetes.go
evrardjp: Yes! Totally.

internal/pkg/skuba/deployments/ssh/kubernetes.go
evrardjp: I have updated to stage 1 and stage 2. Not sure if it's more explicit, but we can refactor later if necessary, if we find a better name :laughing:

internal/pkg/skuba/deployments/ssh/kubernetes.go
evrardjp: Let's keep that for now, if you don't mind.

internal/pkg/skuba/deployments/ssh/kubernetes.go
evrardjp: Done.

evrardjp:
I have resolved @flavio 's comments/concerns, and rebased this to test how this fares with Cilium 1.7 changes.

evrardjp:
I think we should go ahead and merge this PR and the packages into Devel:CaaSP:5, remove the previous kubernetes/cri-o packages in that repo.

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -56,24 +58,102 @@ func kubernetesUploadSecrets(errorHandling KubernetesUploadSecretsErrorBehavior)
 	}
 }
 
-func kubernetesInstallNodePattern(t *Target, data interface{}) error {
```
kkaempf, line 59: I'm missing a comment here, explaining the new return parameters

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -56,24 +58,102 @@ func kubernetesUploadSecrets(errorHandling KubernetesUploadSecretsErrorBehavior)
...
+		return err
+	}
+	if nextV == "" {
+		return errors.New("Incorrect upgrade version")
 	}
-	_, _, err = t.ssh("zypper", "--userdata", "skuba", "--non-interactive", "install", patternName)
+
+	var pkgs []string
+
+	if currentV == "1.17" {
```
kkaempf, line 112: This is brittle and shouldn't be buried in the code. There should be a LAST_CAASP_4_K8S_VERSION (or so) constant.

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -56,24 +58,102 @@ func kubernetesUploadSecrets(errorHandling KubernetesUploadSecretsErrorBehavior)
...
+		// 1.17 is the last version included in CaaSP4. It's the tipping
+		// point where we changed our packaging.
+		// On 1.17 we can't remove kubernetes-1.17-kubeadm, because it doesn't exist.
+		// Removing kubeadm keeps kubelet alive.
+		// The rest needs to be removed on the next stage.
+		pkgs = append(pkgs, "-patterns-caasp-Node-1.17", "-kubernetes-kubeadm", "-cri-o-kubeadm-criconfig")
```
kkaempf, line 118: Frankly, this makes me cringe. Removal of v4 patterns/packages should be handled by `Obsoletes` imho.

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -56,24 +58,102 @@ func kubernetesUploadSecrets(errorHandling KubernetesUploadSecretsErrorBehavior)
...
+	if err != nil {
+		return err
+	}
+	if nextV == "" {
+		return errors.New("Incorrect upgrade version")
+	}
+
+	var pkgs []string
+
+	if currentV == "1.17" {
```
kkaempf, line 141: see comment above

internal/pkg/skuba/deployments/ssh/kubernetes.go
evrardjp: While this is a noble thing we tried, and _should_ be working, we arrived to a libsolv issue/feature that was created for bad packagers, where an obsolete-ing package is considered as an upgrade candidate for the obsolete-d package (the tl;dr:). We can discuss this for later, but right now it works, and is user friendly enough IMO

internal/pkg/skuba/deployments/ssh/kubernetes.go
evrardjp: I can have that 1.17 as a constant, no problem. I suggest to do that as a follow up PR, to not delay this further.

internal/pkg/skuba/deployments/ssh/kubernetes.go
evrardjp: Agreed. Will do so in a follow up PR.

kkaempf:
Please create a github issue as a reminder for the follow up PR and reference this issue here. Then I'm fine to merge.

evrardjp:
I will just create another PR on top right after my meetings this morning.

PR #1179, opened 2020-06-22
Update release constants to v5

jordimassaguerpla:
package in SCC is pointing to v4 images, instead of to v5. This fixes it.

### DIFF
--- a/pkg/skuba/release_constants.go
+++ b/pkg/skuba/release_constants.go
@@ -21,5 +21,5 @@ package skuba
 
 const (
 	BuildType       = "release"
-	ImageRepository = "registry.suse.com/caasp/5"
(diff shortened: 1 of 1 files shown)


pkg/skuba/release_constants.go
```diff
@@ -21,5 +21,5 @@ package skuba
 
 const (
 	BuildType       = "release"
-	ImageRepository = "registry.suse.com/caasp/v4"
+	ImageRepository = "registry.suse.com/caasp/5"
```
davidcassany, line 24: You are missing the `v`

thehejik: just found it out

jordimassaguerpla:
v5 now

PR #1192, opened 2020-06-24
Enable ExtendedResourceToleration by default

cmurphy:
Enable the ExtendedResourceToleration admission controller[1] by default
for all deployments.

This admission controller is beneficial to heterogeneous clusters that
may want to make use of kubernetes node taints[2] for giving
preferential scheduling to workloads that require special hardware, such
as GPUs. This admission controller is a mutating admission controller
that automatically adds a resource toleration to pods that request an
extended resource that is being advertised by a device plugin[3]. It has
no effect on pods that don't request such resources, and the added
toleration would have no effect if the cluster administrator is not
using taints for this purpose, so it is safe to enable by default and a
convenience for cost-aware admins with these hardware requirements.

[1] https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#extendedresourcetoleration
[2] https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#example-use-cases
[3] https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/

Fixes https://github.com/SUSE/avant-garde/issues/1699

### DIFF
--- a/internal/pkg/skuba/kubeadm/configmap.go
+++ b/internal/pkg/skuba/kubeadm/configmap.go
@@ -186,6 +186,7 @@ func setApiserverAdmissionPlugins(initCfg *kubeadmapi.InitConfiguration, cluster
 	admissionPlugins = append(admissionPlugins, "NodeRestriction")
 	// List of skuba-enabled plugins
 	admissionPlugins = append(admissionPlugins, "PodSecurityPolicy")
+	admissionPlugins = append(admissionPlugins, "ExtendedResourceToleration")
 	admissionPlugins = skubautil.UniqueStringSlice(admissionPlugins)
 	initCfg.APIServer.ControlPlaneComponent.ExtraArgs["enable-admission-plugins"] = strings.Join(admissionPlugins, ",")
 }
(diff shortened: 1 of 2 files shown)


cmurphy:
> Personally I would like to see an RFC for this and greater team awareness

I've proposed an RFC here: https://github.com/SUSE/caasp-rfc/pull/62
I'm all for team awareness and getting broader input. Maybe for future design changes we could gather requirements like writing RFCs on the avant-garde issue before we accept it into the sprint.

> Also, we need to make PR's to documentation to remove the steps to add admission plugin but also to let people know that in caaspv5 we will now add tollorations to all resources requested in the manifests files (not just GPU's).

Sure, simplifying the documentation is a large part of the motivation for doing this so I think revising that part of the guide goes without saying :) It's not quite correct to say tolerations are added for all resources requested, though, it's only extended resources that are recognized by a device plugin, and our only example of using a device plugin is the NVIDIA GPU device plugin. Can we keep the mention of this admission controller within the GPU section, or are you wanting to add it somewhere else?

Klaven:
@cmurphy I agree, we should get better about making issues clearer. I'll take responsibility for the fact this was not in the issue and take steps to ensure this type of information is in the next issue.

Klaven:
> Can we keep the mention of this admission controller within the GPU section, or are you wanting to add it somewhere else?
:thinking: honestly I don't know. Part of me things we should post it somewhere else because it will work for other things then just GPU... but really this is probably the main case. it's probably fine here for now.

>  It's not quite correct to say tolerations are added for all resources requested, though, it's only extended resources that are recognized by a device plugin
:100: yes, you are correct.

jordimassaguerpla:
This is not in the release-caasp-5.0.0 branch, so not present in GMC.
Setting label to v5 and not 5.0.0.

PR #1195, opened 2020-06-25
Add image mirror e2e test

pablochacin:
## Why is this PR needed?

Test new images that do not imply changes in skuba code and therefore, have no associated PR. 

## What does this PR do?

If branch repo is specified, set the URL for a mirror repository and set the environment variables to it when deploying the cluster

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/jenkins/pipelines/skuba-e2e-test.Jenkinsfile
+++ b/ci/jenkins/pipelines/skuba-e2e-test.Jenkinsfile
@@ -5,8 +5,12 @@
 // type of worker required by the job 
 def worker_type = 'integration'
 
-// repository branch
-branch_repo = ""
+// Repo and registry branch
+def branch_repo = ""
+def branch_registry = ""
+
+// original (non-branched) registry, only needed when branch_registry is in use
+def original_registry = ""
 
 // worker selection labels 
 def labels = 'e2e'
@@ -31,7 +35,9 @@ node('caasp-team-private-integration') {
 
         if (env.REPO_BRANCH != ""){
                branch_repo = "http://download.suse.de/ibs/Devel:/CaaSP:/5:/Branches:/${env.REPO_BRANCH}/SLE_15_SP2"
-        }
+               branch_registry = "registry.suse.de/devel/caasp/5/branches/${env.REPO_BRANCH}/containers"
+               original_registry = "registry.suse.de/devel/caasp/5/containers/cr/containers"
+         }
     }
 }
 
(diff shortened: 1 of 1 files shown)


ci/jenkins/pipelines/skuba-e2e-test.Jenkinsfile
```diff
@@ -31,7 +35,9 @@ node('caasp-team-private-integration') {
 
         if (env.REPO_BRANCH != ""){
                branch_repo = "http://download.suse.de/ibs/Devel:/CaaSP:/5:/Branches:/${env.REPO_BRANCH}/SLE_15_SP2"
-        }
+               branch_registry = "registry.suse.de/devel/caasp/5/branches/${env.REPO_BRANCH}/containers/cr/containers/caasp/v5"
```
davidcassany, line 38: this is wrong, see the cilium example to check the registry URI. It should be 

`registry.suse.de/devel/caasp/5/branches/cilium17/containers/caasp/v5`

The schema of this url is as follows:

`<host>/<project_path>/<project_repo>/<image_name>:<image_tag>`

I believe that the following should also work and this way we do not include any specific image value inside the configuration (even I doubt we ever will customize the  `caasp/v5` prefix in some of our images but not all):

 `branch_registry = "registry.suse.de/devel/caasp/5/branches/cilium17/containers"`
and
`original_registry ="registry.suse.de/devel/caasp/5/containers/cr/containers"`

Klaven, line 38: it seems to me that you would need to check the `env.BRANCH_REGISTRY` or something like that instead of the `env.REPO_BRANCH` for all of these? or is that just me misunderstanding how this Jenkinsfile works?

davidcassany: This value points to the IBS project by a convention we decided. So if you build a `Devel:CaaSP:5:Branches:cilium17` project `REPO_BRANCH` should be set to `cilium17`.

Then while you are right in the sense that this is not necessarily the same project for images and packages, I don't see any specific need to have them separate in the branch project. So lets assume that the images and packages for that branch are all placed in the same IBS project. This, as a user, allows you just to set a single value to make sure the CI is pulling packages and images from that specific IBS project.

PR #1198, opened 2020-06-25
Fix check for null environment variable

pablochacin:
# Why is this PR needed?

After changes introduced in  https://github.com/SUSE/skuba/pull/1180 the daily e2e tests fail because some environment variables are not defined. 

## What does this PR do?

Fix the check for environment variables not set. 

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


cmurphy:
Looks right according to https://stackoverflow.com/a/45760094 - the jenkins failures seem unrelated?

PR #1168, opened 2020-06-16
Remove scm stage from pr pipeline

pablochacin:
## Why is this PR needed?

Presently the PR code is rebased on top of the target branch as part of the pipeline logic. This has several issues: 
* The PR code is checked (linted) before rebasing to the target branch (e.g. master)
* The pipeline code that is executed is the one from the PR, not the more recent from the target branch, if any.
* The logic for checking-out PR and merging with target branch is split between the Jenkins pipeline and python code in `ci/tasks/rebase.py` executed from the `ci/Makefile`. This makes it difficult to maintain this logic avoiding conflicts and inconsistencies.

Fixes https://github.com/SUSE/avant-garde/issues/1698

## What does this PR do?

Removes the logic for rebasing the code from the PR pipeline and uses the marge option in the job's scm configuration. This way, the PR is merged before the pipeline is executed.

Due to a limitation in the multibranch job plugin, this change impacts the location of the code, as it cannot be checked out into a subdirectory as the current logic does, but directly into the workspace. Therefore, the daily jobs must also be updated to checkout the code directly in the workspace and the code location must be adjusted in all scripts and pipelines.

Also, because the code is merged before the pipeline is executed, the reference to the head of the PR branch is lost and the logic for reporting the status to github must be adapted.

Finally, the unit test make target had to be fixed to support the checkout into a directory different from `skuba`.

## Anything else a reviewer needs to know?

The PR has been tested with the new SCM configuration in [this modified pr pipeline](https://ci.suse.de/job/caasp-jobs/job/dev/job/pablo/job/pr-test-pablo/view/change-requests/job/PR-1168/)

Also, a [daily e2e job has been tested against this PR](https://ci.suse.de/job/caasp-jobs/job/e2e/job/v5/job/vmware/job/test_nginx_deployment-daily/30/) with the modified SCM configuration (checkout into the workspace, not into the skuba subdirectory)

# Merge restrictions

(Please do not edit this)

We are in *v4-maintenance phase*, so we will restrict what can be merged to prevent unexpected surprises:

    What can be merged (merge criteria):
        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine
        there is a PR for updating documentation (or a statement that this is not needed)

PR #1200, opened 2020-06-26
Use the Proper SLE15 SP2 JeOS GM image for testing on OpenStack

dirkmueller:
## Why is this PR needed?

The business case this fixes is that the CI is using the final SP2 image that we ship to customers rather than something older and unsupported. 

As such it brings more relevance to what we test and how we test and that is considered a good thing. 

## What does this PR do?

It fixes the Image configuration to point to the right image. 

There is no manual testing needed. 

## Info for QA

There is no QA needed. if CI passes its good. 

### Related info

This PR can be merged independently.

The wrong image is used. 

The SLE15 SP2 GM JeOS image for OpenStack is being used for OpenStack testing.

## Docs

The docs need to refer to the same image. 

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/openstack/terraform.tfvars.example
+++ b/ci/infra/openstack/terraform.tfvars.example
@@ -1,7 +1,7 @@
 # Name of the image to use
 # EXAMPLE:
-# image_name = "SLE-15-SP1-JeOS-GMC"
-image_name = ""
+# image_name = "
+image_name = "SLES15-SP2-JeOS.x86_64-15.2-OpenStack-Cloud-GM"
 
 # Name of the internal network to be created
 # EXAMPLE:
--- a/ci/infra/openstack/terraform.tfvars.json.ci.example
+++ b/ci/infra/openstack/terraform.tfvars.json.ci.example
@@ -1,5 +1,5 @@
 {
-    "image_name": "SLES15-SP2-JeOS.x86_64-PublicRC2",
+    "image_name": "SLES15-SP2-JeOS.x86_64-15.2-OpenStack-Cloud-GM",
     "internal_net": "testing",
     "internal_subnet": "testing-subnet",
     "internal_router": "testing-router",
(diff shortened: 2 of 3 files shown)


evrardjp:
I am already using those images (but not the reboot fix), and that works for me. :+1:

evrardjp:
I think this should be placed on Z squad's radar. What do you think @jordimassaguerpla ?

pablochacin:
There seem to be two unrelated changes in the PR: the image and the reboot logic. If you drop the second commit from this PR we can take the update of the image and test it in the CI.

The reboot logic, we prefer not to changed in the SAME PR unless it is broken, which is not the case as far as we know and JP reports.

dirkmueller:
@pablochacin I removed the 2nd commit. I already tested it, and so did JP. I think a review would be enough.

evrardjp:
Interesting that the pr-head chokes on a id_shared file... Should we edit the authorized_keys var to add the appropriate ci pubkey into ci/infra/openstack/terraform.tfvars.json.ci.example ? I suppose there is something else in our CI tooling dealing with this...

pablochacin:
> Interesting that the pr-head chokes on a id_shared file... Should we edit the authorized_keys var to add the appropriate ci pubkey into ci/infra/openstack/terraform.tfvars.json.ci.example ? I suppose there is something else in our CI tooling dealing with this...

I observed this same error in PR #1205 The problem seems to be related to the changes in PR #1168 (the location of files change) I'm looking at it.

dirkmueller:
potentially this required a rebase. I've rebased on top of master.

pablochacin:
I run an e2e test job with the PR and the cluster was deployed correctly. The deployment failed with an unrelated issue related to the image repository. So from my perspective, the new image solves the ssh issue.  

I'm still investigating why the PR tests are failing. It also seems to be unrelated, as I mentioned in a previous comment.

pablochacin:
> I'm still investigating why the PR tests are failing. It also seems to be unrelated, as I mentioned in a previous comment.

Effectively, it is due to a problem with PR #1168

kkaempf:
https://github.com/SUSE/avant-garde/issues/1739#issuecomment-651289097 - let's get this merged.

PR #1202, opened 2020-06-26
Use https://download.suse.de

dirkmueller:
download.suse.de is a download redirector that redirects to the most
local installation of the repositories, which is better than hardcoding
the provo site, for cases when VMWare is not deployed in Provo but lets
say, as an example, in Nuernberg.

Also, try to use https:// as it is 2020 after all.

## Why is this PR needed?

The business case is reducing inter-continental traffic between provo and nbg.

## What does this PR do?

See commit message.

No. 

## Info for QA

This does not need to be tested for QA

### Related info

Here's some information on why HTTPS should be used by default: 
https://web.dev/why-https-matters

The wrong download sites are used.

The right download sites are used.

## Docs

No documentation updates needed.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/libvirt/terraform.tfvars.json.ci.example
+++ b/ci/infra/libvirt/terraform.tfvars.json.ci.example
@@ -11,25 +11,25 @@
     "workers": 2,
     "username": "sles",
     "repositories": {
-        "caasp_devel": "http://download.suse.de/ibs/Devel:/CaaSP:/4.5/SLE_15_SP2/",
-        "suse_ca": "http://download.suse.de/ibs/SUSE:/CA/SLE_15_SP2",
-        "sle_server_pool": "http://download.suse.de/ibs/SUSE/Products/SLE-Product-SLES/15-SP2/x86_64/product",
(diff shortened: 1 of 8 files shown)


skuba-update/test/os/suse/install_skuba_update.sh
```diff
@@ -19,15 +19,15 @@ set -xe
...
+    zypper ar --no-gpgcheck https://download.opensuse.org/distribution/leap/15.2/repo/oss/ leap15.2
+    zypper ar --no-gpgcheck https://download.opensuse.org/update/leap/15.2/oss updates
     zypper ref
 fi
 
 zypper in -y python3-setuptools lsof
 
 if [ "$release" = "NAME=\"SLES\"" ]; then
-    zypper rr leap15.1
+    zypper rr leap15.2
```
Klaven, line 30: one quick little nit. this really seems like a unrelated change. in my opinion this would go in another PR, or at least another commit to help with debugging later.

dirkmueller: I removed this part of the change.

Itxaka:
conflict and not touched in several months, closing it. Feel free to reopen if needed.

PR #1205, opened 2020-06-29
Fix the reboot logic

dirkmueller:
There is no need to reboot the node if the kernel didn't change. so
avoid unnecessary reboots and ensure that we have the reboot finished.

## Why is this PR needed?

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/libvirt/master-instance.tf
+++ b/ci/infra/libvirt/master-instance.tf
@@ -134,11 +134,17 @@ resource "null_resource" "master_reboot" {
     }
 
     command = <<EOT
-ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null $user@$host sudo reboot || :
-# wait for ssh ready after reboot
-until nc -zv $host 22; do sleep 5; done
-ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -oConnectionAttempts=60 $user@$host /usr/bin/true
+export sshopts="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -oConnectionAttempts=60"
+if ! ssh $sshopts $user@$host 'sudo needs-restarting -r'; then
+    ssh $sshopts $user@$host sudo reboot || :
+    export delay=5
+    # wait for node reboot completed
+    while ! ssh $sshopts $user@$host 'sudo needs-restarting -r'; do
+        sleep $delay
+        delay=$((delay+1))
+        [ $delay -gt 30 ] && exit 1
+    done
+fi
 EOT
-
   }
 }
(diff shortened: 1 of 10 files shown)


cmurphy:
Is a kernel update really the only time that the node should be rebooted? There may be some cases where other processes (systemd? dbus) ought to be restarted, and a node reboot is the only catch-all to make sure that happens.

dirkmueller:
As far as I can see, the kernel-default-base to kernel+default switch was the sole reason for reboot. I didn't find any 'lets install all available updates' code anywhere, although I agree from a testing perspective that would be useful to have.

I will remove the reboot omission optimisation again.

ci/infra/libvirt/master-instance.tf
```diff
@@ -134,11 +134,17 @@ resource "null_resource" "master_reboot" {
     }
 
     command = <<EOT
-ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null $user@$host sudo reboot || :
-# wait for ssh ready after reboot
-until nc -zv $host 22; do sleep 5; done
-ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -oConnectionAttempts=60 $user@$host /usr/bin/true
+export sshopts="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -oConnectionAttempts=60"
+if ! ssh $sshopts $user@$host 'test -d /lib/modules/$(uname -r)'; then
```
Itxaka, line 138: shouldn't we be checking `/var/run/reboot-needed` instead? or we dont have that installed on our images by default somehow (its part of core zypper I think?)?

dirkmueller: well, the idea was that we identify that we need a reboot (kernel modules installed do not match running kernel) and check that reboot happened (because running kernel matches installed modules). 

if we always want to reboot we're back to the original problem that it is hard to find out when reboot happened. 

I agree that touching /var/run/reboot-needed and waiting for the file to disappear could be a solution for that. I'll try that.

dirkmueller: actually this does not solve it, because I can't distinguish between unreachable node (which is an error) and 'node reachable but file no longer there (so reboot happened - success). 

I'm still thinking my current patch is the best one :/

dirkmueller: Found a way to use needs-rebooting (which is a zypper plugin).

dirkmueller:
@Klaven all of those terraform scripts are terrible pile of copy&pastes ..

PR #1218, opened 2020-07-01
Fix stage2 upgrade

evrardjp:
Without this patch, it's hard for the resolver to know which
version of kubernetes-client to remove, and therefore it will
be breaking skuba-updates requiring kubernetes-client.

By being explicit into what to remove, we keep kubernetes-1.18-client,
and it just works afterwards.

Fixes: https://github.com/SUSE/avant-garde/issues/1664

Nothing

## Info for QA

Needed for upgrades

### Related info

* upstream issues

Can't upgrade cluster to 1.18

Goes further in the upgrade process.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


evrardjp:
Common doesn't exist anymore. Should I write it somewhere?

PR #1227, opened 2020-07-06
Allow any version of k8s-client for skuba-update and allow cri-o bump during stage1

evrardjp:
With the change of the virtual provides from kubernetes- to
k8s- and the version appended packages, we can now to decide
whether this package need _any_ version of a kubernetes, or
a specific one. skuba-update can rely on any version of kubectl,
therefore we should be updating the spec to rely on k8s-client
being present.

## Why is this PR needed?

Without this whole virtual provides, the migration will auto bump k8s to 1.18, which we want to avoid.

## What does this PR do?

Ensure we are not using kubernetes-component* packages.

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/packaging/suse/skuba_spec_template
+++ b/ci/packaging/suse/skuba_spec_template
@@ -60,7 +60,6 @@ Summary:        Utility to automatically refresh and update a skuba cluster
 Group:          System/Management
 Requires:       python3-setuptools
 Requires:       zypper >= 1.14.15
-Requires:       (k8s-client or kubernetes-client)
 Requires:       lsof
 BuildArch:      noarch
 %{?systemd_requires}
--- a/ci/packaging/suse/skuba_spec_template
+++ b/ci/packaging/suse/skuba_spec_template
@@ -60,7 +60,7 @@ Summary:        Utility to automatically refresh and update a skuba cluster
 Group:          System/Management
 Requires:       python3-setuptools
 Requires:       zypper >= 1.14.15
-Requires:       kubernetes-client
+Requires:       (k8s-client or kubernetes-client)
 Requires:       lsof
 BuildArch:      noarch
 %{?systemd_requires}
--- a/internal/pkg/skuba/deployments/ssh/kubernetes.go
+++ b/internal/pkg/skuba/deployments/ssh/kubernetes.go
@@ -113,9 +113,16 @@ func kubernetesUpgradeStageOne(t *Target, data interface{}) error {
 		// 1.17 is the last version included in CaaSP4. It's the tipping
 		// point where we changed our packaging.
 		// On 1.17 we can't remove kubernetes-1.17-kubeadm, because it doesn't exist.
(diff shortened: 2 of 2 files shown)


kkaempf:
lgtm !

evrardjp:
Retriggering, as kubernetes was not properly built in the repo when this PR was up.

ci/packaging/suse/skuba_spec_template
```diff
@@ -60,7 +60,7 @@ Summary:        Utility to automatically refresh and update a skuba cluster
 Group:          System/Management
 Requires:       python3-setuptools
 Requires:       zypper >= 1.14.15
-Requires:       kubernetes-client
```
evrardjp, line 63: To prevent the upgrade of the client without reason, we should _not_ do this.

Instead we should remove the requirement, and be explicit about possible breakages.

evrardjp: If the removal breaks the greenfields's skuba-update (which I doubt we have test coverage for), then we can probably iterate with `Requires: (kubernetes-client or k8s-client)`

evrardjp:
Please hold. Don't merge this yet.

ci/packaging/suse/skuba_spec_template
evrardjp: The Requires of (kubernetes-client or k8s-client) triggers the install of kubernetes-1.18-client during zypper dup, so I removed this require.

internal/pkg/skuba/deployments/ssh/kubernetes.go
```diff
@@ -113,9 +113,16 @@ func kubernetesUpgradeStageOne(t *Target, data interface{}) error {
...
+		// We are removing kubeadm while keeping kubelet alive to its version 1.17.
+		// For the initial migration we need to update crio kubeadm
+		// to 1.18 in stage1, due to conflict resolution: the caasp4
+		// cri-o-kubeadm-criconfig requires kubernetes-kubeadm which is
+		// not provided anymore (when we remove kubernetes-kubeadm, and
+		// because we don't want to have the same provides: on the new
+		// package to avoid upgrade during zypper migration).
+		// we need to remove cri-o in stage2 else 1.17 kubelet could
+		// complain about cri-runtime being absent.
+		pkgs = append(pkgs, "-patterns-caasp-Node-1.17", "-\"kubernetes-kubeadm<1.18\"", "-caasp-config", "-cri-o-kubeadm-criconfig")
```
davidcassany, line 125: oh I didn't know zypper was capable to understand a syntax such as `-kubernetes-kubeadm<1.18` :wink:

evrardjp: Magic! https://media.giphy.com/media/NmerZ36iBkmKk/giphy.gif

PR #1236, opened 2020-07-08
CI: disable routable check for vmware

Itxaka:
There seems to be a race condition that can cause terraform to
timeout while checking if an IP is routable. Disable this check for
the moment.

Related to: https://github.com/SUSE/avant-garde/issues/1682

## Why is this PR needed?

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/vmware/lb-instance.tf
+++ b/ci/infra/vmware/lb-instance.tf
@@ -140,6 +140,7 @@ resource "vsphere_virtual_machine" "lb" {
   datastore_id         = (var.vsphere_datastore == null ? null : data.vsphere_datastore.datastore[0].id)
   datastore_cluster_id = (var.vsphere_datastore_cluster == null ? null : data.vsphere_datastore_cluster.datastore[0].id)
   folder               = var.cpi_enable == true ? vsphere_folder.folder[0].path : null
+  wait_for_guest_net_routable = var.wait_for_guest_net_routable
 
   clone {
     template_uuid = data.vsphere_virtual_machine.template.id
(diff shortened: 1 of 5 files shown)


pablochacin:
I think that this change is dangerous in the context of machines with multiple network interfaces, as some may not be routable. See [comment in this terraform issue](https://github.com/hashicorp/terraform-provider-vsphere/issues/446#issuecomment-377792483)

Itxaka:
I do agree on this @pablochacin but this should be only a temp fix while the issue is fixed upstream, even if the issue is this!

As for the moment we dont expect to have multiple network interfaces on machine provision(which is the only place this should hit) I felt that it was ok to tempfix this.

I will report this upstream tomorrow and try to find if the issue its actually that so we can start moving the real fix along.

pablochacin:
> I do agree on this @pablochacin but this should be only a temp fix while the issue is fixed upstream, even if the issue is this!
The problem is that we deliver these templates to customers. So we should find a way to override this setting only for the CI.

pablochacin:
My suggestion is to make this behavior dependent on a variable we can override in the CI but by default, will behaveas today.

Itxaka:
@pablochacin @jordimassaguerpla @davidcassany updated as mentioned on the meeting.

Tested introducing the change on existing deployed infra and no changes were triggered, so it should be 100% safe to introduce as it has no effect on customers at all unless they want to change this value.

`Plan: 0 to add, 0 to change, 0 to destroy.`

PR #1226, opened 2020-07-06
New command: skuba addon refresh localconfig (bsc#1173055)

jenting:
## Why is this PR needed?

The addons base manifest will be outdated because skuba did not provide a way to update the base addons manifest.
As time goes by, the addon manifest data structure/key might change, it causes the user's kustomize patch is not valid anymore.

We need to provide a command to upgrade existed base addons manifest, also we'll help to migrate legacy addon manifest folder definition. Note that, we'll not keep the content of legacy addon manifest if the admin had changed the default value in it.

. bsc#1173055 & bsc#1172805
. xrefs https://github.com/SUSE/avant-garde/issues/1765

## What does this PR do?

Add `skuba addon refresh localconfig`: it updates the current on-disk local cluster definition folder to the current Kubernetes version, also it supports migrate old addon definition folders (before 4.2.0) to the new ones.

## Anything else a reviewer needs to know?

N/A

## Info for QA

This is info for QA so that they can validate this. This is **mandatory** if this PR fixes a bug.
If this is a new feature, a good description in "What does this PR do" may be enough.

### Related info

An updated [RFC](https://github.com/SUSE/caasp-rfc/pull/65) for this PR.

### Status **BEFORE** applying the patch

We did not have a command to update the local cluster definition folder before, the only way is to create a temporary cluster definition folder and replace the existed local cluster definition folder. For example:
1. we have existed local cluster definition folder called `my-cluster`
2. create a temporary cluster definition folder: `skuba cluster init --control-plane <IP/FQDN> temp-cluster`
3. copy `temp-cluster/addons/<addon>/base/<addon>.yaml` to  `my-cluster/addons/<addon>/base/<addon>.yaml`

### Status **AFTER** applying the patch

We have a new command to update the local cluster definition folder to the current Kubernetes version. For example:
1. we have existed local cluster definition folder called `my-cluster`
2. run `skuba addon refresh localconfig` inside folder `my-cluster`

## Docs

https://github.com/SUSE/doc-caasp/pull/919

# Merge restrictions

(Please do not edit this)

We are in *v4-maintenance phase*, so we will restrict what can be merged to prevent unexpected surprises:

    What can be merged (merge criteria):
        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine
        there is a PR for updating documentation (or a statement that this is not needed)

PR #1247, opened 2020-07-10
Enhance skuba addon upgrade [plan|apply]

jenting:
## Why is this PR needed?

The users would use kustomize to overlay the default base addon manifest value.

However, we did not perform a dry-run check before applied to the current Kubernetes cluster, the user provides kustomize patch manifests might have some problem like manifest indentation error.

Therefore, for `skuba addon upgrade plan`, we not only display is there any new addon or existed addon need to be upgraded, moreover, we also perform the Kubernetes server-side validation by running `kubectl apply -k addons/<addon> --dry-run=server` and outputs the stderr to the user if present.

Besides, we'll generate `kustomization.yaml` inside `addons/<addon>` to make the user have the ability to run `kubectl apply -f addons/<addon>` after addons upgraded because the user might want to add more patches into the cluster and make sure the patches manifests will keep overtime for each addon upgrade.

By the way, before actually perform the `skuba addon upgrade [plan|apply]`, we will check the local addons base manifests match to the current Kubernetes cluster version, to make sure it's up-to-date.

The command `skuba addon refresh localconfig` is in another PR #1226 to not make a single PR too large.

Fixes https://github.com/SUSE/avant-garde/issues/1765

## What does this PR do?

1. Perform local addon base manifests check for `skuba addon upgrade [plan|apply]`.
2. Generates `kustomization.yaml` inside `addons/<addon>` for  `skuba addon upgrade [plan|apply]`
3. Perform Kubernetes server-side dry-run validation for `skuba addon upgrade plan`

N/A

## Info for QA

### Related info

An updated [RFC](https://github.com/SUSE/caasp-rfc/pull/65) for this PR.

1. The `kustomization.yaml` won't exist inside `addons/<addon>`.
2. The `skuba addon upgrade plan` won't do Kubernetes server-side dry-run.

1. It'll check the local addons base manifests is up-to-date for `skuba addon upgrade [plan|apply]`
2. The `kustomization.yaml` will exist after `skuba addon upgrade [plan|apply]`
3. The `skuba addon upgrade plan` will do Kubernetes server-side dry-run.

## Docs

https://github.com/SUSE/doc-caasp/pull/923

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/addons/addons.go
+++ b/internal/pkg/skuba/addons/addons.go
@@ -155,7 +155,7 @@ func addonsByPriority() []Addon {
 
 // DeployAddons loops over the sorted list of addons, checks if each needs to be deployed and
 // triggers its deployment
-func DeployAddons(client clientset.Interface, addonConfiguration AddonConfiguration) error {
+func DeployAddons(client clientset.Interface, addonConfiguration AddonConfiguration, dryRun bool) error {
 	skubaConfiguration, err := skuba.GetSkubaConfiguration(client)
 	if err != nil {
 		return err
@@ -172,7 +172,7 @@ func DeployAddons(client clientset.Interface, addonConfiguration AddonConfigurat
 			continue
 		}
 		if hasToBeApplied {
-			if err := addon.Apply(client, addonConfiguration, skubaConfiguration); err == nil {
+			if err := addon.Apply(client, addonConfiguration, skubaConfiguration, dryRun); err == nil {
 				klog.V(1).Infof("%q addon correctly applied", addonName)
 			} else {
 				klog.Errorf("failed to apply %q addon (%v)", addonName, err)
@@ -339,7 +339,7 @@ func (addon Addon) Write(addonConfiguration AddonConfiguration) error {
 // applyPreflight applies the preflight deployment/daemonset manifest if such
 // is defined by the addon. It returns a bool whether preflight was defined
 // by the addon and an error.
-func (addon Addon) applyPreflight(addonConfiguration AddonConfiguration, sandboxDir string) (bool, error) {
+func (addon Addon) applyPreflight(addonConfiguration AddonConfiguration, rootDir string, dryRun bool) (bool, error) {
 	renderedPreflightManifest, err := addon.RenderPreflight(addonConfiguration)
 	if err != nil {
 		return false, err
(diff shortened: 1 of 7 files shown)


pkg/skuba/actions/addon/upgrade/apply.go
```diff
@@ -36,6 +36,29 @@ func Apply(client clientset.Interface) error {
 	if err != nil {
 		return err
 	}
+
+	clusterConfiguration, err := kubeadm.GetClusterConfiguration(client)
```
innobead, line 40: The same validation code used in `apply` and `plan`. I would suggest to have a common validation for this to reuse.

jenting: Well, the validation already has the function `CheckLocalAddonsBaseManifests` to achieve it.

The only way to have a more common code is the wrap the error message to a single function/variable but IMO it's minor. Or do you have any better ideas :bow: ? 

```
        if !match {
		fmt.Println("Current local addons cluster folder configuration is out-of-date.")
		fmt.Println("Please run \"skuba addon refresh localconfig\" before you perform addon upgrade.")
		return nil
	}
```

innobead: Yes, it's minor, and just checked the code again where some variables are used before/after the context, so the way right now should be fine even there is some little duplication. It's fine for me now ;)

internal/pkg/skuba/addons/addons.go
```diff
@@ -340,28 +370,39 @@ func (addon Addon) Write(addonConfiguration AddonConfiguration) error {
 // applyPreflight applies the preflight deployment/daemonset manifest if such
 // is defined by the addon. It returns a bool whether preflight was defined
 // by the addon and an error.
-func (addon Addon) applyPreflight(addonConfiguration AddonConfiguration, sandboxDir string) (bool, error) {
+func (addon Addon) applyPreflight(addonConfiguration AddonConfiguration, rootDir string, dryRun bool) (bool, error) {
```
evrardjp, line 373: Why the change in the rootDir?

jenting: Just make the naming all the same cross files 😃

PR #1248, opened 2020-07-13
Revert kubelet server tls bootstrap

jenting:
## Why is this PR needed?

The kucero will not officially release at CaaSv4, so remove it from Kubernetes 1.17.

The kucero _must_ existed before kubelet `serverTLSBootstrap: true`, therefore, we revert the code
to configures `serverTLSBootstrap: true`.

We'll enable the kubelet `serverTLSBootstrap: true` in the next coming release.

## What does this PR do?

Revert the code for enabling kubelet `serverTLSBootstrap: true`, related to #1152.

It needs to backport to v5 branch `release-caasp-5.0.0`.

## Info for QA

### Related info

* upstream issues

The kubelet configuration `/var/lib/kubelet/config.yaml` would have `serverTLSBootstrap: true` in fresh install _or_ upgraded cluster.

The kubelet configuration `/var/lib/kubelet/config.yaml` would not have `serverTLSBootstrap: true` in fresh install _or_ upgraded cluster.

## Docs

https://github.com/SUSE/doc-caasp/pull/912

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubeadm/configmap_test.go
+++ b/internal/pkg/skuba/kubeadm/configmap_test.go
@@ -367,10 +367,7 @@ func TestUpdateClusterConfigurationWithClusterVersion(t *testing.T) {
 				}
 				initCfg.APIServer.ControlPlaneComponent.ExtraArgs["enable-admission-plugins"] = currentAdmissionPlugins
 			}
-			_, err := UpdateClusterConfigurationWithClusterVersion(&initCfg, tt.clusterVersion)
-			if err != nil {
-				t.Errorf("expected no error but an error returned: %v", err)
-			}
+			UpdateClusterConfigurationWithClusterVersion(&initCfg, tt.clusterVersion)
 			// Check admission plugins
 			gotAdmissionPlugins := initCfg.APIServer.ControlPlaneComponent.ExtraArgs["enable-admission-plugins"]
 			if gotAdmissionPlugins != expectedAdmissionPlugins {
--- a/internal/pkg/skuba/kubernetes/versions.go
+++ b/internal/pkg/skuba/kubernetes/versions.go
@@ -132,7 +132,6 @@ var (
 				Dex:           &AddonVersion{"2.16.0", 6},
 				Gangway:       &AddonVersion{"3.1.0-rev4", 5},
 				MetricsServer: &AddonVersion{"0.3.6", 1},
-				Kucero:        &AddonVersion{"1.1.1", 0},
 				PSP:           &AddonVersion{"", 4},
 			},
 		},
--- a/internal/pkg/skuba/deployments/ssh/kubelet.go
+++ b/internal/pkg/skuba/deployments/ssh/kubelet.go
@@ -18,11 +18,17 @@
 package ssh
 
 import (
+	"crypto/x509"
 	"io/ioutil"
+	"net"
 	"os"
 	"path/filepath"
+	"strings"
 
+	"github.com/pkg/errors"
+	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/kubernetes/cmd/kubeadm/app/constants"
+	"k8s.io/kubernetes/cmd/kubeadm/app/util/pkiutil"
 	"sigs.k8s.io/yaml"
 
 	"github.com/SUSE/skuba/internal/pkg/skuba/deployments"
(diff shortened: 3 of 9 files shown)


internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -132,7 +132,6 @@ var (
 				Dex:           &AddonVersion{"2.16.0", 6},
 				Gangway:       &AddonVersion{"3.1.0-rev4", 5},
 				MetricsServer: &AddonVersion{"0.3.6", 1},
-				Kucero:        &AddonVersion{"1.1.1", 0},
```
jenting, line 135: remove from Kubernets version `1.17`.

jenting:
This PR needs to backport to branch `release-caasp-5.0.0`, adding `backport` flag back.

jenting:
@evrardjp Could u please take a look, it need to backport to v5 otherwise the upgrade scenario would have metrics-server temporarily in crash loopback state unless the kucero addon installed later in "skuba addon upgrade apply", thx.

jenting:
> If we backport this, serverTLSBootstrap will not be set for 5.0.0, which is fine for upgrades I guess, but I am not sure if it's a good idea for greenfields. Do we intend to have this feature as opt-in ?

> What does "We'll enable the kubelet serverTLSBootstrap: true in the next coming release." mean here? Do we have an update/upgrade path?


For greenfield, it's fine to enable the kubelet "serverTLSBootstrap: true" since kucero addon will be deployed when node bootstrap so the kubelet server CSR signer _kucero_ is in cluster.

However, to make it in general behavior for greenfield and brownfield cluster, I'd suggest enable this in the skuba next release (along with kubernetes 1.19).

evrardjp:
I haven't tested, but it all makes sense. I would love seeing some extra eyes on this.

internal/pkg/skuba/kubeadm/configmap_test.go
```diff
@@ -367,10 +367,7 @@ func TestUpdateClusterConfigurationWithClusterVersion(t *testing.T) {
 				}
 				initCfg.APIServer.ControlPlaneComponent.ExtraArgs["enable-admission-plugins"] = currentAdmissionPlugins
 			}
-			_, err := UpdateClusterConfigurationWithClusterVersion(&initCfg, tt.clusterVersion)
-			if err != nil {
-				t.Errorf("expected no error but an error returned: %v", err)
-			}
```
dannysauer, line 373: This seems odd to remove.  Is an error expected in the old version of k8s? :)

dannysauer: Ok, I see that the function it calls returns differently when the commit is reverted.  I guess this makes sense then.

dannysauer:
> For greenfield, it's fine to enable the kubelet "serverTLSBootstrap: true" since kucero addon will be deployed when node bootstrap so the kubelet server CSR signer _kucero_ is in cluster.

I'm not sure if I understand this.  It looks like the other parts necessary are in place already; why would we _not_ want to enable TLS boostrapping independent of having kucero manage cert renewal?  Seems like that's generally a good idea, and a bootstrap failure due to an expired cluster CA cert would be acceptable (since that cert will need updated anyway due to all the other stuff that will also be failing).  Am I missing something important?

jenting:
> > For greenfield, it's fine to enable the kubelet "serverTLSBootstrap: true" since kucero addon will be deployed when node bootstrap so the kubelet server CSR signer _kucero_ is in cluster.
> 
> I'm not sure if I understand this. It looks like the other parts necessary are in place already; why would we _not_ want to enable TLS boostrapping independent of having kucero manage cert renewal? Seems like that's generally a good idea, and a bootstrap failure due to an expired cluster CA cert would be acceptable (since that cert will need updated anyway due to all the other stuff that will also be failing). Am I missing something important? :)

The problem is when the admin upgrades from 1.17 -> 1.18, the admin would do
- `skuba node upgrade apply` (loop all control plane nodes and worker nodes)
- `skuba addon upgrade apply`

then at the first control plane node upgrade, _usually_ the admin checks all the pods in the stable state, however, since the first control node sets `serverTLSBootstrap: true` which means it sends out kubelet server CSR in the cluster but there is no CSR signer right now (kucero installed later in `skuba addon upgrade apply`).

therefore, the metrics-server would be crash loopback state because the kubelet server does not have it's server certificate right now (it seems like the kubelet does not honor in-disk `kubelet.crt` which is generated by skuba if `serverTLSBootstrap: true`.

if the admin is willing to accept the temporarily metrics-server unstable state during cluster upgrade, I'm fine to close this PR; otherwise, this PR should proceed because it solves the kubelet server certificate will expires after 1 year.

PR #1251, opened 2020-07-14
cri: handle the migration to the new configuration format

mssola:
## Why is this PR needed?

Fixes SUSE/avant-garde#1679
Fixes SUSE/avant-garde#1785
See SUSE/avant-garde#1679
See SUSE/caasp-release-notes#51

## What does this PR do?

Ensure that the local files of the user are properly migrated to the new cri-o configuration format.

## Info for QA

In order to test this you should first create a v4.2.2 cluster with an old skuba. Then, use a skuba built with `master` with this patch applied, and call `skuba cluster upgrade config`. You should see:

- The old `addons/cri/default_flags` should've been removed.
- Now you should have `addons/cri/conf.d` with `01-caasp.conf` and `README` in it.

With this SUSE/caasp-release-notes#51 should not be needed.

### DIFF
--- a/cmd/skuba/cluster/upgrade.go
+++ b/cmd/skuba/cluster/upgrade.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/klog"
 
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
+	"github.com/SUSE/skuba/internal/pkg/skuba/upgrade/cluster"
 	"github.com/SUSE/skuba/pkg/skuba/actions/cluster/upgrade"
 )
 
@@ -37,6 +38,7 @@ func NewUpgradeCmd() *cobra.Command {
 
 	cmd.AddCommand(
 		newUpgradePlanCmd(),
+		newUpgradeLocalConfigCmd(),
 	)
 
 	return cmd
@@ -54,6 +56,20 @@ func newUpgradePlanCmd() *cobra.Command {
 			}
 			if err := upgrade.Plan(clientSet); err != nil {
 				fmt.Printf("Unable to plan cluster upgrade: %s\n", err)
+				os.Exit(1)
+			}
+		},
+		Args: cobra.NoArgs,
+	}
+}
+
+func newUpgradeLocalConfigCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "localconfig",
+		Short: "Upgrades the local configuration",
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := cluster.CriMigrate(); err != nil {
+				klog.Errorf("unable to upgrade the cluster configuration: %s\n", err)
 				os.Exit(1)
 			}
 		},
--- a/internal/pkg/skuba/deployments/ssh/cri.go
+++ b/internal/pkg/skuba/deployments/ssh/cri.go
@@ -83,30 +83,9 @@ func criConfigure(t *Target, data interface{}) error {
 	return err
 }
 
+// criSysconfig will enforce the package sysconfig configuration.
 func criSysconfig(t *Target, data interface{}) error {
-	criFiles, err := ioutil.ReadDir(skuba.CriDir())
-	if err != nil {
-		return errors.Wrap(err, "Could not read local cri directory: "+skuba.CriDir())
-	}
-	defer func() {
-		_, _, err := t.ssh("rm -rf /tmp/cri.d")
-		if err != nil {
-			// If the deferred function has any return values, they are discarded when the function completes
-			// https://golang.org/ref/spec#Defer_statements
-			fmt.Println("Could not delete the cri.d config path")
-		}
-	}()
-
-	for _, f := range criFiles {
-		if err := t.target.UploadFile(filepath.Join(skuba.CriDir(), f.Name()), filepath.Join("/tmp/cri.d", f.Name())); err != nil {
-			return err
-		}
-	}
-
-	if _, _, err = t.ssh("mv -f /etc/sysconfig/crio /etc/sysconfig/crio.backup"); err != nil {
-		return err
-	}
-	_, _, err = t.ssh("mv -f /tmp/cri.d/default_flags /etc/sysconfig/crio")
+	_, _, err := t.ssh("cp -f /usr/share/fillup-templates/sysconfig.crio /etc/sysconfig/crio")
 	return err
 }
 
--- a/internal/pkg/skuba/upgrade/cluster/cri.go
+++ b/internal/pkg/skuba/upgrade/cluster/cri.go
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2020 SUSE LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package cluster
+
+import (
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
+	"github.com/SUSE/skuba/pkg/skuba"
+	clusterinit "github.com/SUSE/skuba/pkg/skuba/actions/cluster/init"
+)
+
+// CriMigrate migrates the old configuration of cri-o < 1.17 to the new format.
+func CriMigrate() error {
+	_, err := os.Stat(skuba.CriDockerDefaultsConfFile())
+	if os.IsNotExist(err) {
+		return nil
+	}
+
+	if err := criGenerateLocalConfiguration(); err != nil {
+		return err
+	}
+
+	if err := criRemoveLocalOldFile(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func criGenerateLocalConfiguration() error {
+	cfg := clusterinit.InitConfiguration{
+		PauseImage:        kubernetes.ComponentContainerImageForClusterVersion(kubernetes.Pause, kubernetes.LatestVersion()),
+		StrictCapDefaults: !criHadStrictCapDefaults(),
+	}
+
+	files := clusterinit.CriScaffoldFiles["criconfig"]
+	for _, file := range files {
+		// Note well: this code will remove the whole local cri configuration
+		// directory if something goes wrong. This is done so to avoid weird
+		// intermediary states in case a file presented a problem. Ideally this
+		// shouldn't happen (it's more of a all or nothing scenario), but take
+		// it into account if you want to reuse this code.
+		if err := clusterinit.WriteScaffoldFile(file, cfg); err != nil {
+			_ = os.RemoveAll(skuba.CriConfDir())
+			return err
+		}
+	}
+	return nil
+}
+
+func criHadStrictCapDefaults() bool {
+	data, err := ioutil.ReadFile(skuba.CriDockerDefaultsConfFile())
+	if err != nil {
+		return false
+	}
+	return strings.Contains(string(data), "--default-capabilities")
+}
+
+func criRemoveLocalOldFile() error {
+	_, err := os.Stat(skuba.CriDockerDefaultsConfFile())
+	if os.IsNotExist(err) {
+		return nil
+	}
+	return os.Remove(skuba.CriDockerDefaultsConfFile())
+}
--- a/pkg/skuba/actions/cluster/init/init.go
+++ b/pkg/skuba/actions/cluster/init/init.go
@@ -142,11 +142,7 @@ func isAddonRequired(addon addons.Addon, initConfiguration InitConfiguration) bo
 }
 
 func writeScaffoldFiles(initConfiguration InitConfiguration) error {
-	scaffoldFilesToWrite := criScaffoldFiles["criconfig"]
-	kubernetesVersion := initConfiguration.KubernetesVersion
-	if kubernetesVersion.Minor() < 18 {
-		scaffoldFilesToWrite = criScaffoldFiles["sysconfig"]
-	}
+	scaffoldFilesToWrite := CriScaffoldFiles["criconfig"]
 
 	if len(initConfiguration.CloudProvider) > 0 {
 		if cloudScaffoldFiles, found := cloudScaffoldFiles[initConfiguration.CloudProvider]; found {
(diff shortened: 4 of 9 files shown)


mjura:
I would rename this new command to `skuba cluster update config` or `skuba cluster update config` because `skuba cluster upgrade apply` suggests that it will upgrade all nodes in the cluster

pkg/skuba/actions/cluster/init/init.go
```diff
@@ -141,10 +141,10 @@ func isAddonRequired(addon addons.Addon, initConfiguration InitConfiguration) bo
 }
 
 func writeScaffoldFiles(initConfiguration InitConfiguration) error {
-	scaffoldFilesToWrite := criScaffoldFiles["criconfig"]
+	scaffoldFilesToWrite := CriScaffoldFiles["criconfig"]
 	kubernetesVersion := initConfiguration.KubernetesVersion
 	if kubernetesVersion.Minor() < 18 {
```
evrardjp, line 146: This condition can be removed, we won't provision 1.17 with this branch of skuba.

kkaempf:
How can this be tested ?

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -114,3 +117,47 @@ func criStart(t *Target, data interface{}) error {
...
+
+	if err := criRemoveLocalOldFile(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func criGenerateLocalConfiguration() error {
+	cfg := clusterinit.InitConfiguration{
+		PauseImage:        kubernetes.ComponentContainerImageForClusterVersion(kubernetes.Pause, kubernetes.LatestVersion()),
```
evrardjp, line 135: is LatestVersion correct, or should that be the _next_ version?

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -114,3 +117,47 @@ func criStart(t *Target, data interface{}) error {
...
+	for _, file := range files {
+		if err := clusterinit.WriteScaffoldFile(file, cfg); err != nil {
+			_ = os.RemoveAll(skuba.CriConfDir())
+			return err
+		}
+	}
+	return nil
+}
+
+func criHadStrictCapDefaults() bool {
```
evrardjp, line 149: Very nice!

cmd/skuba/cluster/upgrade.go
```diff
@@ -60,3 +62,17 @@ func newUpgradePlanCmd() *cobra.Command {
 		Args: cobra.NoArgs,
 	}
 }
+
+func newUpgradeApplyCmd() *cobra.Command {
```
evrardjp, line 66: I am fine with Michal's proposal to have skuba cluster upgrade config

internal/pkg/skuba/deployments/ssh/cri.go
mssola: At least this is giving me this result: `registry.suse.de/devel/caasp/5/containers/containers/caasp/v5/pause:3.2`, which I believe is fine, no?

evrardjp: It is fine for now, but it might not be in the future, if the user is lagging behind more than a version. No?

mssola:
> How can this be tested ?

Updated the PR.

internal/pkg/skuba/deployments/ssh/cri.go
mssola: Well, in this case they should be using the proper `skuba` for each scenario (same as other upgrades). I don't think we have a `next` version somewhere in the code, right?

evrardjp: Mmm interesting, I thought we had that, because we have a path in the `plan` and `apply` commands, when dealing with multiple k8s versions, but I didn't check the code.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -130,24 +130,20 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
-			return err
-		}
-	} else if _, err := os.Stat(skuba.CriDockerDefaultsConfFile()); err == nil {
-		err = target.Apply(nil, "cri.sysconfig")
-		if err != nil {
-			return err
-		}
+	// Always upload crio files, regardless of the version (allows to enforce
+	// user behavior during patch updates). During the upgrade from 1.16 to 1.18
+	// crio, the cri-o package will handle overriding the old crio sysconfig to
```
evrardjp, line 135: The cri-o package doesn't handle overriding sysconfig, so that's misleading. It's merely merging the sysconfigs, which in our case is wrong, because it keeps old content. 

That comment is misleading and should be reworded.

cmd/skuba/cluster/upgrade.go
```diff
@@ -60,3 +62,17 @@ func newUpgradePlanCmd() *cobra.Command {
 		Args: cobra.NoArgs,
 	}
 }
+
+func newUpgradeConfigCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "config",
+		Short: "Upgrades the local configuration",
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := ssh.CriMigrate(); err != nil {
```
evrardjp, line 71: this is very misleading. You are saying migrating local configuration, yet this function run things on the node.

For me the upgrade localconfig should be dealing with just local things, while the things happening on target nodes can be dealt with during `node upgrade`.

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -93,7 +96,7 @@ func criSysconfig(t *Target, data interface{}) error {
 		if err != nil {
 			// If the deferred function has any return values, they are discarded when the function completes
 			// https://golang.org/ref/spec#Defer_statements
-			fmt.Println("Could not delete the cri.d config path")
+			fmt.Println("Could not delete the temporary cri.d config path")
```
evrardjp, line 99: the criSysconfig function is rightfully called, but it doesn't do anything interesting for us in the current code.
This function should be refactored to instead copy sysconfig from the package as a default sysconfig for crio in `/etc/sysconfig` . This way, the node migration will automatically ensure that sysconfig is always the default one, and the local configuration is rightfully applied with criConfigure.

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -114,3 +117,47 @@ func criStart(t *Target, data interface{}) error {
 	_, _, err := t.ssh("systemctl", "enable", "--now", "crio")
 	return err
 }
+
+// CriMigrate migrates the old configuration of cri-o < 1.17 to the new format.
+func CriMigrate() error {
+	if err := criGenerateLocalConfiguration(); err != nil {
```
evrardjp, line 123: the `deployments ssh cri.go` is supposed to happen on the node. This should not touch the local configuration file on the management node.

I don't think it's a good idea to have this function in here.

internal/pkg/skuba/deployments/ssh/cri.go
```diff
@@ -114,3 +117,47 @@ func criStart(t *Target, data interface{}) error {
...
+		return err
+	}
+
+	if err := criRemoveLocalOldFile(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func criGenerateLocalConfiguration() error {
```
evrardjp, line 133: See also my comment for "CriMigrate".

internal/pkg/skuba/deployments/ssh/testdata/addons/cri/default_flags
```diff
@@ -0,0 +1,7 @@
+## Path           : System/Management
```
evrardjp, line 1: We should NOT provide this. This should be deleted. We should trust the sysconfig coming from the package instead, and copy it from the package (in `/usr/` iirc), to the `/etc/sysconfig/`.

pkg/skuba/actions/cluster/init/init.go
```diff
@@ -163,31 +159,41 @@ func writeScaffoldFiles(initConfiguration InitConfiguration) error {
...
-		}
-		str, err := renderTemplate(file.Content, initConfiguration)
-		if err != nil {
-			return errors.Wrap(err, "unable to render template")
-		}
-		_, err = f.WriteString(str)
-		if err != nil {
-			return errors.Wrapf(err, "unable to write template to file %s", f.Name())
+		if err := WriteScaffoldFile(file, initConfiguration); err != nil {
+			return err
```
evrardjp, line 163: nice refactoring.
nit: "WriteScaffoldFile" doesn't mean anything to me though. If people understand what it means, good. Else I would try to rename it with something clearly explaining what it does, like WriteTemplatedFile.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -130,24 +130,20 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
+	// an "empty" sysconfig, and the cri.configure action will be enough.
+	if _, err := os.Stat(skuba.CriDefaultsConfFile()); err != nil {
+		return err
```
evrardjp, line 138: I am not sure if that error is very explicit. It will just say no such file or directory if the people have not upgraded their local config, right?

Should we be clearer?

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -130,24 +130,20 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
+	}
+	err = target.Apply(nil, "cri.configure")
+	if err != nil {
+		return err
+	}
+	err = target.Apply(nil, "cri.sysconfig")
```
evrardjp, line 144: please note you can chain the apply states. See my suggestion above.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -130,24 +130,20 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
-			return err
-		}
+	// Always upload crio files, regardless of the version (allows to enforce
+	// user behavior during patch updates). During the upgrade from 1.16 to 1.18
+	// crio, the cri-o package will handle overriding the old crio sysconfig to
+	// an "empty" sysconfig, and the cri.configure action will be enough.
+	if _, err := os.Stat(skuba.CriDefaultsConfFile()); err != nil {
+		return err
+	}
+	err = target.Apply(nil, "cri.configure")
```
evrardjp, line 140:
```suggestion
	err = target.Apply(nil, "cri.configure", "cri.sysconfig")
```

pkg/skuba/actions/cluster/init/init.go
mssola: I preserved the name on purpose, because I've seen the concept of `scaffold` here and there when it comes to templates. Maybe this should go into another PR that focuses on the naming of things. That being said, and maybe I'm biased because of [Rails](https://github.com/rails/rails), to me `scaffold` and `template` and pretty much synonyms, and in this context I understand it like `configuration and project files that the tool is going to automatically produce for you when calling X command`. So, in this case, it makes sense.

If other people disagree, as I said, it can go into another PR. But I'd say that this kind of change would be out of place for this PR.

internal/pkg/skuba/deployments/ssh/testdata/addons/cri/default_flags
mssola: Bear in mind that this is just test data. This is not planned to be installed anywhere.

cmd/skuba/cluster/upgrade.go
mssola: This function does not touch files on remote nodes. It just updates the local configuration. Thus, I see two things:

1. It should not go into the `ssh` package, as you said in another comment.
2. We may want to change the subcommand name from `config` to `localconfig`, so to reinforce the idea of changes only happening locally, as it has been suggested. What do you think?

internal/pkg/skuba/upgrade/cluster/cri.go
```diff
@@ -0,0 +1,72 @@
...
+func criGenerateLocalConfiguration() error {
+	cfg := clusterinit.InitConfiguration{
+		PauseImage:        kubernetes.ComponentContainerImageForClusterVersion(kubernetes.Pause, kubernetes.LatestVersion()),
+		StrictCapDefaults: !criHadStrictCapDefaults(),
+	}
+
+	files := clusterinit.CriScaffoldFiles["criconfig"]
+	for _, file := range files {
+		if err := clusterinit.WriteScaffoldFile(file, cfg); err != nil {
+			_ = os.RemoveAll(skuba.CriConfDir())
```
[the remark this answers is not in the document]
mssola: Not exactly. `clusterinit.WriteScaffoldFile(file, cfg)` will write scaffold files into that directory. Thus, any files with the same names will be effectively replaced. This means that customers have to back things up before performing the migration (as in with any migration).

That being said, if, for whatever reason, we fail to write these files, we should remove everything instead of sticking into an intermediary result (e.g. imagine that one file was successfully written but the other didn't). We don't want these kinds of scenarios, it's all or nothing.

Therefore, I think that this is actually  the safest approach.

internal/pkg/skuba/upgrade/cluster/cri.go
```diff
@@ -0,0 +1,72 @@
...
+
+func criGenerateLocalConfiguration() error {
+	cfg := clusterinit.InitConfiguration{
+		PauseImage:        kubernetes.ComponentContainerImageForClusterVersion(kubernetes.Pause, kubernetes.LatestVersion()),
+		StrictCapDefaults: !criHadStrictCapDefaults(),
+	}
+
+	files := clusterinit.CriScaffoldFiles["criconfig"]
+	for _, file := range files {
+		if err := clusterinit.WriteScaffoldFile(file, cfg); err != nil {
```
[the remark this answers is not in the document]
mssola: Right now this is the case, yes. But it has been defined just like the template, and who knows what we will do in the future. Thus, I think it's easier for the implementation and it makes sense overall to keep it like this.

internal/pkg/skuba/upgrade/cluster/cri.go
```diff
@@ -0,0 +1,72 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func criHadStrictCapDefaults() bool {
+	data, err := ioutil.ReadFile(skuba.CriDockerDefaultsConfFile())
+	if err != nil {
+		return false
```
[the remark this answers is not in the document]
mssola: Thinking about it, now since we'll check that the file actually exists at the very beginning of CriMigrate, this should not happen anymore.

internal/pkg/skuba/upgrade/cluster/cri.go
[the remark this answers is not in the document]
evrardjp: > Not exactly. `clusterinit.WriteScaffoldFile(file, cfg)` will write scaffold files into that directory. Thus, any files with the same names will be effectively replaced. This means that customers have to back things up before performing the migration (as in with any migration).

I think in this case, the files being replaced are the "caasp defaults", which is fine, and the customer shouldn't touch them, so we are good. Yes, it's indeed safer to backup, while not necessary.

> 
> That being said, if, for whatever reason, we fail to write these files, we should remove everything instead of sticking into an intermediary result (e.g. imagine that one file was successfully written but the other didn't). We don't want these kinds of scenarios, it's all or nothing.

That's what I am not getting my head around. For me, we should fail, and say "There is something wrong that happened here, maybe you can figure it out with this error: %err". Not deleting the whole folder, which might contain user configuration on top of our default configs. That's very disruptive to me.

I agree it's better to bail, but I don't think it's better to delete the folder.

> 
> Therefore, I think that this is actually the safest approach.

Do you mean that, because this migration only runs ONCE, it should be okay to delete the folder, as there is no user configuration yet, and that code will probably never be used for something else? I am fine with that, but I would prefer to add this explicitly in the comments. A comment just before the dir removal should be enough, to clarify that "this is okay to do so because this process only runs once, so there should be no failure, and there should be no user configuration at this point yet"

PR #1253, opened 2020-07-14
Remove cri-tools on upgrade


evrardjp:
cri-tools-1.18 is well installed for new clusters, but
it's incorrectly not removed during the first upgrade for the
caasp 4 to 5 migration.

This should fix it.


kkaempf:
ship it !

evrardjp:
Needs backporting.

PR #1256, opened 2020-07-14
Display the supported status of a node in skuba cluster status

evrardjp:
This is a first pass at showing the supported state of a cluster,
without adding annotations on the final nodes, by simply relying
on skuba's code. It shows to the users that the cluster is
in an unsupported state, which should help trigger the decision
to upgrade.

## Why is this PR needed?

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/pkg/skuba/actions/cluster/status/status.go
+++ b/pkg/skuba/actions/cluster/status/status.go
@@ -19,9 +19,11 @@ package cluster
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"strings"
 
+	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


pkg/skuba/actions/cluster/status/status.go
```diff
@@ -55,19 +57,29 @@ func Status(client clientset.Interface) error {
...
 		"OS-IMAGE:.status.nodeInfo.osImage," +
-		"KERNEL-VERSION:.status.nodeInfo.kernelVersion," +
-		"KUBELET-VERSION:.status.nodeInfo.kubeletVersion," +
+		"KERNEL:.status.nodeInfo.kernelVersion," +
+		"KUBELET:.status.nodeInfo.kubeletVersion," +
 		"CONTAINER-RUNTIME:.status.nodeInfo.containerRuntimeVersion," +
 		"HAS-UPDATES:.metadata.annotations.caasp\\.suse\\.com/has-updates," +
 		"HAS-DISRUPTIVE-UPDATES:.metadata.annotations.caasp\\.suse\\.com/has-disruptive-updates," +
-		"CAASP-RELEASE-VERSION:.metadata.annotations.caasp\\.suse\\.com/caasp-release-version"
+		"RELEASE:.metadata.annotations.caasp\\.suse\\.com/caasp-release-version"
```
jenting, line 82: Remember to update the related readme & documentation.

kkaempf:
This is missing a test case (resp. test description)

kkaempf:
If the intention is to flag the 'zombie' state between a "zypper migration" (SP1 to SP2) and the "skuba migration" (v4 to v5), then 'unsupported' is not expressing the actual status. If a customer gets stuck in the migration, the cluster/node is still supported (in the sense of "you can call support for help").

kkaempf:
That said, flagging the "zombie" state is the right thing to do, so :+1: on the general direction of this PR.

I just wonder if we can find a better wording instead of "unsupported" ?
"unfinished migration" maybe ? "incomplete" ? "updates missing" ?
We could as well tell customers the command that's getting the node out of the zombie state, like `'skuba update' pending` ?!

evrardjp:
Because the stdout already contains numerous details, I didn't want to write extensive details.

We could have something like "Up to date" with values:
- green tick (cluster is up to date);
- warning sign (outdated);
- zombie (unknown state);
- red stop sign (unsupported, too old).

pkg/skuba/actions/cluster/status/status.go
evrardjp: Right :)

evrardjp:
I don't think it would be smart for me to introduce a new logic in this, just before my holidays, if this starts to break. I hope someone can take it away, and add testing. It shouldn't take very long.

evrardjp:
I considered myself that it was a good idea to implement for our users (especially in those migration times), but I have no stake in this game.

If someone wants a feature like this, let's put it in the backlog.

jordimassaguerpla:
> I considered myself that it was a good idea to implement for our users (especially in those migration times), but I have no stake in this game.
> 
> If someone wants a feature like this, let's put it in the backlog.

We would need an issue in github/SUSE/avant-garde to plan for this. Thanks.

evrardjp:
This was just a PoC. If someone wants it, let's make a card for it indeed.

PR #1259, opened 2020-07-15
Ensure services are enabled

evrardjp:
When removing a package and re-installing a new one, like we
do in the upgrade, the systemd service will be automatically
disabled. We should always ensure the services for cri-o
and kubelet are started.

None

## Info for QA

Upon 1.18 upgrade, `systemctl status crio` should show the service `enabled`.
Current state without this patch is `disabled`.

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


jordimassaguerpla:
I see there is the backport label. Please remember to create a PR to backport this to the release-caasp-5.0.0 branch for its inclusion into the release.

PR #1277, opened 2020-07-23
Drain nodes before kubadm upgrade and uncordon afterwards

saschagrunert:
## Why is this PR needed?

This implements a basic drain and uncordon before/after running `kubeadm upgrade`. To achieve that we also have to add a new helper function `kubernetes.UncordonNode()`.

Fixes https://github.com/SUSE/avant-garde/issues/1842

## What does this PR do?

Implements drain and uncordon during kubeadm upgrade.

None

## Info for QA

### Related info

None

No drain or uncordon on upgrade.

Check if the nodes get drained when doing an upgrade.

## Docs

We could add some troubleshooting to help users if the upgrade fails and the node is still cordon'ed.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/pkg/skuba/actions/node/upgrade/apply.go
+++ b/pkg/skuba/actions/node/upgrade/apply.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	clientset "k8s.io/client-go/kubernetes"
@@ -116,6 +117,13 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 		}
 	}
 
+	const drainTimeout = 15 * time.Minute
+	node := nodeVersionInfoUpdate.Current.Node
+	fmt.Printf("Draining node %s (timeout %dmin)\n", target.Nodename, drainTimeout)
+	if err := kubernetes.DrainNode(client, node, drainTimeout); err != nil {
+		return errors.Wrapf(err, "draining node %s", target.Nodename)
+	}
+
 	fmt.Printf("Performing node %s (%s) upgrade, please wait...\n", target.Nodename, target.Target)
 
 	if skubaUpdateWasEnabled {
@@ -194,6 +202,11 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 		if err := kured.Unlock(client); err != nil {
 			return err
 		}
+	}
+
+	fmt.Printf("Uncordon node %s\n", target.Nodename)
+	if err := kubernetes.UncordonNode(client, node); err != nil {
+		return errors.Wrapf(err, "uncordon node %s", target.Nodename)
 	}
 
 	fmt.Printf("Node %s (%s) successfully upgraded\n", target.Nodename, target.Target)
--- a/internal/pkg/skuba/kubernetes/nodes.go
+++ b/internal/pkg/skuba/kubernetes/nodes.go
@@ -66,6 +66,26 @@ func IsControlPlane(node *corev1.Node) bool {
 	return isControlPlane
 }
 
+// UncordonNode makes a node scheduleable again
+func UncordonNode(client clientset.Interface, node *corev1.Node) error {
+	cordonHelper := kubectldrain.NewCordonHelper(node)
+
+	// Check if already uncordoned
+	if updateRequired := cordonHelper.UpdateIfRequired(false); !updateRequired {
+		return nil
+	}
+
+	err, patchErr := cordonHelper.PatchOrReplace(client, false)
+	if patchErr != nil {
+		return errors.Wrap(patchErr, "creating node patch")
+	}
+	if err != nil {
+		return errors.Wrap(err, "updating node status")
+	}
+
+	return nil
+}
+
 // DrainNode cordons, drains and evict given node.
 func DrainNode(client clientset.Interface, node *corev1.Node, drainTimeout time.Duration) error {
 	policyGroupVersion, err := kubectldrain.CheckEvictionSupport(client)
--- a/internal/pkg/skuba/kubernetes/nodes_test.go
+++ b/internal/pkg/skuba/kubernetes/nodes_test.go
@@ -397,3 +397,35 @@ func TestDrainNode(t *testing.T) {
 		})
 	}
 }
+
+func TestUncordonNode(t *testing.T) {
+	for _, tc := range []struct {
+		client    *fake.Clientset
+		node      *corev1.Node
+		expectNil bool
+	}{
+		{
+			client:    fake.NewSimpleClientset(&corev1.NodeList{Items: []corev1.Node{{}}}),
+			node:      &corev1.Node{Spec: corev1.NodeSpec{Unschedulable: true}},
+			expectNil: true,
+		},
+		{
+			client:    fake.NewSimpleClientset(&corev1.NodeList{Items: []corev1.Node{{}}}),
+			node:      &corev1.Node{Spec: corev1.NodeSpec{Unschedulable: false}},
+			expectNil: true,
+		},
+		{
+			client:    fake.NewSimpleClientset(&corev1.NodeList{Items: []corev1.Node{}}),
+			node:      &corev1.Node{Spec: corev1.NodeSpec{Unschedulable: true}},
+			expectNil: false,
+		},
+	} {
+		err := UncordonNode(tc.client, tc.node)
+		if tc.expectNil && err != nil {
+			t.Errorf("expected nil err, got %v", err)
+		}
+		if !tc.expectNil && err == nil {
+			t.Errorf("expected non nil err")
+		}
+	}
+}


pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -116,6 +117,13 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 		}
 	}
 
+	const drainTimeout = 15 * time.Minute
+	node := nodeVersionInfoUpdate.Current.Node
```
saschagrunert, line 121: I'm not sure if that is the right node to be referenced. :thinking:

chentex: I would say is the correct one. I have doubts if this `kubernetes.GetNodeWithMachineID(client, target.Nodename)` would work, because I don't think `MachineID` is the same as `Nodename`.

chentex:
> I think in general this change is ok. But specifically for the v4 to v5 upgrade the node drain might come too late at this point in time.
> 
> This code is only executed after the node was upgrade from SLE15 SP1 to SP2, which is something that happens outside of skuba (and quite some time before calling any skuba commands). At least when taking the upgrade instruction from [SUSE/avant-garde#1769](https://github.com/SUSE/avant-garde/issues/1769) into account.

Totally agree, before doing the migration on the node, it needs to be drained, but that is going to have to be a manual step for the customers.

I am currently testing this with @mssola and we can confirm that the drain comes during `skuba node upgrade` at the correct time.

But after that the upgrade process is failing with this error:

```
Performing node master01 (10.86.1.35) upgrade, please wait...
I0728 11:31:02.101934     631 states.go:35] === applying state skuba-update-timer.disable ===
I0728 11:31:02.272137     631 ssh.go:183] running command: "systemctl disable --now skuba-update.timer"
Unable to apply node upgrade: failed to apply state skuba-update-timer.disable: Process exited with status 1
I0728 11:31:02.486664     631 ssh.go:206] Failed to disable unit: The name org.freedesktop.PolicyKit1 was not provided by any .service files

```

So we cannot confirm that the node is being `uncordoned` after the process.

If anyone has seen this error and can guide us it would be awesome.

chentex:
After running `skuba node upgrade apply` on the master node (there is only one in this scenario) it failed with this error
 
```
I0729 13:47:39.508831   31009 states.go:35] === applying state kubernetes.enable-services ===
I0729 13:47:39.734901   31009 ssh.go:183] running command: "sudo sh -c 'systemctl enable crio kubelet'"
I0729 13:47:40.070836   31009 ssh.go:206] Created symlink /etc/systemd/system/multi-user.target.wants/crio.service → /usr/lib/systemd/system/crio.service.
I0729 13:47:40.100795   31009 ssh.go:206] Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /usr/lib/systemd/system/kubelet.service.
I0729 13:47:41.037570   31009 states.go:40] === state kubernetes.enable-services applied successfully ===
I0729 13:47:41.037587   31009 states.go:35] === applying state skuba-update.start.no-block ===
I0729 13:47:41.304070   31009 ssh.go:183] running command: "sudo sh -c 'systemctl start --no-block skuba-update'"
I0729 13:47:42.066056   31009 states.go:40] === state skuba-update.start.no-block applied successfully ===
I0729 13:47:42.066072   31009 states.go:35] === applying state skuba-update-timer.enable ===
I0729 13:47:42.271791   31009 ssh.go:183] running command: "sudo sh -c 'systemctl enable --now skuba-update.timer'"
I0729 13:47:42.607096   31009 ssh.go:206] Created symlink /etc/systemd/system/timers.target.wants/skuba-update.timer → /usr/lib/systemd/system/skuba-update.timer.
I0729 13:47:43.882897   31009 states.go:40] === state skuba-update-timer.enable applied successfully ===
Unable to apply node upgrade: unable to patch daemonset with kured unlocking annotation: Patch https://10.86.1.126:6443/apis/apps/v1/namespaces/kube-system/daemonsets/kured: EOF
```

Could not get status of the cluster for a couple of minutes

```
$ skuba cluster status
E0729 13:48:33.635938   31703 status.go:43] unable to get cluster status: could not retrieve node list: an error on the server ("") has prevented the request from succeeding (get nodes)
```

After I got the status but the `uncordon` part never got to run

```
$ skuba cluster status                                                                                                                                                            
NAME       STATUS                     ROLE     OS-IMAGE                              KERNEL-VERSION      KUBELET-VERSION   CONTAINER-RUNTIME   HAS-UPDATES   HAS-DISRUPTIVE-UPDATES   CAASP-RELEASE-VERSION
master01   Ready,SchedulingDisabled   master   SUSE Linux Enterprise Server 15 SP2   5.3.18-22-default   v1.18.6           cri-o://1.18.2      no            no                       4.2.2
```

I manually uncordon it

```
kubectl --kubeconfig admin.conf uncordon master01
node/master01 uncordoned
```
Apparently it is all normal

```
skuba cluster status
NAME       STATUS   ROLE     OS-IMAGE                              KERNEL-VERSION      KUBELET-VERSION   CONTAINER-RUNTIME   HAS-UPDATES   HAS-DISRUPTIVE-UPDATES   CAASP-RELEASE-VERSION
master01   Ready    master   SUSE Linux Enterprise Server 15 SP2   5.3.18-22-default   v1.18.6           cri-o://1.18.2      no            no                       4.2.2
worker01   Ready    <none>   SUSE Linux Enterprise Server 15 SP2   5.3.18-22-default   v1.17.9           cri-o://1.16.1      no            no                       4.2.2
worker02   Ready    <none>   SUSE Linux Enterprise Server 15 SP2   5.3.18-22-default   v1.17.9           cri-o://1.16.1      no            no                       4.2.2
```

Both workers were updated without issues, both were drained and then uncordoned to be ready again.

```
$ skuba cluster status
NAME       STATUS   ROLE     OS-IMAGE                              KERNEL-VERSION      KUBELET-VERSION   CONTAINER-RUNTIME   HAS-UPDATES   HAS-DISRUPTIVE-UPDATES   CAASP-RELEASE-VERSION
master01   Ready    master   SUSE Linux Enterprise Server 15 SP2   5.3.18-22-default   v1.18.6           cri-o://1.18.2      no            no                       4.2.2
worker01   Ready    <none>   SUSE Linux Enterprise Server 15 SP2   5.3.18-22-default   v1.18.6           cri-o://1.18.2      no            no                       5
worker02   Ready    <none>   SUSE Linux Enterprise Server 15 SP2   5.3.18-22-default   v1.18.6           cri-o://1.18.2      no            no                       5
```

I'd say the issue I have was because of only one master in the cluster.

saschagrunert:
Thanks for the detailed test and explanations @chentex! :pray:

drpaneas:
what happens if the node-draining times out? This is most likely to happen if there are hundreds of customer-application pods running on that node doing all sorts of stuff. If that happens, maybe try to identify which pods stuck in _terminating_ and try to delete them by force. Then re-try draining that specific node.

saschagrunert:
> what happens if the node-draining times out? This is most likely to happen if there are hundreds of customer-application pods running on that node doing all sorts of stuff. If that happens, maybe try to identify which pods stuck in _terminating_ and try to delete them by force. Then re-try draining that specific node.

As far as I know we still have a maximum of 150 pods/node. Did you encounter something like this in OpenShift? Would increasing the timeout to, let’s say, 10 minutes be “okay”? I can imagine that we drain the node and then run a `crictl rmp -fa` but usually I would expect that draining works somehow. In any case thanks for the concerns @drpaneas 🤗

drpaneas:
> As far as I know we still have a maximum of 150 pods/node. Did you encounter something like this in OpenShift? Would increasing the timeout to, let’s say, 10 minutes be “okay”? I can imagine that we drain the node and then run a `crictl rmp -fa` but usually I would expect that draining works somehow. In any case thanks for the concerns @drpaneas 🤗

Yes, I have seen this happening. It's not common, but it is still something you should consider it might happen :( I don't think increasing the timeout would help here. The `crictl` workaround should work though ;) 

- try to drain the node
- if it timeouts do the `crictl` workaround
- then try to drain the node again (it should work now)

I am not sure how you can reproduce this scenario though. @chentex one idea: maybe try running a pod that is using a volume mount on the host or something that interacts with the disk (e.g. take btrfs snapshots) so a simple smooth deletion won't suffice in this case.

Anyway, this is more of food for thought for another follow-up PR I guess.

Cheers @saschagrunert 🤗

saschagrunert:
That’s interesting, thanks for the insights @drpaneas. 😊 I think having some sort of failure recovery (crictl rmp) could be also done on a second execution of `skuba upgrade`. Then we check that the node is already drained and if workloads are still running we try to remove them. 

One issue I see is the general idempotence of skuba invocations… I have to think more about this. 🤔

jordimassaguerpla:
backpòrt https://github.com/SUSE/skuba/pull/1298/commits

PR #1278, opened 2020-07-23
updated to remove dep on 1.18 methods

Klaven:
it fixes 1.18 calls that where failing and falls back to 1.16 logic when needed.


internal/pkg/skuba/node/common.go
```diff
@@ -95,4 +87,4 @@ func documentMapToJoinConfiguration(gvkmap kubeadmapi.DocumentMap, allowDeprecat
 	}
 
 	return internalcfg, nil
-}
+}
```
Klaven, line 90: I have no idea how this happened. one second i'll fix it because github dislikes it.

dannysauer:
LGTM

PR #1290, opened 2020-07-30
Lock kured after disabling skuba-update.timer

mjura:
## Why is this PR needed?

This eliminates possible race condition with kured service

Fixes: https://github.com/SUSE/avant-garde/issues/1830

## What does this PR do?

It is locking kured service after skuba-update.timer is disabled

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kured/kured.go
+++ b/internal/pkg/skuba/kured/kured.go
@@ -19,6 +19,7 @@ package kured
 
 import (
 	"context"
+	"os"
 
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -29,6 +30,7 @@ import (
 
 const (
 	kuredDSName             = "kured"
+	kuredRebootSentinelFile = "/var/run/reboot-required"
 	kuredLockAnnotationJson = `{"metadata":{"annotations":{"weave.works/kured-node-lock":"{\"nodeID\":\"manual\"}"}}}`
 )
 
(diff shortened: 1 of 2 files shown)


pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -74,12 +74,27 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
 	}
 
+	// Disable skuba-update.timer before upgrade
+	if skubaUpdateWasEnabled {
+		err = target.Apply(nil, "skuba-update-timer.disable")
+		if err != nil {
+			return err
+		}
+	}
+
```
evrardjp, line 84: Note: there is _technically_ still a race here: Disabling skuba-update-timer will just ensure the service doesn't run. What if it ran before the `skuba node upgrade`, let a file in place where kured expects so. When skuba-update-timer gets disabled here, kured could possibly not have a lock yet at L86, and create one (because polling time) before L92 gets executed. That's because we don't check at L85 if a file marking a reboot already exists.

However, I think the proposed approach is currently the best:
- If we bail out at L85 by checking if a file marking the reboot is present, it would effectively prevent admins to upgrade their cluster at the same time a reboot is pending (therefore forcing a larger maintenance window). This could be overcomed by adding a flag to the `node upgrade` command (like `force`), but I have the impression it's over-engineering things.
- The likelihood of this happening is low.

If we really want to fix this, instead of overengineering, we could also (on top of this code) simply add a check if the reboot sentinel file exists, and just remove at the same time we are disabling the update timer at L79-84.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -75,12 +75,34 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
+  // Check if a kured reboot file already exists
+  if kured.RebootFileExists() {
```
evrardjp, line 87: indentation seems a little off, compared to the next paragraph L93+.

evrardjp: you should also record `kuredRebootFilePresent` here, for line 216.

mjura: done

internal/pkg/skuba/kured/kured.go
```diff
@@ -71,3 +73,24 @@ func Unlock(client clientset.Interface) error {
...
+		return false
+	}
+	return true
+}
+
+func RebootFileCreate() error {
+	_, err := os.Create(kuredRebootSentinelFile)
+	if err != nil {
+		return errors.Wrapf(err, "could not create file %q", kuredRebootSentinelFile)
+	}
```
evrardjp, line 88: should this return a nil or something like that?

PR #1293, opened 2020-07-30
Drop hyperkube image version > v1.17

mjura:
## Why is this PR needed?

This will improve check when exactly we have to drop hyperkube image during upgrade.

## What does this PR do?

It will check what is current Kubernetes version and what is latest Kubernetes version.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


mjura:
Please review again, I have updated `if` condition, which will avoid regression.

PR #1294, opened 2020-07-30
updated version to v4.5

Klaven:
Updated versions to v4.5
includes packages and ci

will not pass CI till all the other things are done.

please make sure I did not miss any place that needs updating! 

## Why is this PR needed?

our versions are wrong.

Does it fix an issue? addresses a business case?
corrects our version numbers, and most importantly fixes the where we pull images from.

add a description and link to the issue if one exists.

Refs SUSE/avant-garde#1862

**Reminder**: Add the "fixes bsc#XXXX" to the title of the commit so that it will
appear in the changelog.

## What does this PR do?

please include a brief "management" technical overview (details are in the code)

## Anything else a reviewer needs to know?

Special test cases, manual steps, links to resources or anything else that could be helpful to the reviewer.

## Info for QA

This is info for QA so that they can validate this. This is **mandatory** if this PR fixes a bug.
If this is a new feature, a good description in "What does this PR do" may be enough.

### Related info

SUSE/avant-garde#1862

### Status **BEFORE** applying the patch

How can we reproduce the issue? How can we see this issue? Please provide the steps and the prove
this issue is not fixed.

### Status **AFTER** applying the patch

How can we validate this issue is fixed? Please provide the steps and the prove this issue is fixed.

## Docs

If docs need to be updated, please add a link to a PR to https://github.com/SUSE/doc-caasp.
At the time of creating the issue, this PR can be work in progress (set its title to [WIP]),
but the documentation needs to be finalized before the PR can be merged. 

# Merge restrictions

(Please do not edit this)

We are in *v4-maintenance phase*, so we will restrict what can be merged to prevent unexpected surprises:

    What can be merged (merge criteria):
        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine
        there is a PR for updating documentation (or a statement that this is not needed)

PR #1300, opened 2020-08-03
WIP: Skip node reboot test

jordimassaguerpla:
This is to test what happens if I skip the node reboot when doing the 4.5 renaming

This is https://github.com/SUSE/skuba/pull/1294/commits but with removing the node reboot test.


jordimassaguerpla:
https://github.com/SUSE/skuba/pull/1294 it is green without it

PR #1324, opened 2020-08-13
fix kured version to match latest 4.2.2 release

jordimassaguerpla:
kured version was mismatching. Now it should match master and maintenance branch

fixes https://github.com/SUSE/avant-garde/issues/1901

https://github.com/SUSE/skuba/blob/master/internal/pkg/skuba/kubernetes/versions.go#L131
https://github.com/SUSE/skuba/blob/maintenance-caasp-v4/internal/pkg/skuba/kubernetes/versions.go#L105

### DIFF
--- a/internal/pkg/skuba/kubernetes/versions.go
+++ b/internal/pkg/skuba/kubernetes/versions.go
@@ -128,7 +128,7 @@ var (
 			},
 			AddonsVersion: AddonsVersion{
 				Cilium:        &AddonVersion{"1.6.6-rev5", 4},
-				Kured:         &AddonVersion{"1.3.0-rev4", 5},
+				Kured:         &AddonVersion{"1.3.0", 4},
 				Dex:           &AddonVersion{"2.16.0-rev6", 7},
 				Gangway:       &AddonVersion{"3.1.0-rev4", 6},
 				MetricsServer: &AddonVersion{"0.3.6", 1},


internal/pkg/skuba/kubernetes/versions.go
```diff
@@ -128,7 +128,7 @@ var (
 			},
 			AddonsVersion: AddonsVersion{
 				Cilium:        &AddonVersion{"1.6.6-rev5", 4},
-				Kured:         &AddonVersion{"1.3.0-rev4", 5},
+				Kured:         &AddonVersion{"1.3.0", 4},
```
ereslibre, line 131: Why the 5 -> 4 on the manifest version? This number should always get increased -- what also reminds me that we should bump these numbers on `master` as well, so any skuba build branched from master in any point in the future will always have at least the same (or greater) manifest version.

jordimassaguerpla: This is just synchronizing with the contents of 4.2.2. Note this is the section for k8s 1.17.4, not the 1.18.* which is caaspv5. Because this section is not in sync with was released for 4.2.2, migration is failing. So we need it to be the same as in 4.2.2.

ereslibre: Ah ok, thank you @jordimassaguerpla!

PR #1325, opened 2020-08-17
[WIP] update suggestions for add features in aws mutlti zone support

cclhsu:
update suggestions from code review for add features in aws mutlti zone support

Signed-off-by: cclhsu <clark.hsu@suse.com>

## Why is this PR needed?

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


ci/infra/aws/master-instance.tf
```diff
@@ -1,27 +1,58 @@
...
+  # etcd - internal
+  ingress {
+    from_port   = 2379
+    to_port     = 2380
+    protocol    = "tcp"
+    cidr_blocks = [var.cidr_block]
+    description = "etcd"
+  }
+
+  # api-server - everywhere
```
Klaven, line 23: Same here, Api server should be internal only as well.

ci/infra/aws/load-balancer.tf
```diff
@@ -1,19 +1,55 @@
...
+
+  # HTTPS access from anywhere
+  ingress {
+    from_port   = 443
+    to_port     = 443
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  ingress {
```
Klaven, line 31: @dannysauer Correct me if i'm wrong but I don't think we should expose the API server too the web. this is an easy first step to securing them, making it so that only inside the cluster can directly hit the api server. I would be for removing this block.

Klaven (commented):
Hey, Looked over the PR a little bit this morning, had a few quick questions regarding it.

1) could you fill out the PR template a bit more?  While i'm a bit fan of "the code should speak for itself" a lot of the time the code does not give a good "why" and that would be helpful!

2) the PR is quite large :D anything you could do to slim it down would make it easier to review

3) generally, I don't think the API server should be exposed externally unless someone absolutely needs it... and lets be honest, no one actually needs it. A large part of the security CVE's can be mitigated by keeping the control plane safe.

ci/infra/aws/load-balancer.tf
dannysauer: I'm not completely clear why we open 80/443 to begin with; nothing listens on those ports by default, right? Was that just a copy-paste from an example that was left behind all along?

dannysauer: Also, yes, exposing the api server seems undesirable. I guess it's possible to control access to the whole vpc and that might be an acceptable control, but I honestly haven't read the entire config to see if that's being done. It still seems like it'd be preferable to elect a source IP as "management node" or identify a management network.

dannysauer: Also, yes, exposing the api server seems undesirable. I guess it's possible to control access to the whole vpc and that might be an acceptable control, but I honestly haven't read the entire config to see if that's being done. It still seems like it'd be preferable to elect a source IP as "management node" or identify a management network.

cclhsu:
> Hey, Looked over the PR a little bit this morning, had a few quick questions regarding it.
> 
> 1. could you fill out the PR template a bit more?  While i'm a bit fan of "the code should speak for itself" a lot of the time the code does not give a good "why" and that would be helpful!
> 2. the PR is quite large :D anything you could do to slim it down would make it easier to review
> 3. generally, I don't think the API server should be exposed externally unless someone absolutely needs it... and lets be honest, no one actually needs it. A large part of the security CVE's can be mitigated by keeping the control plane safe.

Ok, I will slim PR down to focus just multiple zone. I am add too many formatting change and little customized improvemnt here and there.

ci/infra/aws/load-balancer.tf
cclhsu: This is from original code. I just try to rearrange code by function. I will test to remove 80/443 to see if any side effect

ci/infra/aws/master-instance.tf
cclhsu: `var.cidr_block` is internal.

Klaven (commented):
From my point of view this looks fine. I don't know enough about our AWS deployment to say though. @flavio would be a good reviewer.

dannysauer (commented):
I'm in the same boat as @Klaven ; this looks fine to me as a cleanup, but I don't have enough "AWS via Terraform" background to say it's good to go.

ci/infra/aws/output.tf
```diff
@@ -1,11 +1,55 @@
-output "control_plane_public_ip" {
-  value = "${zipmap(aws_instance.control_plane.*.id, aws_instance.control_plane.*.public_ip)}"
+output "username" {
```
c3y1huang, line 1: As discovered, this is currently no-op and instance automatically uses `ec2-user`.

c3y1huang (changes requested):

ci/infra/aws/output.tf
c3y1huang: Resolved in https://github.com/SUSE/skuba/pull/1325/commits/1317c3c94359df6664a37a5a81e01249ba8751f8

PR #1337, opened 2020-08-25
Init terraform before destroy

pablochacin:
## Why is this PR needed?

Fixes https://github.com/SUSE/avant-garde/issues/1484

## What does this PR do?

Call the terraform init before attempting a destroy to prevent terraform to fail because the working directory has not been initialized.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/testrunner/platforms/terraform.py
+++ b/ci/infra/testrunner/platforms/terraform.py
@@ -31,12 +31,24 @@ class Terraform(Platform):
                           self.tfjson_path]
 
     def destroy(self, variables=[]):
+        self._tf_init()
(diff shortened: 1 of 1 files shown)


pablochacin:
Running the destroy without provisioning generates the following error in VMware 
```
2020-08-25 17:26:52 ERROR] 
Error: error fetching datacenter: default datacenter resolves to multiple instances, please specify
```

davidcassany (approved):
I like this simple solution. Looks simpler that trying to keep failure status across steps and run conditionals during clean up steps. :+1:

PR #1343, opened 2020-08-27
Update versions of manifests to avoid overlaps

evrardjp:
## Why is this PR needed?

Fixes: https://github.com/SUSE/avant-garde/issues/1917

## What does this PR do?

Without this patch, overlaps in manifests versions can happen between
4.5 and 4.2 branch.

This is a problem, as identical versions which would prevent
any cluster addon upgrade. Additionally, it would be easy to get confused
what would the version 5 of a manifest means: is it an update on the 4.2.x
branch, or is it a 4.5.x branch change?

This fixes it by having real unique ranges which don't overlap.
The version used is just for convenience (4500 is easy to mentally
map to 4.5 release, due to hundreds numbers matching major/minor release
without dots).

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


davidcassany (approved):
Looks good to me :+1: Thanks @evrardjp

ereslibre (approved):

PR #1344, opened 2020-09-01
Bump etcd image tag

davidcassany:
## Why is this PR needed?

This PR is needed to make sure the new ectd image version is being used when upgrading from v4.2.3 to v4.5.0. Note that despite the etcd version is the same the image is different as the first is based on SLE-15-SP1 and the latter on SLE-15-SP2.

Fixes SUSE/avant-garde#1924

## What does this PR do?

This commit appends the revision to the etcd image reference. This is
done to force kubeadm to write a new manifest including the new image
reference when upgrading from v4.2.3 (SLE15-SP1 based) to v4.5.0
(SLE15-SP2 based) CaaSP version.

## Info for QA

Fixes SUSE/avant-garde#1924, after the migration from v4.2.3 to v4.5.0 the upgraded clusters should be using the etcd image tagged with `3.4.3-rev1`.

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubernetes/versions.go
+++ b/internal/pkg/skuba/kubernetes/versions.go
@@ -96,7 +96,7 @@ var (
 				ControllerManager: &ContainerImageTag{Name: "kube-controller-manager", Tag: "v1.18.6"},
 				Scheduler:         &ContainerImageTag{Name: "kube-scheduler", Tag: "v1.18.6"},
 				Proxy:             &ContainerImageTag{Name: "kube-proxy", Tag: "v1.18.6"},
-				Etcd:              &ContainerImageTag{Name: "etcd", Tag: "3.4.3"},
+				Etcd:              &ContainerImageTag{Name: "etcd", Tag: "3.4.3-rev1"},
 				CoreDNS:           &ContainerImageTag{Name: "coredns", Tag: "1.6.7"},
 				Pause:             &ContainerImageTag{Name: "pause", Tag: "3.2"},
 				Tooling:           &ContainerImageTag{Name: "skuba-tooling", Tag: "0.1.0"},


evrardjp:
I am not sure how -rev1 parses vs 3.4.3.0

jenting (approved):

davidcassany:
> I am not sure how -rev1 parses vs 3.4.3.0

This is a valid concern, in fact according to https://semver.org the relationship across those two tags is `3.4.3-rev1 < 3.4.3`. However this is unclear to me if this is the actual comparison that is happening behind the scenes.

evrardjp:
Exactly, this is why i was concerned.

evrardjp:
i double checked with a quick PoC: https://github.com/evrardjp/poc-kubernetes-version-parse

evrardjp:
Can we do 3.4.4?

davidcassany:
> Can we do 3.4.4?

I don't thing so, this would be misleading if the actual version is 3.4.3. @evrardjp do you think `3.4.3` vs `3.4.3.1` would do the trick? Could you verify if something like that crashes the comparison code (probably it does, this is no longer semver).

evrardjp (changes requested):
That won't work (see PoC). We should abandon this.

evrardjp:
3.4.3.1 gets errors in parsing, so I would say no.

davidcassany:
Closing, this not a good solution. Probably we should rethink the three tags strategy, probably should have been `+revX` instead of `-revX`. In any case we are not considering chaning the tagging strategy now.

PR #1346, opened 2020-09-07
[CI] Implement pipeline for testing 4.2 to 4.5 upgrades

pablochacin:
## Why is this PR needed?

Automate testing  the upgrade from 4.2 to 4.5 to prevent regressions when maintaining any of the branches.

Fixes: https://github.com/SUSE/avant-garde/issues/1836

## What does this PR do?

Implement a process with two jobs: one that creates a temporary worker and triggers the upgrade test on these cluster.

Test executions:
* Trigger job: https://ci.suse.de/job/caasp-jobs/job/ci-dev/job/pablo/job/trigger-upgrade-from-4-2-daily/
* Upgrade test job: https://ci.suse.de/job/caasp-jobs/job/ci-dev/job/pablo/job/upgrade-from-4-2/

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/jenkins/jobs.yaml
+++ b/ci/jenkins/jobs.yaml
@@ -60,6 +60,7 @@
 
 - project:
     name: caasp-jobs/e2e/v4.5
+    branch: master
     repo-name: skuba
     repo-owner: SUSE
     repo-credentials: github-token
@@ -81,6 +82,8 @@
     jobs:
         - '{name}/{platform}/{test}-daily'
         - '{name}/{platform}/update-daily'
+        - '{name}/{platform}/trigger-upgrade-from-4-2-daily'
+        - '{name}/{platform}/upgrade-from-4-2-daily'
 
 - project:
     name: caasp-jobs/dev
--- a/ci/jenkins/pipelines/skuba-trigger-upgrade-from-4.2.Jenkisfile
+++ b/ci/jenkins/pipelines/skuba-trigger-upgrade-from-4.2.Jenkisfile
@@ -1,0 +1,98 @@
+def worker_name
+def worker_host
+def keep_worker = false
+
+pipeline {
+    agent { node {label 'caasp-team-private-experimental && openstack' }}
+
+    environment {
+        OPENRC = credentials('ecp-openrc')
+        JENKINS_ID = credentials('jenkins-ssh')
+        CAASP_CI   = credentials('ci.suse.de.caasp-user')
+    }
+    
+    stages {
+        stage('cleanup'){
+            steps {
+                dir("${WORKSPACE}/credentials") {
+                    deleteDir()
+                }
+            }
+        }
+        stage('Setup') {
+            steps {
+                // Get automation code from a GitHub repository
+                git url: 'https://github.com/SUSE/caasp-automation.git', credentialsId: 'github-token'
+
+                // setup credentials
+                dir('jenkins-worker'){
+                    sh 'mkdir -p credentials >> /dev/null || true'
+                    sh "cp ${env.JENKINS_ID} credentials/jenkins"
+                    sh "cp ${env.OPENRC} credentials/ecp.openrc"
+                    sh "cp ${env.CAASP_CI} credentials/ci.suse.de.caasp-user"
+                }
+            }
+        }
+        
+        stage('Create-worker') {
+            steps { script {
+                worker_name = "caasp-team-${BUILD_NUMBER}-dynamic-worker"
+                dir('jenkins-worker'){
+                    sh "./manage-workers.sh -c --worker-name ${worker_name} --worker-type dynamic --worker-template ${WORKER_TEMPLATE} --labels ${worker_name} --ssh-key ${env.JENKINS_ID}"
+                    worker_host  = sh(script: "./manage-workers.sh -i --worker-name ${worker_name}", returnStdout: true).trim()
+                    echo "woker ip ${worker_host}"
+                }
+            }}
+        }
+        stage('run-test'){
+            steps { script {
+                build   job: env.DOWNSTREAM_JOB, 
+                        wait: true, 
+                        propagate: true,
+                        parameters: [
+                            string(name: 'WORKER_LABEL', value: "${worker_name}"),
+                            string(name: 'WORKER_HOST', value: "${worker_host}"),
+                            string(name: 'BRANCH', value: env.BRANCH),
+                            string(name: 'RETAIN_CLUSTER', value: env.RETAIN_CLUSTER),
+                            string(name: 'RETENTION_PERIOD', value: env.RETENTION_PERIOD),
+                            string(name: 'INITIAL_VERSION', value: env.INITIAL_VERSION),
+                            string(name: 'TARGET_VERSION', value: env.TARGET_VERSION)
+                        ]
+            }}    
+        }
+    }
+    post {
+        always { script {
+            if (Boolean.parseBoolean(env.RETAIN_WORKER)) {
+                def retention_period= env.RETENTION_PERIOD?env.RETENTION_PERIOD:24
+                try{
+                    timeout(time: retention_period, unit: 'HOURS'){
+                        input(message: 'Waiting '+retention_period +' hours before cleaning up worker \n. ' +
+                                        'Press <abort> to cleanup inmediately, <keep> for keeping it',
+                                ok: 'keep')
+                        keep_worker = true
+                    }
+                }catch (err){
+                    // either timeout occurred or <abort> was selected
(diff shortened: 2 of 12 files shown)


ci/jenkins/jobs.yaml
```diff
@@ -81,6 +82,8 @@
     jobs:
         - '{name}/{platform}/{test}-daily'
         - '{name}/{platform}/update-daily'
+        - '{name}/{platform}/trigger-upgrade-from-4-2-daily'
+        - '{name}/{platform}/upgrade-from-4-2-daily'
```
davidcassany, line 86: I am not sure I understand this differentiation between `trigger-upgrade` and `upgrade`. One is creating the worker and the other one using the created worker to actually run the upgrade test? I find it a bit confusing. Is this done to avoid having a very complex node setup (a setup that includes creating the node)?

ci/jenkins/pipelines/skuba-upgrade-from-4.2.Jenkinsfile
```diff
@@ -0,0 +1,45 @@
...
+    
+    stages {
+        stage('upgrade') {
+            agent { label "${WORKER_LABEL}" }
+            
+            environment {
+                REG_CODE = credentials('scc-regcode-caasp')
+            }
+            steps {
+                echo 'upgrading'
```
davidcassany, line 13: I assume we should provision and deploy a cluster here and then run the upgrade steps. Am I correct?

ci/jenkins/jobs.yaml
pablochacin: Yes,it is to avoid a massive complex job that takes care of everything. I think it is better to split the process int two. Each job has a well-defined purpose. Besides, the upgrade can be tested in isolation.

ci/jenkins/pipelines/skuba-upgrade-from-4.2.Jenkinsfile
pablochacin: Actually there are additional steps: build the skuba, provision the cluster, deploy the cluster (v4.2). Those steps are part of a previous stage

davidcassany (commented):
I have doubts about the schedule of the `trigger-upgrade-from-4-2-daily` job in relation with the `upgrade-from-4-2`. To me it looks like both are scheduled, however I had the impression that the  `trigger-upgrade...` job already calls the  `upgrade-from-4-2` job, thus I don't see the need of scheduling the later. Am I missing something?

ci/jenkins/pipelines/skuba-trigger-upgrade-from-4.2.Jenkisfile
```diff
@@ -0,0 +1,102 @@
...
+                }catch (err){
+                    // either timeout occurred or <abort> was selected
+                    keep_worker = false
+                }
+            }
+        }}
+        cleanup { script {
+            if (!keep_worker){
+                dir('jenkins-worker'){
+                    sh "./manage-workers.sh -d --worker-name ${worker_name} || true"
```
Itxaka, line 88: should we wrap this on a retry in case it fails so we can be sure that we tried to clean up as much as possible?

https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#retry-retry-the-body-up-to-n-times

ci/jenkins/pipelines/skuba-upgrade-from-4.2.Jenkinsfile
```diff
@@ -0,0 +1,133 @@
...
+        GITHUB_TOKEN = credentials('github-token')
+        VMWARE_ENV_FILE = credentials('vmware-env')
+        TERRAFORM_STACK_NAME = "${BUILD_NUMBER}-${JOB_NAME.replaceAll("/","-")}".take(70)
+     }
+
+     stages {
+
+        stage("Build skuba"){
+            steps{
+                sh(script: "git checkout ${INITIAL_VERSION}")
```
Itxaka, line 19: yikes, isnt there a better way to change branches? I heard using git directly was not recommended although I cant remember why.

ci/jenkins/pipelines/skuba-upgrade-from-4.2.Jenkinsfile
```diff
@@ -0,0 +1,133 @@
...
+            steps {
+                sh(script: "sudo zypper in -y --no-recommends go1.13")
+                sh(script: "git checkout -f ${TARGET_VERSION}")
+                sh(script: "make install", label: "Build ${TARGET_VERSION}")
+            }   
+        }
+
+        stage('upgrade cluster'){
+            environment {
+                REG_CODE = credentials('scc-regcode-caasp')
```
Itxaka, line 79: cant we set the REG_CODE on the top level environment?

ci/jenkins/pipelines/skuba-upgrade-from-4.2.Jenkinsfile
```diff
@@ -0,0 +1,133 @@
...
+                
+            }
+            steps {
+                sh(script: "make -f ci/Makefile test SUITE='test_upgrade_from_4_2' SKIP_SETUP='deployed'", label: "Test upgrade")
```
Itxaka, line 83: should we make the test suite a param of the job?

ci/jenkins/templates/trigger-upgrade-from-4-2-daily.yaml
```diff
@@ -0,0 +1,60 @@
...
+            description: The platform to perform the tests on
+        - string:
+            name: DOWNSTREAM_JOB
+            description: Job to be triggered downstream
+            default: '{name}/{platform}/upgrade-from-4-2-daily'  
+        - string:
+            name: WORKER_TEMPLATE
+            description: Template for worker
+            default: 'maintenance-caasp-v4.2'
+        - string:
```
Itxaka, line 26: is there any case in which we would need to add labels? As the worker is created just for the downstream job...seems like there is no need to add extra labels to it.

Itxaka (approved):
All commments are comments/nits/question, feel free to ignore them as they are not blockers, nice work!

ci/jenkins/pipelines/skuba-upgrade-from-4.2.Jenkinsfile
pablochacin: we can review this in a follow up.

ci/jenkins/pipelines/skuba-upgrade-from-4.2.Jenkinsfile
pablochacin: We don't have to. It is only required here.

ci/jenkins/pipelines/skuba-upgrade-from-4.2.Jenkinsfile
pablochacin: We could, however, this job is specific for that test so I don't see any advantage. Moreover, I see some potential issues because the test is the actual upgrade process. Please, clarify which use cases you see it would be necessary to have this as a parameter.

ci/jenkins/templates/trigger-upgrade-from-4-2-daily.yaml
pablochacin: Maybe not.

ci/jenkins/pipelines/skuba-upgrade-from-4.2.Jenkinsfile
Itxaka: yeah, I was thinking when moving this into a shared library, having the test suite as a parameter so we could reuse it.

Butr it doesnt make much sense in this case, or now.

pablochacin:
> I have doubts about the schedule of the `trigger-upgrade-from-4-2-daily` job in relation with the `upgrade-from-4-2`. To me it looks like both are scheduled, however I had the impression that the `trigger-upgrade...` job already calls the `upgrade-from-4-2` job, thus I don't see the need of scheduling the later. Am I missing something?

Good catch, only the trigger should be scheduled. I'll fix that.

ci/jenkins/pipelines/skuba-trigger-upgrade-from-4.2.Jenkisfile
pablochacin: Makes sense, but we should then try to figure out what the failure was in order to retry. I would prefer to defer this from when the logic moved to the shared library.

Itxaka (approved):

davidcassany (approved):
LGTM

PR #1350, opened 2020-09-09
[TF] Handle resource name errors

Itxaka:
There seems to be an issue when handling the output in openstack
in which the resource name is not available which results into
the output file failing to write due to zipmap operating into an
empty tuple

This patch fixes that by recovering from an empty resource name and
generating the same names based on the length of the addresses list

This should be safe to use but the real problem is somewhere else if
terraform is not able to provide the name, it could also not provide
the addresses for the resources, but the bug report seems to indicate
that only the names are not available while the addresses are there
so it should cover that failure and have a safe fallback

Fixes #1171653
Fixes https://github.com/SUSE/avant-garde/issues/1469

## Why is this PR needed?

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/ci/infra/openstack/output.tf
+++ b/ci/infra/openstack/output.tf
@@ -1,9 +1,19 @@
 output "ip_masters" {
-  value = "${zipmap(openstack_compute_instance_v2.master.*.name, openstack_networking_floatingip_v2.master_ext.*.address)}"
+  value = zipmap(
+    concat(
+      openstack_compute_instance_v2.master.*.name, [ for i in range(length(openstack_compute_instance_v2.master), var.masters) : format("master-%d (not provisioned)", i ) ]),
+    concat(
+      openstack_networking_floatingip_v2.master_ext.*.address,  [ for _ in range(length(openstack_networking_floatingip_v2.master_ext.*.address), var.masters) : "" ])
+    )
 }
 
 output "ip_workers" {
-  value = "${zipmap(openstack_compute_instance_v2.worker.*.name, openstack_networking_floatingip_v2.worker_ext.*.address)}"
+  value = zipmap(
+    concat(
+      openstack_compute_instance_v2.worker.*.name, [ for i in range(length(openstack_compute_instance_v2.master), var.workers) : format("worker-%d (not provisioned)", i ) ]),
+    concat(
+      openstack_networking_floatingip_v2.worker_ext.*.address,  [ for _ in range(length(openstack_networking_floatingip_v2.worker_ext.*.address), var.workers) : "" ])
+  )
 }
 
 output "ip_internal_load_balancer" {
@@ -11,5 +21,9 @@ output "ip_internal_load_balancer" {
 }
 
 output "ip_load_balancer" {
-  value = "${zipmap(list(openstack_lb_loadbalancer_v2.lb.name), list(openstack_networking_floatingip_v2.lb_ext.address))}"
+  value = zipmap(
+    length(openstack_lb_loadbalancer_v2.lb.*) == 1 ? list(openstack_lb_loadbalancer_v2.lb.name) : list("${var.stack_name}-lb"),
+    length(openstack_networking_floatingip_v2.lb_ext.*) == 1 ? list(openstack_networking_floatingip_v2.lb_ext.address) :
+      list("Could not obtain ip address, please retry terraform refresh to update the output")
+  )
 }


pablochacin:
> There seems to be an issue when handling the output in openstack
> in which the resource name is not available which results into
> the output file failing to write due to zipmap operating into an
> empty tuple
I think the problem could potentially also happen in other platforms because of the way zipmap function works. Maybe the difference is that other providers fail in a different way

> This patch fixes that by recovering from an empty resource name and
> generating the same names based on the length of the addresses list
> 
> This should be safe to use but the real problem is somewhere else if
> terraform is not able to provide the name, it could also not provide
> the addresses for the resources, but the bug report seems to indicate
> that only the names are not available while the addresses are there
> so it should cover that failure and have a safe fallback

I think the root cause is that even when some resources were not allocated (generally the problem was with the floating ip addresses), terraform tries to generate the output.

ci/infra/openstack/output.tf
```diff
@@ -1,15 +1,23 @@
 output "ip_masters" {
-  value = "${zipmap(openstack_compute_instance_v2.master.*.name, openstack_networking_floatingip_v2.master_ext.*.address)}"
+  value = zipmap(
+    length(openstack_compute_instance_v2.master.*) > 0 ? openstack_compute_instance_v2.worker.*.name :
+      [for i,v in openstack_networking_floatingip_v2.master_ext.*.address: format("caasp-master-${var.stack_name}-%d", i)],
+    openstack_networking_floatingip_v2.master_ext.*.address)
 }
```
pablochacin, line 6: If I understand it correctly, the patch assumes the problem is that the list of masters is empty. But I think a more general problem is that zipmap assumes both lists are the same length. It can still happen (and IIRC it actually happened) that the list of IP addresses didn't match the list of master (some ip address could not allocatted?)  Maybe checking that both lists have the same length and the length is not 0 would be a more robust check?

Itxaka: umm, that is right, but the only occurrence I have seen of this is that the list of names is empty due to whatever happened behind the scenes, so Im assuming that terraform is unable to obtain the names of ALL the resources, not just one.

Do you think it could happen that terraform is able to obtain some of the names, i.e. a partial list of names?

In that case, the complexity and ugliness of this just increased tenfold, if we cant asume that it either fails to obtain everything :(

Itxaka: I mean, I can write a equal check in there, but if either names or addresses are missing, what is the expected output? Do we just NOT write the output? Do we write an empty output and let the user figure it out?

This is the problem with incomplete reports, there is not much to understand from a small subset of logs, having the full logs of that failure will help us understand what was really missing and if terraform actually failed to allocate the machine (I thought a failure to allocate a resource means no output :O). That or at least a way to reproduce it :)

pablochacin: We don't have logs and we cannot solve that issue. Regarding the "incompleteness" of the report, I think that this doesn't change the rationale of my comment: the root cause of the problem is that both lists should match. Checking for a) lengths match AND b) length> 0 should address any other possible case, including the only case reported.

Itxaka:
As far as I can see there is several scenarios in which we need to fix or add safe features around:

 - ALL resource names are not obtained
 - ALL resources are not created/on error state
 - SOME resource names are not obtained
 - SOME resources are not created/on error state
 - SAME for ipaddresses resources

This means:
 - We cannot just compare one list with the other as there can be a mismatch on either side
 - We cannot check for a resource list to be > 0 to confirm that is working
 - We cannot just generate the names out of the blue based on the ipaddress list, as those resources could be on error state or not created (instances and ipaddresses are 2 different resources)

To be honest I would have expected terraform to raise an error if a resource name could not be obtained or the resource creation failed but it seems like in this case it didnt.

In conclusion, we probably should try to obtain the number of resources that we set to be created, compare them to the list and then raise an error if its a mismatch. If its a temporal problem of obtaining the names only, terraform apply can be launched again and it will retry, otherwise the burden of checking the infra is on the user.

Itxaka:
PR updated with what I think is the most optimal solution we can reach currently

pablochacin:
@Itxaka I think we have a communication problem around this issue. The intention is not to fix the TF problem that caused the mismatch in the resource lists, but simply to prevent the output to fail obscuring the previous errors that terraform may have issued. 

Regarding the fix

> * We cannot just compare one list with the other as there can be a mismatch on either side

If the lists have the same length, the `zipmap` will work and the output will be generated. 

> * We cannot check for a resource list to be > 0 to confirm that is working

We don't need to check if it is working. This is not the purpose of the issue. We only must ensure the output doesn't fail. 

> * We cannot just generate the names out of the blue based on the ipaddress list, as those resources could be on error state or not created (instances and ipaddresses are 2 different resources)

I never proposed to do this and I'm particularly concerned that your fix is actually generating such names. 


I think that  this would satisfy the following code fix the issue with the output by simply padding each list. 
```
value = zipmap(
    concat(openstack_compute_instance_v2.master.*.name, [ for i in range(length(openstack_compute_instance_v2.master), var.masters) : format("master-%d (not provisioned)", i ) ]) : concat(openstack_networking_floatingip_v2.master_ext.*.address,  [ for _ in range(length(openstack_networking_floatingip_v2.master_ext.*.address), var.masters)): "" ])
  )
```

Itxaka:
I just want to get this over with :D

pablochacin:
> I just want to get this over with :D

Fair enough. I'm most interested in solving the issue. If you agree we can close this PR. I can take it from here.

Itxaka:
> Fair enough. I'm most interested in solving the issue. If you agree we can close this PR. I can take it from here.

I mean that I dont want to spend too much energy here, as it doesnt make sense. I updated the PR to use your approach, as long as we can merge this and get it out of the way, I dont care at all.

pablochacin:
This is a recent [output](https://ci.suse.de/job/caasp-jobs/job/dev/job/e2e-test/160/console) with the issue:
```
2020-09-21 08:01:27 ERROR] Error: Error in function call

2020-09-21 08:01:27 ERROR] 
  on output.tf line 2, in output "ip_masters":
2020-09-21 08:01:27 ERROR]   on output.tf line 2, in output "ip_masters":
2:   value = "${zipmap(openstack_compute_instance_v2.master.*.name, openstack_networking_floatingip_v2.master_ext.*.address)}"
2020-09-21 08:01:27 ERROR] 2:   value = "${zipmap(openstack_compute_instance_v2.master.*.name, openstack_networking_floatingip_v2.master_ext.*.address)}"
    |----------------
2020-09-21 08:01:27 ERROR]     |----------------
    | openstack_compute_instance_v2.master is tuple with 3 elements
2020-09-21 08:01:27 ERROR]     | openstack_compute_instance_v2.master is tuple with 3 elements
    | openstack_networking_floatingip_v2.master_ext is empty tuple
2020-09-21 08:01:27 ERROR]     | openstack_networking_floatingip_v2.master_ext is empty tuple
```

The root cause is that floating ip addresses could not be allocated:
```
2020-09-21 08:00:23 ERROR] Error: Error creating openstack_networking_floatingip_v2: Expected HTTP response code [] when accessing [POST https://engcloud.prv.suse.net:9696/v2.0/floatingips], but got 409 instead
{"NeutronError": {"message": "Quota exceeded for resources: ['floatingip'].", "type": "OverQuota", "detail": ""}}
```

pablochacin:
Running the e2e job with this PR's branch [show the same floating ip ](https://ci.suse.de/job/caasp-jobs/job/dev/job/e2e-test/161/console)allocation described above but the output error does not occurred, which seems to confirm the changes are effective.

pablochacin (commented):
LGTM

pablochacin (approved):

davidcassany (approved):
Looks good :+1:

evrardjp (commented):
As this will change the output, we should change the documentation (IIRC, that's our user interface, and it's documented). Can you confirm?

Else it looks good at first sight, but I didn't test/run this.

Itxaka:
> As this will change the output, we should change the documentation (IIRC, that's our user interface, and it's documented). Can you confirm?

> Else it looks good at first sight, but I didn't test/run this.


It *should* not change the output, at least not in a meaningful way. When this kicks in it means that there has been an issue acquiring the resources, so basically the tfout values are...trash.

The only reason for this is because instead of failing properly, either terraform or the openstack provider are stupid and they are trying to parse incomplete resources into the output template. IMO this should immediately fail and not trying to parse probably non-existant resource into a template ¬_¬

PR #1351, opened 2020-09-10
Fix vbom url

Itxaka:
Looks like the vanity url for the vbom.ml package has been dropped
and the new one its just pointing to github according to [0][1]

The vbom dependency comes from kubernetes, they already fixed it
upstream on version 1.19.1[2] so until we update our dependency
to match that, we need this as a workaround.

This workaround replaces the old url with the new one

[0] fvbommel/util#5
[1] fvbommel/util#6
[2] kubernetes/kubernetes@ad3891d

## Why is this PR needed?

Fixes https://github.com/SUSE/avant-garde/issues/1941

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -43,4 +43,6 @@ replace (
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.18.6
 )
 
+replace vbom.ml/util => github.com/fvbommel/util v0.0.0-20160121211510-db5cfe13f5cc
+
 go 1.13


Itxaka:
Im not even sure if this is correct, please check!

Itxaka:
Manually tested this with a new environment with no modules cached.

 - Without this patch it will fail to find the modules
 - With this patch it will correctly get the modules

evrardjp:
Would it make sense to fix one of our upstream dependencies which is consuming that module? Is it tracked in another card while we implement this workaround?

Itxaka:
I think this is one of the kubernetes dependencies? https://github.com/kubernetes/kubernetes/commit/ad3891df1badb92034371ce02fc929652c9f4263

Which means we cannot just update our dependencies as it requires a newer kubernetes version...1.19.1 at least I think? Im not clear on that to be fair, but that change has not been backported to other kubernetes branches and we are requiring 1.18.6... seems like a lot of work.

evrardjp:
Seems like it's only used in kubectl, and updated upstream already: https://github.com/kubernetes/kubectl/commit/ac345319b3f7acf35f4c3978257f1ecdb4aa7165#diff-37aff102a57d3d7b797f152915a6dc16

evrardjp:
I would just mention as an extra message in the commit message that this is a temporary fix until upstream releases a new version of the kubectl module for 0.18.x that includes that fix

Itxaka:
> I would just mention as an extra message in the commit message that this is a temporary fix until upstream releases a new version of the kubectl module for 0.18.x that includes that fix

Also on kubernetes which we require an older version. Im not sure we are bumping kubernetes versions anymore @evrardjp  :D

evrardjp:
Well I am new to go, so it's indeed weird to see in go mod graph that kubernetes indeeds requires kubectl v0.0.0
Not sure what's going on there, or what it means. But I feel like this is a problem to solve upstream, which is why I would highlight the fact (in the commit message) that it's a workaround until we get it fixed upstream

evrardjp (approved):
LGTM, I would just update the commit message

Itxaka:
updated!

evrardjp (approved):
Perfect!

evrardjp:
Future me thank you already for the change of the commit message.

Klaven (approved):
/lgtm

pablochacin:
This need to be backported to maintenance-caasp-v2 where the issue was detected.

PR #1352, opened 2020-09-10
fix: Ensure the target node OS for node upgrades is on SP2

atighineanu:
## Why is this PR needed?
Without this patch, user might try to upgrade k8s 1.17 to 1.18 on SLES 15SP1. 
This will eventually fail "will say packages not found", but there will be no clear informing why.

Related issue:
https://github.com/SUSE/avant-garde/issues/1667

Fixes #

appear in the changelog.

## What does this PR do?
- triggers a check during skuba node upgrade apply if OS corresponds to the "latest" kubernetes version or not
- informs the user if an OS upgrade is required
- stops the upgrade apply process 

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/deployments/os.go
+++ b/internal/pkg/skuba/deployments/os.go
@@ -29,7 +29,7 @@ const (
 	SUSEOSID = "suse"
 )
 
-func (t *Target) OSRelease() (map[string]string, error) {
+func (t *Target) osRelease() (map[string]string, error) {
 	if len(t.Cache.OsRelease) > 0 {
 		return t.Cache.OsRelease, nil
 	}
@@ -50,7 +50,7 @@ func (t *Target) OSRelease() (map[string]string, error) {
 }
 
 func (t *Target) hasOS(os string) (bool, error) {
-	osRelease, err := t.OSRelease()
+	osRelease, err := t.osRelease()
 	if err != nil {
 		return false, errors.Wrap(err, "could not retrieve OS release information")
 	}
--- a/internal/pkg/skuba/deployments/os.go
+++ b/internal/pkg/skuba/deployments/os.go
@@ -29,7 +29,7 @@ const (
 	SUSEOSID = "suse"
 )
 
-func (t *Target) osRelease() (map[string]string, error) {
+func (t *Target) OSRelease() (map[string]string, error) {
 	if len(t.Cache.OsRelease) > 0 {
 		return t.Cache.OsRelease, nil
 	}
@@ -50,7 +50,7 @@ func (t *Target) osRelease() (map[string]string, error) {
 }
 
 func (t *Target) hasOS(os string) (bool, error) {
-	osRelease, err := t.osRelease()
+	osRelease, err := t.OSRelease()
 	if err != nil {
 		return false, errors.Wrap(err, "could not retrieve OS release information")
 	}
--- a/internal/pkg/skuba/deployments/ssh/ssh.go
+++ b/internal/pkg/skuba/deployments/ssh/ssh.go
@@ -122,7 +122,6 @@ func (t *Target) GetDeployment(nodename string, role *deployments.Role, verboseL
 		port:         t.port,
 		verboseLevel: verboseLevel,
 	}
-	res.OSRelease()
 	return &res
 }
 
--- a/internal/pkg/skuba/deployments/ssh/ssh.go
+++ b/internal/pkg/skuba/deployments/ssh/ssh.go
@@ -122,6 +122,7 @@ func (t *Target) GetDeployment(nodename string, role *deployments.Role, verboseL
 		port:         t.port,
 		verboseLevel: verboseLevel,
 	}
+	res.OSRelease()
 	return &res
 }
 
--- a/pkg/skuba/actions/node/upgrade/apply.go
+++ b/pkg/skuba/actions/node/upgrade/apply.go
@@ -48,9 +48,7 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 	}
 	currentVersion := currentClusterVersion.String()
 	latestVersion := kubernetes.LatestVersion().String()
-	if strings.Contains(target.Cache.OsRelease["VERSION_ID"], "15.1") && !strings.Contains(latestVersion, "1.17") {
-		fmt.Printf("Node OS migration/upgrade required first.\n\n")
-	}
+
 	nodeVersionInfoUpdate, err := upgradenode.UpdateStatus(client, target.Nodename)
 	if err != nil {
 		return err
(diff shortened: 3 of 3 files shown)


internal/pkg/skuba/deployments/os.go
```diff
@@ -29,7 +29,7 @@ const (
 	SUSEOSID = "suse"
 )
 
-func (t *Target) osRelease() (map[string]string, error) {
+func (t *Target) OSRelease() (map[string]string, error) {
```
evrardjp, line 32: let's not change the interface if it's not necessary. Can you remove this part of your commit?

internal/pkg/skuba/deployments/os.go
```diff
@@ -50,7 +50,7 @@ func (t *Target) osRelease() (map[string]string, error) {
 }
 
 func (t *Target) hasOS(os string) (bool, error) {
-	osRelease, err := t.osRelease()
+	osRelease, err := t.OSRelease()
```
evrardjp, line 53: ditto.

internal/pkg/skuba/deployments/ssh/ssh.go
```diff
@@ -122,6 +122,7 @@ func (t *Target) GetDeployment(nodename string, role *deployments.Role, verboseL
 		port:         t.port,
 		verboseLevel: verboseLevel,
 	}
+	res.OSRelease()
```
evrardjp, line 125: Why would you include that in GetDeployment? Is there a reason why not doing it immediately in the target call that needs it? Like the upgrade apply/plan?

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -48,6 +48,9 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 	}
 	currentVersion := currentClusterVersion.String()
 	latestVersion := kubernetes.LatestVersion().String()
+	if strings.Contains(target.Cache.OsRelease["VERSION_ID"], "15.1") && !strings.Contains(latestVersion, "1.17") {
+		fmt.Printf("Node OS migration/upgrade required first.\n\n")
```
evrardjp, line 52: shouldn't we exit by issuing a Fatal instead?

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -48,6 +48,9 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 	}
 	currentVersion := currentClusterVersion.String()
 	latestVersion := kubernetes.LatestVersion().String()
+	if strings.Contains(target.Cache.OsRelease["VERSION_ID"], "15.1") && !strings.Contains(latestVersion, "1.17") {
```
evrardjp, line 51: As this code will most likely not be backported to 4.2 branch, there is no point of the last part of the conditional, which should always be true.

For the first part of the conditional, I would have written a more strict equality (version = 15.1, rather than contains 15.1). However, this is not necessary. You don't need to change anything, as I doubt we'll get to a point we'll have 15.10.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -48,6 +48,9 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
+		fmt.Printf("Node OS migration/upgrade required first.\n\n")
+	}
```
evrardjp, line 53: should we include a similar kind of test (but not having a fatal output) for the upgrade plan?

evrardjp (changes requested):
A few changes are required before we merge this.

internal/pkg/skuba/deployments/os.go
atighineanu: True. Done.

pkg/skuba/actions/node/upgrade/apply.go
atighineanu: Fixed.

pkg/skuba/actions/node/upgrade/apply.go
atighineanu: Fixed.

pkg/skuba/actions/node/upgrade/apply.go
atighineanu: yes, we definitely should.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -57,6 +58,11 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 	fmt.Printf("Latest Kubernetes version: %s\n", latestVersion)
 	fmt.Printf("Current Node version: %s\n", nodeVersionInfoUpdate.Current.KubeletVersion.String())
 	fmt.Println()
+	target.IsSUSEOS()
```
evrardjp, line 61: I would add a comment above L61 to say:
```
// Populate os release version in target Cache to execute OS checks
```

It's not 100% clear the link between 61 and 62.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -57,6 +58,11 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
+	if strings.Contains(target.Cache.OsRelease["VERSION_ID"], "15.1") {
```
evrardjp, line 62: Nit: I would probably use a constant for 15.1, so that the comparison is clearer. Not 100% necessary.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -57,6 +58,11 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
+		fmt.Println("You are still running 15.1, and the upgrade to this branch expects running on SP2,\n please upgrade to the latest skuba for SP1, and migrate to SP2 before doing this.\n")
+		os.Exit(1)
```
evrardjp, line 64: instead of os.Exit(1) and fmt.Println, you can simply log.Fatalf("")

evrardjp (changes requested):
A few last improvements

pkg/skuba/actions/node/upgrade/apply.go
atighineanu: That's a good idea. Thanks.

pkg/skuba/actions/node/upgrade/apply.go
atighineanu: I did that. And the output was just uglier.

pkg/skuba/actions/node/upgrade/apply.go
evrardjp: So guess what! Change of plans! The node info is gathered by kubelet, so we can assume it's always up to date.

If you can just compare the node info to SP1, that would be good.

``nodeVersionInfoUpdate.Current.Node.Status.NodeInfo.OSImage`` contains the string ``SUSE Linux Enterprise Server 15 SP1`` for example.

evrardjp:
That check can also be moved to the NodeUpgradeableCheck function

evrardjp:
@atighineanu the CI complains about pkg/skuba/actions/node/upgrade/apply.go:63:3: Println arg list ends with redundant newline
 . Can you update your code with my latest comments and fix this, please?

pkg/skuba/actions/node/upgrade/apply.go
evrardjp: Well, that was wrong. I think we need to load it live, as the deployer might not have rebooted after SP2 migration yet.

evrardjp:
@atighineanu I missed the fact that this PR is targetting stable 4.5. We should probably keep the code (with the modifications mentionned in the reviews), and create a new PR with it, to target master.

Stable branch should be backported from master or whatever, but here we would miss the change in master branch.

evrardjp (changes requested):
Can you move the code to NodeUpgradeableCheck and target master branch, please?

PR #1353, opened 2020-09-11
fix: Prevent unwanted upgrades

evrardjp:
## Why is this PR needed?

Without this patch, a user might try to use a SP2 skuba to
upgrade between 1.17 versions (for which packages don't exist,
and images are pointing to an incorrect location), or use a SP2
skuba to upgrade to 1.18 from old incompatible versions.

This is a problem, as those actions would brick the upgrade.

This fixes it by stopping the upgrade when those cases are
encountered.

Fixes: https://github.com/SUSE/avant-garde/issues/1943

## What does this PR do?

See above

## Info for QA

We need to test those multiple cases:
- SP2 skuba to upgrade between 1.17.x versions
- SP2 skuba to upgrade from below 1.17

will now throw an error message

You can do all you want! But that's not good.

You are now restricted. It's not user friendly, but at least it
doesn't break.

## Docs

None. Still keeps our current process.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/upgrade/node/versions.go
+++ b/internal/pkg/skuba/upgrade/node/versions.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
@@ -33,6 +34,7 @@ import (
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
 	"github.com/SUSE/skuba/internal/pkg/skuba/upgrade/addon"
 	upgradecluster "github.com/SUSE/skuba/internal/pkg/skuba/upgrade/cluster"
+	skubaconstants "github.com/SUSE/skuba/pkg/skuba"
 )
 
 type NodeVersionInfoUpdate struct {
@@ -95,6 +97,16 @@ func (nviu NodeVersionInfoUpdate) IsFirstControlPlaneNodeToBeUpgraded(client cli
 // If all preconditions are met for the given node, no error will be returned.
 func (nviu NodeVersionInfoUpdate) NodeUpgradeableCheck(client clientset.Interface, currentClusterVersion *version.Version) error {
 	errorMessages := []string{}
+	if strings.HasPrefix(nviu.Update.KubeletVersion.String(), skubaconstants.LastCaaSP4KubernetesVersion) {
+		errorMessages = append(errorMessages, fmt.Sprintf("Cannot use this version of skuba to upgrade to kubernetes version %s. Please use a skuba version below 2.0.0 instead.", skubaconstants.LastCaaSP4KubernetesVersion))
+	}
+	isKubeletTooOld, err := nviu.Update.KubeletVersion.Compare("1.17.0")
+	if err != nil {
+		return err
+	}
+	if isKubeletTooOld < 0 {
+		errorMessages = append(errorMessages, fmt.Sprintf("Make sure your node is at least to version %s before upgrading", skubaconstants.LastCaaSP4KubernetesVersion))
+	}
 	isFirstControlPlaneNodeToBeUpgraded, err := nviu.IsFirstControlPlaneNodeToBeUpgraded(client)
 	if err != nil {
 		return err


internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -95,6 +97,16 @@ func (nviu NodeVersionInfoUpdate) IsFirstControlPlaneNodeToBeUpgraded(client cli
 // If all preconditions are met for the given node, no error will be returned.
 func (nviu NodeVersionInfoUpdate) NodeUpgradeableCheck(client clientset.Interface, currentClusterVersion *version.Version) error {
 	errorMessages := []string{}
+	if strings.HasPrefix(nviu.Update.KubeletVersion.String(), skubaconstants.LastCaaSP4KubernetesVersion) {
+		errorMessages = append(errorMessages, fmt.Sprintf("Cannot use this version of skuba to upgrade to kubernetes version %s. Please use a SP1 skuba binary instead.", skubaconstants.LastCaaSP4KubernetesVersion))
```
chentex, line 101: I don't think the user will understand with `SP1 skuba`, maybe provide a package version? Or `skuba with version < X.X.X`

chentex (commented):
I just have this nit pick, for the message

internal/pkg/skuba/upgrade/node/versions.go
evrardjp: Good idea! What does this map to? Do you know by any chance?

chentex: According to this https://build.suse.de/package/view_file/SUSE:SLE-15-SP2:Update:Products:CaaSP:4.5/skuba/skuba.spec?expand=1 it should be `< 2.0.0`

evrardjp: the last release seems to be 2.1.4 or something, so Yes it might be wise to have 2.0.0.

internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -95,6 +97,16 @@ func (nviu NodeVersionInfoUpdate) IsFirstControlPlaneNodeToBeUpgraded(client cli
 // If all preconditions are met for the given node, no error will be returned.
 func (nviu NodeVersionInfoUpdate) NodeUpgradeableCheck(client clientset.Interface, currentClusterVersion *version.Version) error {
 	errorMessages := []string{}
+	if strings.HasPrefix(nviu.Update.KubeletVersion.String(), skubaconstants.LastCaaSP4KubernetesVersion) {
+		errorMessages = append(errorMessages, fmt.Sprintf("Cannot use this version of skuba to upgrade to kubernetes version %s. Please use a skuba version below 2.0.0 instead.", skubaconstants.LastCaaSP4KubernetesVersion))
```
Klaven, line 101: this version is a bit confusing too me. skuba version 2.0.0? should this not be caasp versions? to be honest. I don't even know what `skuba version 2.0.0` is.

Klaven (commented):
To be honest the commit message and the PR message where almost as confusing as the error messages. I think I understand the concept that is needed and what this solves but I would like to see some work on the wording.

Also, as skuba does not support multiple versions of kubernetes. aka, CaaSP 4.2 supports k8s 1.17 I think it would just add confusion to talk about versions other then CaaSP/skuba version.

chentex (approved):

internal/pkg/skuba/upgrade/node/versions.go
evrardjp: Can you propose an alternative? I used SP1 before and it was confusing.

evrardjp:
What's confusing in the commit message?

internal/pkg/skuba/upgrade/node/versions.go
evrardjp: Would "skuba from CaaSP 4 on SLES 15 SP1 (version below 2.0)" be better?

I will do that in a separate PR.

evrardjp:
@atighineanu if you could validate this from QA perspective, that would be awesome.

internal/pkg/skuba/upgrade/node/versions.go
Klaven: I like that, thought I am still not sure about adding the `version 2.0` stuff as unless you look under `tags` of this repository it is meaningless. even our branching is based off of CaaSP version. but that is fine.

evrardjp:
I should probably mention I haven't tested those :s

internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -95,6 +97,16 @@ func (nviu NodeVersionInfoUpdate) IsFirstControlPlaneNodeToBeUpgraded(client cli
...
+	}
+	isKubeletTooOld, err := nviu.Update.KubeletVersion.Compare("1.17.0")
+	if err != nil {
+		return err
+	}
+	if isKubeletTooOld < 0 {
+		errorMessages = append(errorMessages, fmt.Sprintf("Make sure your node is at least to version %s before upgrading", skubaconstants.LastCaaSP4KubernetesVersion))
```
evrardjp, line 108: "before trying an upgrade with this version of skuba for CaaSP 4.5 on SLES 15 SP2" is probably a better wording.

atighineanu:
e2acfaeff94a:~/go/src/mkcaasp/tests/ginkgoscenarios/imba-cluster # skuba node upgrade apply --user sles --sudo --target 10.84.73.249
Unable to apply node upgrade: worker-pimp-comrade-00 is not upgradeable until all control plane nodes are upgraded


It looks nice. However, when I tried to upgrade a master - skuba continued unimpeded trying to upgrade an SP1 master to k8s 1.18 (of course it failed). Note: I run skuba 2.0 from outside the cluster. None of the caasp nodes are a management node.

evrardjp:
The idea was to have this version of skuba (on sp2), but preventing upgrading an upgrade from something below 1.17.x to 1.17.x

atighineanu:
```
e2acfaeff94a:~/go/src/mkcaasp/tests/ginkgoscenarios/imba-cluster # skuba node upgrade apply --user sles --sudo --target 10.84.72.234
Current Kubernetes cluster version: 1.16.2
Latest Kubernetes version: 1.18.6
Current Node version: 1.16.2


Unable to apply node upgrade: node master-general-00 cannot be upgraded yet. The following errors were detected:
 - Cannot use this version of skuba to upgrade to kubernetes version 1.17. Please use a skuba version below 2.0.0 instead.
 - There are addon upgrades available for the current cluster version (1.16.2) that need to be applied first```

Nice. 
LGTM.

evrardjp:
ofc I will adapt the wording as a follow up commit, to be consistent across #1354, #1353, and #1352.

chentex (dismissed):

mssola (approved):

chentex (approved):
LGTM

PR #1356, opened 2020-09-14
Fix maintenance caasp v4 build

pablochacin:
## Why is this PR needed?

The build process in maintenance-caasp-v4 branch is broken due to outdated dependencies. 

## What does this PR do?

- Updates shellchecker version
- Fixes broken url to go module (vbom)

These changes are backports from changes applied to master branch

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,7 +99,7 @@ lint: deps
 .PHONY: deps
 deps:
 	test -f $(BINPATH)/golangci-lint || curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(BINPATH) v1.21.0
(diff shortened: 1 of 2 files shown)


jordimassaguerpla (approved):

Klaven (commented):
I'm not sure why we would want to go to an older version of spell check. did I miss something? I'm also assuming that this is the same package though you changed where we where getting it from.

pablochacin:
> I'm not sure why we would want to go to an older version of spell check. did I miss something? I'm also assuming that this is the same package though you changed where we where getting it from.

We are using the same version than in 4.5.

Itxaka (approved):

davidcassany (approved):

jodavis-suse:
As far as a backport, this looks ok and the urls resolve.  There are newer versions, but I think the point here is to match what works.

PR #1357, opened 2020-09-14
Rename caasp-maintenance-v4 to caasp-maintenance-v4.2

jordimassaguerpla:
Because CaaSP 4.5 is master, so to avoid confusions.

fixes https://github.com/SUSE/avant-garde/issues/1866

### DIFF
--- a/ci/jenkins/jobs.yaml
+++ b/ci/jenkins/jobs.yaml
@@ -19,7 +19,7 @@
           - openstack
           - vmware
       version:
-          - v4:
+          - v4.2:
               sonobuoy_version: 'v0.17'
               schedule: '1-3'
               branch: maintenance-caasp-v4.2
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
maintenance-caasp-v4.2 needs to be rebased

jordimassaguerpla:
After merging this, maintenance-caasp-v4 should be renamed to old-maintenance-caasp-v4

Itxaka (approved):
lgtm

ci/jenkins/jobs.yaml
```diff
@@ -22,7 +22,7 @@
           - v4:
```
pablochacin, line 22: I would change this to 4.2 for consistency

pablochacin (changes requested):

pablochacin:
I noticed that this PR is against the maintenance-v4 branch and it is unnecessary because the job definition is only read from master. I'm more in the line of either a) removing the job definition from this branch to avoid confusion or b) backport the definition from master to updated it.

jordimassaguerpla:
@pablochacin what is the simplest thing to do ?

jordimassaguerpla:
We do not need this one. The jobs we run are from master. We should instead remove the jobs because this is confusing.

PR #1365, opened 2020-09-21
Enable kubelet server TLS bootstrap

jenting:
## Why is this PR needed?

The _kucero_ had been released at skuba 4.5.0, so we could enable kubelet server TLS bootstrap to let kubelet send out CSR when kubelet server certificate is going to expire, then _kucero_ will sign the CSR and kubelet will save it to local disk `/var/lib/kubelet/pki` and update the symbolic link `/var/lib/kubelet/pki/kubelet-server-current.pem` points to the latest server certificate.

refs to https://github.com/SUSE/avant-garde/issues/1958

## What does this PR do?

Previously, the kubelet server certificate is self-signed by skuba with kubelet CA cert/key pair.
With this PR, the kubelet sends server CSR in K8s cluster, and kucero signed the server certificate to kubelet with kubelet CA cert/key pair, then kubelet save the server certificate in `/var/lib/kubelet/pki/kubelet-server-current.pem`.

N/A

## Info for QA

### Related info

N/A

The kubelet configuration `/var/lib/kubelet/config.yaml` would have `serverTLSBootstrap: false` in fresh install _or_ upgraded cluster.

The kubelet configuration `/var/lib/kubelet/config.yaml` would not have `serverTLSBootstrap: true` in fresh install _or_ upgraded cluster.

## Docs

https://github.com/SUSE/doc-caasp/pull/1002

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/kubelet.go
+++ b/internal/pkg/skuba/deployments/ssh/kubelet.go
@@ -18,17 +18,11 @@
 package ssh
 
 import (
-	"crypto/x509"
 	"io/ioutil"
-	"net"
 	"os"
 	"path/filepath"
-	"strings"
 
-	"github.com/pkg/errors"
-	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/kubernetes/cmd/kubeadm/app/constants"
-	"k8s.io/kubernetes/cmd/kubeadm/app/util/pkiutil"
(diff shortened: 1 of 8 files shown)


jenting:
close, let _kucero_ do it for you.

PR #1369, opened 2020-09-22
Update kubelet.conf during node upgrade (bsc#1176578)

jenting:
# Why is this PR needed?

If the cluster is deployed priori to Kubernetes version 1.17,
the kubelent client cert and key is embedded in /etc/kubernetes/kubelet.conf with base64 encoded.
however, it should points to the /var/lib/kubelet/pki/kubelet-client-current.pem for kubelet client cert rotation automatically.
and, the kubeadm did not takes care update the kubelet.conf during kubeadm upgrade.

Therefore, we let skuba to help on this to update the /etc/kubernetes/kubelet.conf from base64-encoded data to
point to the file /var/lib/kubelet/pki/kubelet-client-current.pem

Fixes bsc#1176578
xrefs to https://github.com/SUSE/avant-garde/issues/1961

## What does this PR do?

Calls during upgrade:
- `sed -i "s|client-certificate-data: .*|client-certificate: /var/lib/kubelet/pki/kubelet-client-current.pem|" /etc/kubernetes/kubelet.conf`
- `sed -i "s|client-key-data: .*|client-key: /var/lib/kubelet/pki/kubelet-client-current.pem|" /etc/kubernetes/kubelet.conf`

https://github.com/kubernetes/kubeadm/issues/1753#issuecomment-695886896
https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#check-certificate-expiration

## Info for QA

### Related info

N/A

1. Prepare a cluster bootstrap priority to Kubernetes 1.17
2. Watch /etc/kubernetes/kubelet.conf, you could see that the client cert/key is embbed in configuration file with base64 encoded
3. Upgrade cluster by skuba, the /etc/kubernetes/kubelet.conf client cert/key _won't_ points to `/var/lib/kubelet/pki/kubelet-client-current.pem`

1. Prepare a cluster bootstrap priority to Kubernetes 1.17
2. Watch /etc/kubernetes/kubelet.conf, you could see that the client cert/key is embbed in configuration file with base64 encoded
3. Upgrade cluster by skuba, the /etc/kubernetes/kubelet.conf client cert/key _would_ points to `/var/lib/kubelet/pki/kubelet-client-current.pem`

## Docs

https://github.com/SUSE/doc-caasp/pull/998

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/kubelet.go
+++ b/internal/pkg/skuba/deployments/ssh/kubelet.go
@@ -40,6 +40,7 @@ import (
 func init() {
 	stateMap["kubelet.rootcert.upload"] = kubeletUploadRootCert
 	stateMap["kubelet.servercert.create-and-upload"] = kubeletCreateAndUploadServerCert
+	stateMap["kubelet.update.config"] = kubeletUpdateConfig
 	stateMap["kubelet.configure"] = kubeletConfigure
 	stateMap["kubelet.enable"] = kubeletEnable
 }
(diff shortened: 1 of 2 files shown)


innobead (approved):
LGTM.

internal/pkg/skuba/deployments/ssh/kubelet.go
```diff
@@ -139,6 +140,18 @@ func kubeletCreateAndUploadServerCert(t *Target, data interface{}) error {
 	return nil
 }
 
+func kubeletEnableClientCertRotation(t *Target, data interface{}) error {
+	// If the cluster is deployed priori to Kubernetes version 1.17,
+	// the kubelent client cert and key is embedded in /etc/kubernetes/kubelet.conf with base64 encoded.
+	// however, it should points to the /var/lib/kubelet/pki/kubelet-client-current.pem for kubelet client cert rotation automatically.
+	// and, the kubeadm did not takes care update the kubelet.conf during kubeadm upgrade.
```
c3y1huang, line 147:
```suggestion
	// and, the kubeadm did not take care to update the kubelet.conf during kubeadm upgrade.
```

jenting:
I found that update the kubelet.conf does not work on the worker nodes, change to use `sed` to update the kubelet.conf.

jenting:
Also, I need to check if we need to restart kubelet daemon manually.
If yes, is it more suitable to build-in kucero instead?

innobead:
> I found that update the kubelet.conf does not work on the worker nodes, change to use `sed` to update the kubelet.conf.

What's the problem? Was it overrieded by the following steps? (curious)

jenting:
> > I found that update the kubelet.conf does not work on the worker nodes, change to use `sed` to update the kubelet.conf.
> 
> What's the problem? Was it overrieded by the following steps? (curious)

After discussing with @innobead in RC,  we think it's more suitable to implement it in _kucero_, close this PR first.

PR #1370, opened 2020-09-22
Informs the user about wrong OS on node

atighineanu:
## Why is this PR needed?

Without this fix, a user can try to upgrade a cluster node with an OS SLES15-SP1  while using skuba-2 (CaaSP-4.5).
The upgrade fails, but the information is not necessarily clear to the user ( packages not found).

With this fix, user will be properly informed what steps are needed to be done prior to upgrading the node.
Fixes #

appear in the changelog.

## What does this PR do?

Prior to further k8s-related versioning checks, we check if the OS corresponds to skuba-2 (CaaSP-4.5).

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/upgrade/node/versions.go
+++ b/internal/pkg/skuba/upgrade/node/versions.go
@@ -34,7 +34,7 @@ import (
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
 	"github.com/SUSE/skuba/internal/pkg/skuba/upgrade/addon"
 	upgradecluster "github.com/SUSE/skuba/internal/pkg/skuba/upgrade/cluster"
-	"github.com/SUSE/skuba/pkg/skuba"
+	skubaconstants "github.com/SUSE/skuba/pkg/skuba"
 )
 
 type NodeVersionInfoUpdate struct {
@@ -103,7 +103,7 @@ func (nviu NodeVersionInfoUpdate) NodeUpgradeableCheck(client clientset.Interfac
 	}
 	//Check if node OS SP compatible with kubernetes.latestVersion
 	if strings.Contains(nviu.Current.Node.Status.NodeInfo.OSImage, "SUSE Linux Enterprise Server 15 SP1") {
-		errorMessages = append(errorMessages, fmt.Sprintf("You are still running 15 SP1 on your node. Please upgrade to kubernetes version %s on SP1 first, and migrate to SP2 before triggering this upgrade.", skuba.LastCaaSP4KubernetesVersion))
+		errorMessages = append(errorMessages, fmt.Sprintf("You are still running 15 SP1 on your node. Please upgrade to kubernetes version %s on SP1 first, and migrate to SP2 before triggering this upgrade.", skubaconstants.LastCaaSP4KubernetesVersion))
 	}
 	if isFirstControlPlaneNodeToBeUpgraded {
 		// First check if all schedulable workers will tolerate the version we are upgrading to. If they don't, they need to be upgraded first.
--- a/internal/pkg/skuba/upgrade/node/versions.go
+++ b/internal/pkg/skuba/upgrade/node/versions.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
@@ -33,6 +34,7 @@ import (
 	"github.com/SUSE/skuba/internal/pkg/skuba/kubernetes"
 	"github.com/SUSE/skuba/internal/pkg/skuba/upgrade/addon"
 	upgradecluster "github.com/SUSE/skuba/internal/pkg/skuba/upgrade/cluster"
+	"github.com/SUSE/skuba/pkg/skuba"
 )
 
 type NodeVersionInfoUpdate struct {
@@ -98,6 +100,10 @@ func (nviu NodeVersionInfoUpdate) NodeUpgradeableCheck(client clientset.Interfac
 	isFirstControlPlaneNodeToBeUpgraded, err := nviu.IsFirstControlPlaneNodeToBeUpgraded(client)
 	if err != nil {
 		return err
+	}
+	//Check if node OS SP compatible with kubernetes.latestVersion
+	if strings.Contains(nviu.Current.Node.Status.NodeInfo.OSImage, "SUSE Linux Enterprise Server 15 SP1") {
+		errorMessages = append(errorMessages, fmt.Sprintf("You are still running 15 SP1 on your node. Please upgrade to kubernetes version %s on SP1 first, and migrate to SP2 before triggering this upgrade.", skuba.LastCaaSP4KubernetesVersion))
 	}
 	if isFirstControlPlaneNodeToBeUpgraded {
 		// First check if all schedulable workers will tolerate the version we are upgrading to. If they don't, they need to be upgraded first.
--- a/pkg/skuba/actions/node/upgrade/apply.go
+++ b/pkg/skuba/actions/node/upgrade/apply.go
@@ -63,6 +63,8 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 		return nil
 	}
 
+	// Refreshing cache info about target OS
+	target.IsSUSEOS()
 	// Check if the node is upgradeable (matches preconditions)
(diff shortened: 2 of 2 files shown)


internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -99,6 +102,10 @@ func (nviu NodeVersionInfoUpdate) NodeUpgradeableCheck(client clientset.Interfac
 	if err != nil {
 		return err
 	}
+	//Check if node OS SP compatible with kubernetes.latestVersion
+	if strings.Contains(nviu.Current.Node.Status.NodeInfo.OSImage, "SUSE Linux Enterprise Server 15 SP1") {
+		log.Fatalf("You are still running 15 SP1 on your node. Please upgrade to kubernetes version %s on SP1 first, and migrate to SP2 before triggering this upgrade.", skuba.LastCaaSP4KubernetesVersion)
```
evrardjp, line 107: do not log.Fatalf, instead, append the error into the list of error messages.

internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -99,6 +102,10 @@ func (nviu NodeVersionInfoUpdate) NodeUpgradeableCheck(client clientset.Interfac
 	if err != nil {
 		return err
 	}
+	//Check if node OS SP compatible with kubernetes.latestVersion
```
evrardjp, line 105: Are you sure that the node information is up to date? I would refresh the cache by checking the target OS (call of ` 	target.IsSUSEOS()` before running this function. (Like you did in the previous PR).

evrardjp (changes requested):
You should probably add a:

```
Related issue: SUSE/avant-garde#1667
```

Into the PR and/or commit message.

The rest looks good.

internal/pkg/skuba/upgrade/node/versions.go
atighineanu: Changed.

chentex (dismissed):
LGTM

evrardjp (commented):
The code looks good at first sight.

Can you check what's going on with the CI tests?

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -63,6 +63,8 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 		return nil
 	}
 
+	// Refreshing cache info about target OS
+	target.IsSUSEOS()
```
evrardjp, line 67: It might be worth checking the returns of this function: If it's a non SUSE OS or an error, we should indeed bail as soon as possible.

atighineanu: Corrected that. Propagated the error and checked for isSUSE bool just in case.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -63,6 +63,16 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
 	}
 
+	// Refreshing cache info about target OS
+	isSUSE, err := target.IsSUSEOS()
+	// Check if target node OS is valid
+	if !isSUSE {
+		fmt.Println("Invalid OS on target node.")
+		os.Exit(1)
+	}
+	if err != nil {
```
chentex, line 73: This should be before checking `isSUSE` variable

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -63,6 +63,16 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 		return nil
 	}
 
+	// Refreshing cache info about target OS
+	isSUSE, err := target.IsSUSEOS()
+	// Check if target node OS is valid
+	if !isSUSE {
+		fmt.Println("Invalid OS on target node.")
+		os.Exit(1)
```
chentex, line 71: Maybe just return the error instead of printing and making an os.Exit

chentex (changes requested):
I just have this two comments

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -63,6 +63,16 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
 		return nil
 	}
 
+	// Refreshing cache info about target OS
+	isSUSE, err := target.IsSUSEOS()
+	// Check if target node OS is valid
```
drpaneas, line 68: the comment here is superfluous

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -63,6 +63,15 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
+	if err != nil {
+		return err
+	}
+	if !isSUSE {
+		return err
```
evrardjp, line 73: If there is no error in hasOS, and the OS is not recognized or not matching expectations (but not incorrect), isSUSE will be false, and the error will be nil. In this case, we should (at this line) return something else than nil, IMO.

However, I am not sure what _IsSUSEOS_ intended to be (as it might have been something to support openSUSE), and therefore, we might just want to ignore the error (and therefore remove the return stanza, replacing it with a warning of some sort I guess).

An experienced skuba maintainer should give an opinion here.

atighineanu: Yeah, that was what I wrote before, an os.Exit and println. @chentex disagreed with this approach so I've changed the code.

evrardjp: I agree with chentex on the not having os.Exist and println :) 
But I also want to make sure this return is relevant.

You could instead _produce_ an error and return it, if necessary. Or log a warning. But the code, as is, looks weird to me (return nil for error while there was an error doesn't sound right)

atighineanu: Log a warning sounds fair to me.

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -63,6 +64,15 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
 	}
 
+	// Refreshing cache info about target OS
+	isSUSE, err := target.IsSUSEOS()
+	if err != nil {
+		return err
+	}
+	// Check if target node OS is valid
+	if !isSUSE {
+		log.Println("Invalid target node OS")
```
chentex, line 74: Just one small thing, sorry I didn't catch it before, can you change `log` with `klog` like here: https://github.com/SUSE/skuba/blob/master/cmd/skuba/cluster/init.go#L54 that way we maintain standard logs.

chentex (changes requested):
Sorry for this

pkg/skuba/actions/node/upgrade/apply.go
evrardjp: :+1:

chentex (approved):
LGTM thank you @atighineanu

drpaneas:
if it's not too much trouble, please squash your commits

pkg/skuba/actions/node/upgrade/apply.go
```diff
@@ -63,6 +63,16 @@ func Apply(client clientset.Interface, target *deployments.Target) error {
...
 	}
 
+	// Refreshing cache info about target OS
+	isSUSE, err := target.IsSUSEOS()
+	if err != nil {
+		return err
+	}
+	// Check if target node OS is valid
+	if !isSUSE {
+		err = errors.New(fmt.Sprintf("Invalid target node OS: %s", nodeVersionInfoUpdate.Current.Node.Status.NodeInfo.OSImage))
```
jenting, line 73: Suggest to use `fmt.Errorf` is much simpler.

evrardjp: that can even be bundled in the return stanza as:
`return fmt.Errorf("Invalid target node OS: %s", nodeVersionInfoUpdate.Current.Node.Status.NodeInfo.OSImage)`

atighineanu: Thanks a lot for hints, guys!

atighineanu:
> if it's not too much trouble, please squash your commits

Done, sir!

jenting (approved):

chentex (approved):
LGTM

PR #1383, opened 2020-10-01
remove the patch from the package

jordimassaguerpla:
We have a patch that removes the load balancer and CI specifics so they
are not released to our customers.

However, this patch very often needs to be rebased and is a paint to do
this. This blocks any automation we can put in place.

This is a "hacky" attempt to remove this code without having to do much
in the terraform code but instead in the packaging side.

### DIFF
--- a/ci/infra/vmware/output.tf
+++ b/ci/infra/vmware/output.tf
@@ -6,6 +6,10 @@ output "ip_workers" {
   value = "${zipmap(vsphere_virtual_machine.worker.*.name, vsphere_virtual_machine.worker.*.default_ip_address)}"
 }
 
+# LOAD_BALANCER
+# Please do NOT touch above comment nor below END comment. This is for removing this section,
+# specific to LOAD_BALANCER, from the package, because this is not supported.
 output "ip_load_balancer" {
   value = "${zipmap(vsphere_virtual_machine.lb.*.name, vsphere_virtual_machine.lb.*.default_ip_address)}"
 }
+# END_LOAD_BALANCER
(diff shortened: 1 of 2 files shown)


jordimassaguerpla:
This is how I tested it.
I build a skuba package with this changes.
Then I extracted the package:

`unrpm skuba-2.1.6-0.x86_64.rpm`

then I looked for the files that should have been removed:

```
find . -name terraform.tfvars.json.ci.example
find . -name lb-instance.tf
find . -name lb.tpl

```

No files were found.

Then I looked into `/usr/share/caasp/terraform/vmware/output.tf`

```
output "ip_masters" {
  value = "${zipmap(vsphere_virtual_machine.master.*.name, vsphere_virtual_machine.master.*.default_ip_address)}"
}

output "ip_workers" {
  value = "${zipmap(vsphere_virtual_machine.worker.*.name, vsphere_virtual_machine.worker.*.default_ip_address)}"
}

```

no mention of the ip balancer

dannysauer (approved):
I think this all looks good.  I particularly like adding the section markers into the terraform file and removing them at build time, as that unambiguously allows people who edit the file later to see that code in that section will not be included in a release.  The old process was too opaque; this makes more sense.

davidcassany (approved):
Regarding the load balancer I am not so sure the spec is the best place to apply this sort of changes. As it will also be part of the source package we are shipping. I had imagined to strip these files before creating the source tarball, but, in any case, I have no strong opinion about it. @dannysauer is also right that having intermediate steps tends to be opaque so this is a good argument for the current approach.

Then regarding the example files I am fine with the current hack, however all this should be solved as part of SUSE/avant-garde#1947 These files are not examples (otherwise we would ship them), these files are part of the CI logic, they shouldn't have ever been part of this folder.

So all in all, I consider it a good mid step to unblock skuba packaging automation.

PR #1390, opened 2020-10-08
Revert "Merge pull request #1384 from mrostecki/4.2-remove-cilium-init"

rktidwell:
This reverts commit 79a5c1f5a084a10e1e3d65cc2d5597b1eb4fd895, reversing
changes made to 748de90f14a49d1921f0f852faae562e6429d251.

## Why is this PR needed?

This change is reverting a commit that seems to have broken e2e tests in CI.

fixes:: https://github.com/SUSE/avant-garde/issues/1981

See https://github.com/SUSE/skuba/commit/c7c1a31881a9ee8ad43c53e92a19f95d27e1cef7

## What does this PR do?

This change is reverting a commit that seems to have broken e2e tests in CI.

### DIFF
--- a/internal/pkg/skuba/addons/cilium.go
+++ b/internal/pkg/skuba/addons/cilium.go
@@ -31,7 +31,11 @@ import (
 )
 
 func init() {
-	registerAddon(kubernetes.Cilium, renderCiliumTemplate, renderCiliumPreflightTemplate, ciliumCallbacks{}, normalPriority, []getImageCallback{GetCiliumOperatorImage, GetCiliumImage})
+	registerAddon(kubernetes.Cilium, renderCiliumTemplate, renderCiliumPreflightTemplate, ciliumCallbacks{}, normalPriority, []getImageCallback{GetCiliumInitImage, GetCiliumOperatorImage, GetCiliumImage})
+}
+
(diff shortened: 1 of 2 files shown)


dannysauer (approved):
This all looks pretty reasonable with my limited knowledge of what's going on. :)

mrostecki (changes requested):
The cilium-image is not on IBS anymore. Accepting that revert will again cause issues.

Instead, we should properly fix CI to not expect that image to be present!!!

rktidwell:
should have added the "do not merge" label, this is just to figure out what's going on with CI

rktidwell:
> The cilium-image is not on IBS anymore. Accepting that revert will again cause issues.
> 
> Instead, we should properly fix CI to not expect that image to be present!!!

Actually, it looks like we need to keep to revert this because there are some variables referenced that we haven't unwound yet. Also, there are cases where we will encounter cilium 1.5 so removing it completely in this way will cause problems. I think the best course of action is to merge this and look at a different way to unwind cilium 1.5.

mrostecki (approved):
Yes. Sorry for rejecting it before. We should actually merge it and re-open #1923.

PR #1395, opened 2020-10-12
[WIP] Autorelease jenkins

dannysauer:
## What does this PR do?

Create Jenkins workflow to automate release of skuba.

### DIFF
--- a/ci/jenkins/pipelines/skuba-autorelease.Jenkinsfile
+++ b/ci/jenkins/pipelines/skuba-autorelease.Jenkinsfile
@@ -1,0 +1,1 @@
+# comment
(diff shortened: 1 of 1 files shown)


dannysauer:
Thoughts copied from #1371:
> the Jenkins flow is quite a bit simpler. Because the Action creates a release with an artifact containg the OBS files, no checkout is necessary. All the Jenkens script needs to do is:
> 
>1.    get OBS tarball from the current skuba release (see #1371 (comment) above)
>1.    extract tarball
>1.    prepend changelog from tarball to current changelog
>1.    update non-changelog files in OBS repo direct from release tarball
>1.    run services to update vendored modules
>1.    commit
>1.    create SR (maybe)
>
> I've been looking at the details of the OBS services we have, and several of those steps might be doable via _service directly; what I'd really like to do is be able to just hit the "run service" endpoint in IBS and have it do the download, extract, and changelog update. But we might need to do the download and changelog update separately, and then merely let obs do the extraction & vendoring

Itxaka:
closing

PR #1408, opened 2020-10-23
Allow to upgrade addons

mjura:
## Why is this PR needed?

There might be situation for multi-master setup when
```
> skuba node upgrade apply --target <MASTER2> -s -u sles -v 5

Unable to apply node upgrade: node master02 cannot be upgraded yet. The following errors were detected:
 - There are addon upgrades available for the current cluster version (1.17.13) that need to be applied first

> skuba addon upgrade apply                                   

Unable to Apply addons upgrade: [apply] Not all nodes match clusterVersion 1.17.13

> skuba node upgrade apply --target <MASTER2> -s -u sles -v 5

Unable to apply node upgrade: node master02 cannot be upgraded yet. The following errors were detected:
 - There are addon upgrades available for the current cluster version (1.17.13) that need to be applied first
```

## What does this PR do?

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/pkg/skuba/actions/addon/upgrade/apply.go
+++ b/pkg/skuba/actions/addon/upgrade/apply.go
@@ -48,7 +48,7 @@ func Apply(client clientset.Interface) error {
 	fmt.Println()
 
 	if !allNodesMatchClusterVersion {
-		return errors.Errorf("[apply] Not all nodes match clusterVersion %s", currentVersion)
+		fmt.Printf("[apply] Not all nodes match clusterVersion %s\n", currentVersion)
 	}
 
 	clusterConfiguration, err := kubeadm.GetClusterConfiguration(client)


davidcassany:
I love the simplicity of this patch :+1: thanks much @mjura 
I am not approving it as I am not familiar enough with skuba to foresee the impact of this change. Is it just changing the return code on addons upgrade plan command? If this is the case and it solves the issue I believe this is a very interesting option.

dannysauer:
So the error is just downgraded to a warning?  Seems reasonable.  I'll approve and merge once @chentex has a successful test result.

mjura:
Kubeadm doesn't require for updating addons that all nodes in the cluster will be in the same version. It is even impossible. Previous `return error` was too restrictive. We can print information and not block addons update.

thehejik:
I'm testing this PR. In general it works, but there is one possible issue... I performed `skuba node upgrade apply` on master node and it says `successfully upgraded` but when I run `skuba node upgrade plan <master-node>` once again it does NOT return `Node caasp-master-th-423to424-0 is up to date` but a list of components to upgrade (where etcd version differs) instead:
```
thehejik@th-caasp2:/app/cluster_mrb/cluster> skuba node upgrade plan caasp-master-th-423to424-0
Current Kubernetes cluster version: 1.17.13
Latest Kubernetes version: 1.17.13
Current Node version: 1.17.13

Component versions in caasp-master-th-423to424-0
  - apiserver: 1.17.13 -> 1.17.13
  - controller-manager: 1.17.13 -> 1.17.13
  - scheduler: 1.17.13 -> 1.17.13
  - etcd: 3.4.3 -> 3.4.13
  - kubelet: 1.17.13 -> 1.17.13
  - cri-o: 1.16.1 -> 1.16.1
```

It works normally on worker nodes but not for master. Most probably the problem is caused by etcd for which latest `etcd:3.4.13` image has been pulled but the pod is still using outdated `etcd:3.4.3` image. Also this can be a reason why the upgrade works fine for worker nodes (no etcd on worker nodes).

* used repo: http://download.suse.de/ibs/Devel:/CaaSP:/4.0:/Staging:/4.2.4/SUSE_SLE-15-SP1_Update_Products_CASP40_Update_standard 
* used registry mirror: `registry.suse.de/devel/caasp/4.0/staging/4.2.4/suse_sle-15-sp1_update_products_casp40_update_containers/caasp/v4`

thehejik:
For upgrade multi-master cluster I had to skip `skuba addon upgrade apply` step (after triggering `skuba-update` on nodes and before `skuba node upgrade`) which is still complaining that the k8s versions on nodes are not the same. 

If I skip that and go directly to step `skuba node upgrade apply` it works - but not sure if this is desired/expected solution.

Until now we always had to update addons with "old" platform first before running upgrade and then once again (update addons) with "new" platform.

Also when I wanted to make sure the cluster has been updated successfully I saw this:
```
thehejik@th-caasp2:/app/cluster_iaM/cluster> skuba cluster upgrade plan
Current Kubernetes cluster version: 1.17.13
Latest Kubernetes version: 1.17.13

Some nodes do not match the current cluster version (1.17.13):
  - caasp-master-th-423to424-0: up to date
  - caasp-master-th-423to424-1; current kubelet package version: 1.17.13, kube-apiserver container version: 1.17.4 (upgrade suggested)
  - caasp-master-th-423to424-2; current kubelet package version: 1.17.13, kube-apiserver container version: 1.17.4 (upgrade suggested)
  - caasp-worker-th-423to424-0: up to date
  - caasp-worker-th-423to424-1: up to date

Addons at the current cluster version 1.17.13 are up to date.
```

mjura:
I was able to reproduce following issue on multi-master CaaSP v4.2.3 cluster

After `/usr/bin/skuba node upgrade apply -s -u sles -t MASTER_IP -v5`

```
I1027 08:09:12.123277   14618 ssh.go:193] [upgrade/etcd] Non fatal issue encountered during upgrade: the desired etcd version for this Kubernetes version "v1.17.13" is "3.4.3-0", but the current etcd version is "3.4.3". Won't downgrade etcd, instead just continue
```

Let's move this discussion to https://github.com/SUSE/avant-garde/issues/2028

PR #1409, opened 2020-10-23
Fix nodeVersion for all upgrade plan

dannysauer:
(cherry picked from commit 11f2c49d4adc3889a9891b73fffa04a2eeb42db7)
Signed-off-by: Danny Sauer <dsauer@suse.com>

## Why is this PR needed?

Backport version detection code to 4.2.x

This backport (and the tag) was added to branch after the previous PR was merged.  This branch can be deleted after merging if so desired, but it would be ideal to merge this one instead of backporting again so the 1.4.5 tag will continue to work properly. :)

### DIFF
--- a/pkg/skuba/actions/cluster/upgrade/plan.go
+++ b/pkg/skuba/actions/cluster/upgrade/plan.go
@@ -86,9 +86,17 @@ func plan(client clientset.Interface, availableVersions []*version.Version, clus
 			if nodeVersionInfo.EqualsClusterVersion(currentClusterVersion) {
 				fmt.Printf("  - %s: up to date", nodeName)
 			} else if !nodeVersionInfo.ToleratesClusterVersion(currentClusterVersion) {
-				fmt.Printf("  - %s; current version: %s (upgrade required)", nodeName, nodeVersionInfo.String())
+				if nodeVersionInfo.IsControlPlane() {
+					fmt.Printf("  - %s; current kubelet package version: %s, kube-apiserver container version: %s (upgrade required)", nodeName, nodeVersionInfo.KubeletVersion.String(), nodeVersionInfo.APIServerVersion.String())
+				} else {
+					fmt.Printf("  - %s; current kubelet package version: %s (upgrade required)", nodeName, nodeVersionInfo.KubeletVersion.String())
+				}
 			} else {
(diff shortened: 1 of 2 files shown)


pkg/skuba/actions/cluster/upgrade/plan.go
```diff
@@ -86,9 +86,17 @@ func plan(client clientset.Interface, availableVersions []*version.Version, clus
 			if nodeVersionInfo.EqualsClusterVersion(currentClusterVersion) {
 				fmt.Printf("  - %s: up to date", nodeName)
 			} else if !nodeVersionInfo.ToleratesClusterVersion(currentClusterVersion) {
-				fmt.Printf("  - %s; current version: %s (upgrade required)", nodeName, nodeVersionInfo.String())
+				if nodeVersionInfo.IsControlPlane() {
+					fmt.Printf("  - %s; current kubelet package version: %s, kube-apiserver container version: %s (upgrade required)", nodeName, nodeVersionInfo.KubeletVersion.String(), nodeVersionInfo.APIServerVersion.String())
```
evrardjp, line 90: current kubelet **package** version is highly confusing, as kubeletversion is the version reported by kubelet, which is a runtime value. It means the package can technically be upgraded, as long as the process is not restarted, it will report the old version.

In the case of kubelet and an upgrade, the kubelet package _must be upgraded_ at some point and then restarted during the process.

I think it would be best to name this "Current reported kubelet version", but this is a nit. As this is a backport, I am fine with keeping this, but it tickles me a little bit.

evrardjp (commented):
I am sorry, I couldn't grok the commit message or the PR text. Maybe it's because the morning, I don't know. Could you rephrase the PR so that it's understandable, 6 months from now, without context? 

The rest looks okay.

chentex (approved):
LGTM

pkg/skuba/actions/cluster/upgrade/plan.go
chentex: I created a new issue to tackle this.

Itxaka (approved):

pkg/skuba/actions/cluster/upgrade/plan.go
mjura: To be consistent we should report is as `current running kubelet version`

mjura (approved):
:+1:

PR #1416, opened 2020-11-03
Add file permission for upload file method to fix CVE-2020-8029 (bsc#1177362)

jenting:
## Why is this PR needed?

fix CVE-2020-8029 (aka bsc#1177362), add file permission for the upload file method.
It prevents when uploading the key, the attacker could steal the key by watching the file
because the file creates with permission `0644` first, and then `chmod` to `0400`.

Fixes bsc#1177362
Fixes CVE-2020-8029

## What does this PR do?

Add set the file permission parameter to `UploadFile`/`UploadFileContents` method
- caller `UploadFile`: it re-uses the local file's permission
- caller `UploadFileContents`: it uses file permission `0644` 

N/A

## Info for QA

N/A

### Related info

N/A

When node bootstrap/join, we could use another user account and watch `/var/lib/kubelet/pki/kubelet.key`, this user has the ability to get the content of `/var/lib/kubelet/pki/kubelet.key` since the file permission is `0644` when creating, and then set to `0400`.

When node bootstrap/join, we could not use another user account to get the content of `/var/lib/kubelet/pki/kubelet.key` since the file permission re-uses the local file permission `0600`.

## Docs

N/A

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/deployments/ssh/files.go
+++ b/internal/pkg/skuba/deployments/ssh/files.go
@@ -20,6 +20,7 @@ package ssh
 import (
 	"encoding/base64"
 	"fmt"
+	"os"
 	"path"
 
 	"k8s.io/klog"
@@ -27,7 +28,7 @@ import (
 
 // UploadFileContents creates a file with the content sent
 // into a target system's specific path.
-func (t *Target) UploadFileContents(targetPath, contents string) error {
+func (t *Target) UploadFileContents(targetPath, contents string, perm os.FileMode) error {
 	klog.V(1).Infof("uploading to remote file %q with contents", targetPath)
 	dir, _ := path.Split(targetPath)
 	encodedContents := base64.StdEncoding.EncodeToString([]byte(contents))
--- a/internal/pkg/skuba/deployments/ssh/files.go
+++ b/internal/pkg/skuba/deployments/ssh/files.go
@@ -35,8 +35,31 @@ func (t *Target) UploadFileContents(targetPath, contents string, perm os.FileMod
 	if _, _, err := t.silentSsh("mkdir", "-p", dir); err != nil {
 		return err
 	}
-	_, _, err := t.silentSshWithStdin(encodedContents, "base64", "-d", "-w0", fmt.Sprintf("> %s", targetPath))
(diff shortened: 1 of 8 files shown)


internal/pkg/skuba/deployments/ssh/files.go
```diff
@@ -20,24 +20,48 @@ package ssh
...
-func (t *Target) UploadFileContents(targetPath, contents string) error {
+func (t *Target) UploadFileContents(targetPath, contents string, perm os.FileMode) error {
 	klog.V(1).Infof("uploading to remote file %q with contents", targetPath)
 	dir, _ := path.Split(targetPath)
 	encodedContents := base64.StdEncoding.EncodeToString([]byte(contents))
 	if _, _, err := t.silentSsh("mkdir", "-p", dir); err != nil {
 		return err
 	}
-	_, _, err := t.silentSshWithStdin(encodedContents, "base64", "-d", "-w0", fmt.Sprintf("> %s", targetPath))
+	_, _, err := t.silentSshWithStdin(encodedContents, "umask", osFileModeToUmaskSymbolMode(perm), "&&", "base64", "-d", "-w0", fmt.Sprintf("> %s", targetPath))
```
ereslibre, line 38: We can avoid the inverse calculation of the umask if we do `install -m <mode> /dev/null <target> && base64...` to create the file -- it also reinforces the mode if the file exists already.

ereslibre (commented):
Thank you @jenting, one comment.

internal/pkg/skuba/deployments/ssh/files.go
jenting: > We can avoid the inverse calculation of the umask if we do `install -m <mode> /dev/null <target> && base64...` to create the file -- it also reinforces the mode if the file exists already.

Sound good, I had thought of using `install` before, let me update the PR.

ereslibre (approved):
LGTM, thank you @jenting!

innobead (approved):
LGTM :1st_place_medal:

PR #1422, opened 2020-11-03
Remove kucero addon for k8s 1.17.4

davidcassany:
Kucero backport was done before including 1.17.13 k8s in versions.go as
a consequence it was backported for 1.17.4 release which is not the
intended result. Kucero should only appear in versions.go starting from
1.17.13 k8s version.


jenting (approved):

chentex (approved):
LGTM

PR #1423, opened 2020-11-04
Check if kubelet was upgraded before APIserver

mjura:
## Why is this PR needed?

We have to check if kubelet package was upgraded before kube-apiserver, which is incorrect.

## What does this PR do?

Checking if kubelet package is in right version.

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine

### DIFF
--- a/internal/pkg/skuba/kubernetes/node_version.go
+++ b/internal/pkg/skuba/kubernetes/node_version.go
@@ -118,14 +118,25 @@ func (nvi NodeVersionInfo) DriftsFromClusterVersion(clusterVersion *version.Vers
 
 func (nvi NodeVersionInfo) ToleratesClusterVersion(clusterVersion *version.Version) bool {
 	if nvi.IsControlPlane() {
+		// When worker checks for tolerates cluster version, the API servers should ALL have been
+		// upgraded at the same version. No drift should be allowed.
 		if nvi.APIServerVersion.Major() != clusterVersion.Major() ||
-			nvi.APIServerVersion.Minor() != clusterVersion.Minor() {
+			(nvi.APIServerVersion.Major() == clusterVersion.Major() &&
+				nvi.APIServerVersion.Minor() != clusterVersion.Minor()) ||
+			(nvi.APIServerVersion.Major() == clusterVersion.Major() &&
+				nvi.APIServerVersion.Minor() == clusterVersion.Minor() &&
+				nvi.APIServerVersion.Patch() != clusterVersion.Patch()) {
 			return false
 		}
 	}
 
-	return nvi.KubeletVersion.Minor() == clusterVersion.Minor() ||
-		nvi.KubeletVersion.Minor()+1 == clusterVersion.Minor()
+	// When checking this function, we should prevent kubelets from being ahead.
+	// We should only tolerate the workers to be lagging behind before upgrading them
+	return (nvi.KubeletVersion.Major() == clusterVersion.Major() &&
+		nvi.KubeletVersion.Minor()+1 == clusterVersion.Minor()) ||
+		(nvi.KubeletVersion.Major() == clusterVersion.Major() &&
+			nvi.KubeletVersion.Minor() == clusterVersion.Minor() &&
+			nvi.KubeletVersion.Patch() <= clusterVersion.Patch())
 }
 
 // AllNodesVersioningInfo returns the version info for all nodes in the cluster
--- a/internal/pkg/skuba/upgrade/node/versions.go
+++ b/internal/pkg/skuba/upgrade/node/versions.go
@@ -98,6 +98,19 @@ func (nviu NodeVersionInfoUpdate) NodeUpgradeableCheck(client clientset.Interfac
 	if err != nil {
 		return err
 	}
+
+	// Explain user to downgrade kubelet package when it was installed ahead kube-apiserver.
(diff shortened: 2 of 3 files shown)


internal/pkg/skuba/kubernetes/node_version.go
```diff
@@ -259,7 +259,7 @@ func AllControlPlanesMatchVersionWithVersioningInfo(allNodesVersioningInfo NodeV
 		if !nodeInfo.IsControlPlane() {
 			continue
 		}
-		if !nodeInfo.ToleratesClusterVersion(clusterVersion) {
+		if !nodeInfo.EqualsClusterVersion(clusterVersion) {
```
evrardjp, line 262: I think this should stay ToleratesClusterVersion should be kept and modified.
Obviously the expectation is not the same as before, and we don't want to tolerate a different patch version, which can be added just after https://github.com/SUSE/skuba/blob/2d340d8b05b0a5a45e567c68333484d9540dd8d4/internal/pkg/skuba/kubernetes/node_version.go#L123
This would allow us to continue with our test coverage.

Also, in that same function https://github.com/SUSE/skuba/blob/2d340d8b05b0a5a45e567c68333484d9540dd8d4/internal/pkg/skuba/kubernetes/node_version.go#L129 might need editing if we don't tolerate the same things for workers now (which is not the scope of this PR).

internal/pkg/skuba/upgrade/node/versions.go
```diff
@@ -98,6 +98,17 @@ func (nviu NodeVersionInfoUpdate) NodeUpgradeableCheck(client clientset.Interfac
 	if err != nil {
 		return err
 	}
+	if nviu.Current.IsControlPlane() {
```
evrardjp, line 101: If you fix the ToleratesClusterVersion above,  you don't need to add this:
it will be covered a little below on kubernetes.AllControlPlanesMatchVersion which calls AllControlPlanesMatchVersionWithVersioningInfo .
That check can be moved up if necessary.

evrardjp (commented):
The idea is okay, and its implementation will work.
However, I think it would be best to reuse the existing function and harden it, as we effectively changed the requirements.

internal/pkg/skuba/upgrade/node/versions.go
mjura: Nope, this `if` condition is part of function `NodeUpgradeableCheck` and `ToleratesClusterVersion` or `EqualsClusterVersion` will no check if local `kubelet` package is in correct version to corresponding `kube-apiserver`

mjura: It have to stay otherwise we will hit problem from https://github.com/SUSE/avant-garde/issues/2028#issuecomment-720449984

internal/pkg/skuba/kubernetes/node_version.go
mjura: In CaaSP we always shipping fixed and consistent versions of Kubernetes packages. `TolerateClusterVersion` is not checking Patch() version which breaking Patch level upgrade for our case CaaSP v4.2.4

mjura:
> The idea is okay, and its implementation will work.

Yeap, it was already tested and confirmed by QA and Jenkins

> However, I think it would be best to reuse the existing function and harden it, as we effectively changed the requirements.

I agree that's way we are going to reuse `EqualsClusterVersion` function, but additional check for kubelet version is needed. ControlPlane should be upgraded first before kubelet package, if it was done accidentally then kubelet package should be downgraded.

Yes, skuba should hold and mark kubernetes packages as not to upgrade, but it is work for other PR.

evrardjp:
Sorry, I meant that ToleratesClusterVersion could be adapted to have an extra check for the patch version if necessary.
This way:
- Equals stays the same and is for strict equality (used for "cluster is up to date" and nothing else)
- Tolerate is for upgrades (allows lower versions of kubelet (but not higher ofc!), as it doesn't need to be strictly equal). Tolerates is used for plan and for the apply, so it makes it reused in multiple places.
- Of course, we could reuse Equal inside Tolerate if necessary to avoid duplication of code.

I agree on the hold and mark being on another PR.

BTW this is just my recommendation for Tolerates. I think if it's enough tested and we don't have any regression by using another function we should go ahead with this implementation. I am just playing on the careful side :)

evrardjp:
> Yeap, it was already tested and confirmed by QA and Jenkins

I trust those! :firefighter:

chentex (approved):
If QA gives green light let's go ahead with this.

mjura:
I have included comments from @evrardjp and re-tested this change again.

@thehejik Please re-test once again and let's merge it

internal/pkg/skuba/upgrade/node/versions.go
evrardjp: I am fine with having a check in here.

internal/pkg/skuba/kubernetes/node_version.go
evrardjp: Yes that was exactly what I meant :)

chentex (approved):
LGTM

jenting (approved):

Itxaka (approved):

PR #1429, opened 2020-11-06
Do not tolerate APIServer patch level version

mjura:
## Why is this PR needed?

When worker checks for tolerates cluster version, the API servers should ALL have been upgraded at the same version. No drift should be allowed.

Fixes: https://github.com/SUSE/avant-garde/issues/2028

# Merge restrictions

(Please do not edit this)

We are in *v4-maintenance phase*, so we will restrict what can be merged to prevent unexpected surprises:

    What can be merged (merge criteria):
        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine
        there is a PR for updating documentation (or a statement that this is not needed)

PR #1430, opened 2020-11-09
Do not tolerate APIServer patch level version

mjura:
## Why is this PR needed?

It was backported from https://github.com/SUSE/skuba/pull/1429

When worker checks for tolerates cluster version, the API servers should ALL have been upgraded at the same version. No drift should be allowed.

Fixes: SUSE/avant-garde#2028

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


chentex (approved):
LGTM

jenting (approved):

PR #1435, opened 2020-11-12
CI tests can't distinguish patch level versions of k8s

davidcassany:
On the CI we are always installing the latest k8s version for a given
minor version. So that the previous and current k8s version for the
upgrade test can't be under the same minor version, a patch level
divergence is not enough to distinguish them. The distinction is done
at pattern level and the patterns only divide minor version changes, not
patch level changes.

Fixes SUSE/avant-garde#2027

### DIFF
--- a/ci/infra/testrunner/tests/utils.py
+++ b/ci/infra/testrunner/tests/utils.py
@@ -2,7 +2,7 @@ import signal
 import time
 import yaml
 
-PREVIOUS_VERSION = "1.17.4"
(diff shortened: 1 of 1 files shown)


davidcassany:
I am convinced this is required, however I am not that sure that only this patch will fix the upgrade daily tests on maintenance branch as I have not dived into it.

chentex (approved):

Itxaka (approved):

PR #1458, opened 2020-12-04
Drop kucero completely

Itxaka:
kucero was backported to 4.2 and then dropped, but the addon is
still being registered so it tries to render and fails due to
some constants being empty like KuceroImage

Just drop the whole thing if we are not using it

## Why is this PR needed?

Fixes #

appear in the changelog.

## What does this PR do?

## Info for QA

### Related info

* upstream issues

this issue is not fixed.

## Docs

# Merge restrictions

(Please do not edit this)

        2 approvals:
            1 developer: code is fine
            1 QA: QA is fine


davidcassany (changes requested):
This might turn our tests green but it does not solve the underlaying issue of the tests and it removes a feature we are about to ship to customers. I'd say the upgrade tests are wrong by design, we were using a trick to not upgrade skuba within the tests and now it finally raised an issue.

@Itxaka if you don't mind I am adding the `dont merge` label. I assume this PR is for testing purposes only.

jenting (changes requested):
It's a feature request at v4.2, please do not remove it but fix the root cause.

Itxaka:
Thanks @davidcassany and @jenting , yes this was a test pr to understand what was going on and it seems like I do understand it now.

Im wondering if the initialization of the kucero addon could be done directly on the `1.17.13` path only as otherwise it will conflict when deploying any other lower versions (1.16 or 1.15) as they will both try to render the kucero and fail because there is no `KuceroImage` var in their path. Or the addon removed from the consts if we are going a non supported path? Or maybe even caching the error when rendering and disabling the addond completely?

Looks like a skuba bug in managing versions

PR #1466, opened 2021-01-19
Align versions.go with latest container images

davidcassany:
## Why is this PR needed?

Make sure to include bpftool fix in cilium image and properly align ganway image with the latest released image.

## What does this PR do?

Updates the version.go file to ensure the updated images for release 4.5.3 are being pulled during the upgrade.


manuelbuil (commented):
Wouldn't this force customer to pull the rev5 image? And that one is still creating problems in CI, or?

davidcassany:
yes this forces to make use of the new cilium image. I removed the WIP label because the tests for this specific PR are now passing. The problem was that the updated cilium-operator-image and cilium-image were not tagged exactly the same. Once I solved that on the registry the PR tests went back to green.

manuelbuil (approved):
LGTM

PR #1467, opened 2021-05-06
Update container addons references for a new release

davidcassany:
## Why is this PR needed?

Required for renewal of addon containers for CaaSP 4.5.4 release

## What does this PR do?

bump addon image references


davidcassany:
Merging since I just realized our workers in the CI got updated with wrong terraform packages. Fixing it requires changing the repo set and redeploying them. I am not going to do that now for this simple change on addons image versions.

PR #1474, opened 2022-01-14
Add systemd as cgroup-driver to kubelet, bsc#1191654

mjura:
## Why is this PR needed?

We want to migrate cgroup-driver from cgroupfs to systemd. Which resolve problem with error log in kubelet
```
Failed to create existing container
```

Fixes #

fixes bsc#1191654

### DIFF
--- a/internal/pkg/skuba/node/config_test.go
+++ b/internal/pkg/skuba/node/config_test.go
@@ -43,6 +43,7 @@ nodeRegistration:
     cni-bin-dir: /usr/lib/cni
     hostname-override: worker-0
     pod-infra-container-image: registry.suse.de/devel/caasp/4.0/containers/containers/caasp/v4/pause:3.1
+    cgroup-driver: systemd
   name: worker-0
 `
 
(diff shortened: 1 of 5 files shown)


davidcassany (changes requested):
Terraform version should be kept, there is no terraform 0.13 in CaaSP.

To be explicit terraform 0.13 is part of Public Cloud module, however CaaSP ships its own terraform 0.12 i has some strict requirements on v0.12 in CaaSP patterns (`Conflicts: terraform > 0.12.19`)

ci/infra/openstack/versions.tf
```diff
@@ -1,4 +1,15 @@
 
 terraform {
-  required_version = ">= 0.12"
+  required_version = ">= 0.13"
```
davidcassany: I doubt this is correct. We are not supporting terraform 0.13 in CaaSP, so this should always fail.

We did not follow public cloud on terraform release 0.13 because this happened once CaaSP EOF was already announced and running this upgrade was involving changes in terraform syntax config files. So in order to avoid potential regressions and having to test some config files refactor we just let it be.

mjura: Fixed that, thank you for pointing it

davidcassany (approved):
Looks good, thanks!

